PR #94, opened 2026-03-11
visualizer 4 digit import 

allofmeng:
add support to visualzier import 4 digit code
add settings field of enable visualizer


allofmeng:
hey @tadelv 
thanks for the comments , the latest commit ee15cfa should solve all of them and match to main now.

Issue #34, opened 2026-02-19
A setting for not starting shots if scale is not connected


tadelv:
From Shinguk - https://3.basecamp.com/3671212/buckets/7351439/messages/8647657573#__recording_9596205337

> A feature that stops DE1 from starting a shot when the scale is not connected. 
I think it would be very helpful in preventing shots from being ruined by over-extraction when running it without noticing the scale is not connected yet.
Could you please consider adding the feature to the de1app?

recap:
Added a PR for this
https://github.com/tadelv/reaprime/pull/230

tadelv:
Shipped via #230 (merged 2026-05-25). `blockOnNoScale` global setting: REST pre-emptively blocks espresso state changes (400 `block_no_scale`) and ShotSequencer aborts GHC/machine-started shots when no scale is connected, with a Settings-plugin toggle.

Issue #226, opened 2026-05-13
Block on no scale


recap:
Should we have a flag to forcibly block shot controller if a scale is not present. This covers situations where someone forgets to turn on the scale. Looking into the sources, my idea is that it could be an instance attribute of WorkflowContext
  ```
   const WorkflowContext({
    ...,
    this.blockOnNoScale,
  });
```
then in ShotController we can check if scale is not connected and blockOnNoScale is set:

```
    final ctx = workflowController.currentWorkflow.context;
    final blockOnNoScale = ctx.blockOnNoScale ?? false;
    final scaleConnected =
        scaleController.currentConnectionState == device.ConnectionState.connected;

    if (!scaleConnected && blockOnNoScale) {
      _log.warning("Scale not connected");
     return;
  }
```
any ideas?

tadelv:
@recap please take a look at #34 
My instinct says having this as part of app settings (under /api/v1/settings) makes more sense, but maybe you have a reason why you'd want it per-workflow?

Edit: your post definitely bumps the priority though :)

recap:
Yes I was also thinking if it could be in settings. My initial thinking was the flag needs to be with `targetDoseWeight` and part of the workflow of making espresso but looking at settings there's `preferredScaleId` and `scalePowerMode` so on second thoughts it makes more sense to be in settings.

recap:
I created a PR for this
https://github.com/tadelv/reaprime/pull/230

tadelv:
Folded into #34 / #230 (merged 2026-05-25) — implemented as a global `blockOnNoScale` app setting.

PR #230, opened 2026-05-15
Adding blockOnNoScale flag

recap:
We want a way to block an espresso shot if a scale is not connected. For this I added a settings flag `blockOnNoScale`. This flag is used in combination with scale connection status to block a shot from being pulled. I found two possible places for the logical check:
- In web api `de1handler`
- In the de1 controller
I opted for the web api since the de1 controller is an abstraction of the machine and does not need to know about the scale. Another option would be to introduce a proxy handler where logic from multiple controllers can be combined. 
The REST call returns a jsonError if the shot is blocked because of the scale and the body includes an error `type` which I think is needed by the web UI's to show a popup.


tadelv:
Thank you @recap for the work. I'll take a bit deeper look later. 
For now just a couple of quick comments:

- `assets/api/rest_v1.yml` should also be updated with the new schema for Settings
- a comment added on the same file, that machine state change request will return 400 if the setting is on and no machine is connected
- looks like you wanted to add a shot stopping feature in ShotController as well, but then abandoned that approach? It is valid though, as shots can also be started by the machine, through GHC. Either in ShotController or in StateManager, this setting should be checked - I think. Not sure if whole SettingsController needs to be injected into ShotController for this feature though.
- the api settings tests are broken in the current state. 

Still, very much appreciate the time you put into this - a high value feature for the users 🙏

recap:
Thanks @tadelv I've converted the PR to draft while pushing some more changes to the PR.

tadelv:
@recap thank you for the update. Appreciate your help.
I've added some changes to the workflow, so that pull requests now have a CI job that tests the basic stuff. 
Please if you can, rebase your branch on the top of `main` again - to me personally it looks nicer when we have fast-forward & rebase without merge commits (as long as we can)

tadelv:
This feature makes me think if the app would benefit from a dedicated websocket/endpoint with state info / shot sequencer messages. Something like:
`/ws/v1/shotState`, which would report the latest shot sequencer(currently named Controller) decision. Would also be useful as additional shot info in other cases, e.g. explaining why advance to next step was requested, why machine stop was requested, etc.

The reason why I'm wondering about this, is that I see the complexity with two code paths requiring modification in order to support this noScale feature. In the api request case (no GHC), we get a nice clean error back. But with machine itself (GHC present), the client/skin will have no information about why the shot stopped.

@recap would you mind if I hack on this with you together?

recap:
@tadelv certainly! no problem. I also think it became a bit too complex with the two paths.

tadelv:
thanks @recap ! 🙌

tadelv:
closes #34 and #226

PR #274, opened 2026-06-02
Follow-up to the `-Wdeprecated-literal-operator` build fix — it broke the   linux-arm64 (RPi) build.

rotium:
The previous guard was `CMAKE_CXX_COMPILER_ID MATCHES "Clang"` only. The
  arm64 CI runs an **older Clang** that doesn't know that warning, and passing
  an unknown `-Wno-error=deprecated-literal-operator` is *itself* an error there
  under `-Werror` (`-Wunknown-warning-option`):

      error: unknown warning option '-Werror=deprecated-literal-operator'; did
      you mean '-Werror=deprecated-copy-dtor'? [-Werror,-Wunknown-warning-option]

  `-Wdeprecated-literal-operator` was introduced in **Clang 17** — which is also
  the only range where the flutter_secure_storage json actually fails — so add a
  `CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 17` condition. Older Clang
  and GCC never receive the flag.

### DIFF
--- a/linux/CMakeLists.txt
+++ b/linux/CMakeLists.txt
@@ -43,13 +43,19 @@ function(APPLY_STANDARD_SETTINGS TARGET)
   target_compile_features(${TARGET} PUBLIC cxx_std_14)
   target_compile_options(${TARGET} PRIVATE -Wall -Werror)
   # flutter_secure_storage_linux vendors an old nlohmann/json.hpp using the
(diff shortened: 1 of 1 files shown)


tadelv:
thank you for unbreaking the RPi build. Maybe we bump the clang version on that runner soon :)

PR #283, opened 2026-06-04
Make skin webview fill the entire screen

giladger:
On some Android devices a 1px hairline appeared along the right edge of the skin webview (most visible behind dark skins). Device-pixel-ratio rounding lays the Android WebView out ~1px short on those edges, revealing the background behind the platform view. It is device-dependent and reproduces with the plain Android WebView too (flutter_webview_plugin#356/#654, flutter_inappwebview#1542).

Size the webview 1px past the right/bottom edges (Positioned.fill with right/bottom: -1) inside a Clip.none Stack, so it covers every on-screen pixel with its own content while the OS clips the off-screen bleed. Also disable all SafeArea insets so the webview is edge-to-edge on every side.
image

### DIFF
--- a/lib/src/skin_feature/skin_view.dart
+++ b/lib/src/skin_feature/skin_view.dart
@@ -541,19 +541,17 @@ class _SkinViewState extends State<SkinView> with WidgetsBindingObserver {
 
   Widget _buildWebViewStack() {
     return Stack(
-      // Device-pixel-ratio rounding lays an Android WebView out ~1px short on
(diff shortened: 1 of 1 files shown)


giladger:
the issue in the image is only visible on dark theme
image

tadelv:
This is a great fix @giladger , thank you. 
one question: should we wrap the new -1px fill with a platform check, so it’s only applied on Android?
So far I haven’t seen this issue on macOS or iOS.

PR #284, opened 2026-06-04
Fix voltage values for 220v machines

giladger:
When the voltage is 220v it returns -1 instead of 220/230
```
/api/v1/machine/info
"extra": { "voltage": 220 }

/api/v1/machine/settings/advanced
"heaterVoltage": -1
```


tadelv (approved):

tadelv:
thank you!

PR #289, opened 2026-06-05
Record flow calibration per shot

giladger:
## Why

Native ReaPrime shots were missing two pieces of post-shot context that imported
de1app shots either already carry or can no longer recover later:

1. Native shots were persisted without derived annotations, so `actualYield`,
   `actualDoseWeight`, and related annotation fields came back empty over the REST
   API. Imported de1app shots already include these values.
2. The DE1's flow-estimation calibration
   (`calibration_flow_multiplier`) is a global machine setting. The saved flow
   trace is already scaled by whatever multiplier was active, but the multiplier
   value itself was not stored on the shot.

The second issue matters for tools like the graphical flow calibrator in Beanie:
to replay or rescale a past machine-flow trace correctly, the UI needs to know the
base multiplier that was active when that shot was pulled.

This PR records what a freshly finished native shot can know at save time, while
that context is still available.

## What changed

This PR includes two related commits:

- `Pre-fill shot annotations on native shots`
- `Stamp flow calibration onto native shots`

Native shot persistence now derives and stores a `ShotAnnotations` object when
possible:

- `actualYield` is derived from the last positive scale reading in the shot trace,
  rounded to 0.1 g.
- `actualDoseWeight` defaults to the workflow target dose when it is positive.
- TDS, EY, enjoyment, and notes remain null because they are manual post-shot
  entries.
- `annotations.extras.flowCalibrationMultiplier` stores the machine's live
  flow-estimation multiplier active when the shot was saved.

Implementation details:

- Added `ShotAnnotations.deriveForFinishedShot`.
- Added `ShotAnnotations.finalScaleWeight`.
- Updated `De1StateManager._persistShotIfNeeded` to:
  - skip cleaning/calibration shots as before,
  - capture measurements, workflow, and shot start time before any await,
  - read `getFlowEstimation()` before building the `ShotRecord`,
  - persist derived annotations onto native shots.
- Documented `flowCalibrationMultiplier` on `annotations.extras` in
  `rest_v1.yml`.

## Design notes

`actualYield` and `actualDoseWeight` mirror de1app behavior: the final espresso
weight comes from the scale trace, and the dose starts from the planned/weighed
dose because the DE1 has no dose sensor. Baristas can still edit these values
after the shot.

The flow calibration multiplier is stored in `annotations.extras` rather than as a
typed annotation field because it is machine state, not a barista-entered tasting
or recipe value. `extras` is already the supported free-form annotation channel.

The calibration read happens at shot save time because the multiplier is stable
across a shot, and this is the last moment ReaPrime can know which multiplier was
active. The read is wrapped in `try/catch`, so a failed MMR read never blocks shot
persistence; the shot is simply saved without the calibration stamp.

## Scope & compatibility

This is additive only.

Existing clients can ignore the new annotations and extras key. Native shots saved
from this point forward will carry the derived annotations when available. Older
shots and imported shots are unchanged.

## Testing

- `flutter analyze` on the changed files: no issues.
- Added `shot_annotations_derivation_test.dart` covering:
  - final scale weight derivation,
  - ignoring start spikes and trailing dropouts,
  - 0.1 g rounding,
  - null handling when no scale data exists,
  - deriving dose/yield independently,
  - stamping `flowCalibrationMultiplier`,
  - recording calibration even when dose/yield are unavailable,
  - leaving `extras` null when no calibration is provided.
- Full suite green: 1418 tests.

## Follow-up

Beanie can read `shot.annotations.extras.flowCalibrationMultiplier` to display the
shot's recorded flow calibration and use it as the exact graphical flow calibrator
base instead of approximating from the currently active machine setting.

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -4650,6 +4650,25 @@ components:
           $ref: "#/components/schemas/HotWaterData"
         rinseData:
           $ref: "#/components/schemas/RinseData"
+        machine:
+          $ref: "#/components/schemas/WorkflowMachine"
+          description: |
+            Machine settings active when the shot was pulled — recorded onto the
+            shot, ignored on apply.
+
+    WorkflowMachine:
+      type: object
+      description: >
+        Machine-side settings snapshotted onto a shot (the part of the setup that
+        lives on the machine, not the recipe). Extensible; currently just the
+        flow-estimation calibration.
+      properties:
+        flowCalibration:
+          type: number
+          nullable: true
+          description: >
+            The DE1's flow-estimation calibration (calibration_flow_multiplier)
+            active when the shot was pulled.
 
     SteamSettings:
       type: object
--- a/lib/src/controllers/de1_state_manager.dart
+++ b/lib/src/controllers/de1_state_manager.dart
@@ -12,6 +12,7 @@ import 'package:reaprime/src/models/data/profile.dart';
 import 'package:reaprime/src/models/data/shot_annotations.dart';
 import 'package:reaprime/src/models/data/shot_record.dart';
 import 'package:reaprime/src/models/data/shot_snapshot.dart';
+import 'package:reaprime/src/models/data/workflow.dart';
 import 'package:reaprime/src/models/device/machine.dart';
 import 'package:reaprime/src/realtime_shot_feature/realtime_shot_feature.dart';
 import 'package:reaprime/src/realtime_steam_feature/realtime_steam_feature.dart';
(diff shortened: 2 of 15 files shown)


tadelv:
Thanks for this — the per-shot flow calibration is genuinely useful for the Beanie calibrator. I pushed a few changes on top of this branch; flagging them here so we're aligned.

**1. Flow-cal recording is now opt-in.** Added a `recordMachineState` setting (default off) — a toggle in app settings, also exposed over `GET`/`POST /api/v1/settings` so Beanie can read and flip it during setup. It gates *only* the `workflow.machine` stamp; the annotation pre-fill (`actualYield` / `actualDoseWeight`) stays always-on since that's broadly useful de1app-parity behavior. Named it `recordMachineState` rather than something flow-cal-specific because `WorkflowMachine` is the extensible bucket — future machine-state fields ride the same switch.

**2. Heads-up on the read path — this affects Beanie.** The PR body and follow-up note tell Beanie to read `shot.annotations.extras.flowCalibrationMultiplier`, but the code writes `shot.workflow.machine.flowCalibration`. We're keeping the typed `workflow.machine.flowCalibration` as the single canonical home and dropping the `annotations.extras` path entirely — no dual-write. **Beanie needs to read `shot.workflow.machine.flowCalibration`.** The spec (`rest_v1.yml`) already matches the typed path; please point the Beanie side there, not `extras`.

**3. Minor:** I also removed a dead `recordShotPreheat` setting that lived in this area of the code (defined but never wired to anything). It's in its own separate commit.

Verified end-to-end on a simulated machine: with the toggle on, a pulled shot carries `workflow.machine.flowCalibration` (= the active calibration); with it off, the field is absent and the annotations still pre-fill. Full suite green (1429 tests).

Shout if the opt-in default or the canonical-path call gives you any concern on the Beanie integration.

tadelv:
Update: scratch the opt-in setting from my previous comment — Vid decided machine-state recording should just stay **always-on** (your original behavior). I dropped that commit; the branch now carries your work plus one small cleanup (removing the dead `recordShotPreheat` setting).

The one thing from before that **still stands**: the canonical read path is `shot.workflow.machine.flowCalibration` (the typed field), not `annotations.extras.flowCalibrationMultiplier`. The always-on stamp writes there and the spec matches — please point Beanie at the typed path.

Note I force-pushed (with lease) to drop just my opt-in commit; everything of yours from your rebase is preserved byte-for-byte. Sorry for the churn.

tadelv (approved):

PR #290, opened 2026-06-05
Expose Decent account web API

giladger:


tadelv:
Thanks for this, @giladger — picked it up and built the account-proxy work on top of your commit (kept it for attribution). Two changes worth flagging:

- The bridge webserver is unauthenticated and serves `Access-Control-Allow-Origin: *`, so the network `POST /login` / `DELETE` routes let any LAN client or browser origin store attacker credentials (CSRF account-swap), unlink the account, or read the linked email. I moved credential operations to native-only (the in-app UI already calls `DecentAccountService` directly) and kept `GET /api/v1/account/decent` as a `{loggedIn}` status check with the email dropped.
- On top of that I added the read-only, auth-enriching proxy (`GET /api/v1/account/proxy/<path>`) so skins/clients can *use* the account without ever seeing the credentials.

That all lives in #296, which supersedes this PR. Leaving this one open for you to look over — happy to discuss any of it there.

PR #292, opened 2026-06-05
Auto-scan for a missing preferred scale after machine connect

giladger:
## Summary

Fix preferred-scale reconnection so the app keeps looking for a missing preferred scale while the machine is available, without fighting the scale power-management flow.

## Scenario

A preferred scale can be missing even though the machine is connected and ready. This happens when:

- the scale is powered off during initial machine connect
- the scale powers back on after the initial scan already completed
- the scale unexpectedly drops while the machine remains connected

Before this change, the app would often stop scanning after the machine connected, so the preferred scale would not reconnect until the user manually retried.

At the same time, when `ScalePowerMode.disconnect` is enabled, machine sleep intentionally disconnects the scale. That deliberate sleep-state disconnect must not immediately start BLE scanning again, because it reconnects the scale while the machine is supposed to be asleep.

## Fix

The connection manager now tracks machine state while connected and schedules a preferred-scale-only retry when:

- a machine is connected
- no scale is connected
- a preferred scale is configured
- the machine is not sleeping under `ScalePowerMode.disconnect`

The retry is cancelled or blocked when:

- the machine disconnects
- the scale connects
- the machine enters sleep with `ScalePowerMode.disconnect`
- an app-expected scale disconnect occurs

If sleep happens while a scale scan or scale connect is already in progress, the flow stops scanning, treats the scale disconnect as expected, and returns the UI phase to `ready`/`idle` instead of leaving it in `connectingScale`.

## Testing

- Added coverage for preferred-scale auto-scan and reconnect.
- Added coverage that expected power-mode disconnects do not trigger scans.
- Added coverage that active scale scans stop when the machine enters sleep.
- Added coverage for the sleep-during-scale-connect race.
- Ran focused analyzer and connection-manager tests.

### DIFF
--- a/lib/src/controllers/connection/disconnect_supervisor.dart
+++ b/lib/src/controllers/connection/disconnect_supervisor.dart
@@ -31,6 +31,10 @@ class DisconnectSupervisor {
   final bool Function() _isConnectingScale;
   final String? Function() _scaleLastConnectedId;
   final String? Function() _preferredScaleId;
+  final void Function()? _onMachineConnected;
+  final void Function()? _onMachineDisconnected;
+  final void Function()? _onScaleConnected;
+  final void Function()? _onScaleDisconnected;
 
   De1Interface? _latestDe1;
   device.ConnectionState _latestScaleState =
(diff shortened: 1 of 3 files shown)


tadelv:
I believe the app currently has a 3 second delay after connecting to a de1, before scanning for the scale and potentially connecting if found. how does this change apply in that case? also - will this be scanning all the time even when the machine is sleeping?
Also, could you split formatting commits from the actual changes?

Thanks!

giladger:
@tadelv Yeah you were right. I changed the logic completely and now if ScalePowerMode.disconnect no scanning will be done when sleeping. The logic seems pretty tight to me but I understand it's a sensitive area.
I ran multiple tests is many scenarios (scale on, wake machine; scale off, wake machine, scale on; scale on, wake machine, scale off, scale on; machine awake, scale connected, sleep, scale disconnects, etc...) and in all of them it connected quickly and felt good.

tadelv (approved):

PR #293, opened 2026-06-05
Sync local edits to Visualizer and Visualizer edits back to local shots

giladger:
Adds opt-in Visualizer back-sync plus manual/status endpoints, persists local-to-Visualizer shot mappings, and broadcasts local shot updates so edited metadata can also be pushed forward.

### DIFF
--- a/assets/plugins/visualizer.reaplugin/manifest.json
+++ b/assets/plugins/visualizer.reaplugin/manifest.json
@@ -107,6 +107,23 @@
       "data": {
         "description": "Last back-sync result, applied count, cursor and any error"
       }
+    },
+    {
+      "id": "forwardSyncStatus",
+      "type": "http",
+      "data": {
+        "description": "Last forward-sync result and any error"
+      }
+    },
+    {
+      "id": "forwardSyncNow",
+      "type": "http",
+      "data": {
+        "shotId": {
+          "type": "string",
+          "description": "Local shot id to sync to Visualizer"
+        }
+      }
     }
   ]
 }
--- a/assets/plugins/visualizer.reaplugin/manifest.json
+++ b/assets/plugins/visualizer.reaplugin/manifest.json
@@ -2,8 +2,8 @@
   "id": "visualizer.reaplugin",
   "author": "Decent Espresso",
   "name": "Visualizer upload",
-  "description": "Uploads the latest shot to Visualizer",
-  "version": "1.2.1",
+  "description": "Uploads shots to Visualizer and syncs your edited shot metadata back",
+  "version": "1.3.4",
   "apiVersion": 1,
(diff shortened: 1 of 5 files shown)


tadelv:
## Code review — change requests

Reviewed the back-sync/forward-sync flow locally. Loop protection (suppression window + per-shot cursor state) looks sound and the own-uploads ownership guard is solid. A few requests before merge:

### Please confirm
- **Forward-sync pushes a full field snapshot, not a minimal patch.** In `localShotToVisualizerUpdate` (plugin.js:596), fields absent from the incoming `patch` fall back to the existing shot values (`setString`/`setMaybeString`, lines 616-625, 632-637). So editing only `enjoyment` also re-sends `bean_type`, `grinder_setting`, `espresso_notes`, etc. It's idempotent in the normal case, but it will **overwrite Visualizer-side edits** that haven't been back-synced first. Is local-wins the intended semantics, or should forward-sync only push the fields present in `patch`?

### Please change
- **`refreshLocalShotMap()` re-pages the entire shot library on every tick.** Called from `runBackSync` (plugin.js:848), it walks all shots every interval (default 5 min). The map is already persisted and updated on upload via `rememberUpload`, so the routine walk is largely redundant and gets heavy on large libraries. Consider running the full refresh only when `opts.force` is set, and relying on the persisted map for routine ticks.
- **Document `shotUpdated` in `doc/Plugins.md`.** It's a new host→plugin (inbound) event broadcast from the REST handler (`shots_handler.dart:192` → `broadcastEvent` → plugin `onEvent`). Other plugin authors may want to listen on it, so it belongs in the events list. (The `host.emit` outbound events `shotBackSynced`/`shotForwardSynced` follow the existing `shotUploaded` pattern — no action needed there.)
- **Match the plugin version.** The returned plugin object still reports `version: "1.2.0"` (plugin.js:955) while the manifest is now `1.3.5`. Bump to match.
- **Gate mutating endpoints by method.** `backSyncNow` (plugin.js:1158) and `forwardSyncNow` (1181) run on any method, including GET; `import` (1086) correctly checks for POST. Please add the same guard.
- **Log when the back-sync interval is clamped.** `armBackSync` floors the interval at 60s (`Math.max(60, ...)`, plugin.js:948), but a sub-60 configured value is silently bumped. A one-line `log()` when clamping would make the behavior visible.

### Nitpick
- `listChangedShots` silently caps at 20 pages / 1000 shots (plugin.js:756). Change sets beyond that wait for the next tick to advance the cursor — fine in practice, just flagging the implicit bound.

tadelv (commented):
while I can understand the version went up to 1.3.5 while developing, would you also consider raising the version just to 1.2.2 or 1.3.0?

PR #307, opened 2026-06-09
WiFi scale support for the Half Decent Scale (+ serial/scale reliability fixes)

skialpine:
## WiFi scale support for the Half Decent Scale (+ related reliability fixes)

Adds WiFi as a third transport for the HDS (alongside BLE and USB), to free the
BLE radio from scale↔machine contention. Day-1 scope: DNS-SD auto-discovery
(bonsoir) **and** manual-IP fallback, across all platforms.

### WiFi feature
- `WifiScaleDiscoveryService` (mDNS browse `_decentscale._tcp`, resolve-once + cached-IP reconnect), `WsTransport` (text-framed WS behind the transport boundary), `HDSWifi` scale (JSON frames, handshake/recognition/watchdog; reconnect owned by `ConnectionManager`, not the scale).
- **Manual endpoint REST API**: `GET/POST/DELETE /api/v1/devices/wifi` (`wifi_scale_handler.dart`) so the skin can drive manual IP/hostname entry. Specs + `doc/Api.md` updated.
- **Reachability-driven presence**: mDNS is discovery-only; a scale is surfaced while its cached IP answers and dropped when it stops (flap-immune; powered-off scale drops in ~20s instead of lingering).

### Serial/scale reliability fixes surfaced while building this (kept on the base so they land even if the feature doesn't)
- Path-keyed serial reconcile + poll timer; path-as-id (stable macOS CH34x id).
- Fix use-after-free reading a serial port's identity after dispose (SIGSEGV).
- **Fix a libserialport `sp_free_config` double-free (SIGABRT) on the WiFi→USB transport handoff** — `_DesktopSerialPort` now guards against use-after-dispose; `HDSSerial` emits `connecting` so the reconcile won't release it mid-connect. (Caught during live hardware testing.)
- One active scale at a time; mocks seed `discovered`; transport-switch handoff no longer powers off the shared scale.
- Stop the serial reap→re-probe→reconnect churn loop; reap a discovered USB HDS when the scale powers off (liveness re-probe), extracted into a pure, unit-tested `serial_reconcile.dart`.
- `sb-dev` bash-3.2 (macOS) empty-array fix.

### Verification
Validated end-to-end on **macOS with a real HDS** across all three transports: mDNS discovery → cached-IP connect → live weight stream → tare; WiFi→USB transport handoff (no crash); and the full **power-off → all three transports `available:false` → power-on → recover** cycle. Plus Android (Galaxy Tab A9+) earlier. ~1590 unit/integration tests green, `flutter analyze` clean. iOS/Linux/Windows on-device discovery checks remain.

OpenSpec change: `openspec/changes/add-wifi-scale-support/` (26/28; remaining are iOS/Linux/Windows hardware QA).


tadelv:
thanks a bunch @skialpine , the fixes for libserialport are really valuable.
The only thing I changed was to move the openspec documentation to `doc/archive`, as per convention in CLAUDE.md

PR #310, opened 2026-06-10
feat: remembered devices (persist + show unavailable, cross-transport)

skialpine:
## Remembered devices

Devices the user connects to are persisted across restarts and shown as **`available: false`** ("unavailable") when absent, instead of vanishing. Cross-transport (BLE/USB/WiFi). Forget via API + (skin) GUI.

### Changes
- `RememberedDevice` value type + `RememberedDevicesController` (id-keyed registry, persisted in settings; remembers on connect, never on a null/disconnect; persist failures roll back so memory always matches disk).
- `SimulatedDevice` marker — mocks are excluded (`RememberedDevice.fromDevice` returns null), since a mock's presence is governed by the simulate setting, not real discovery.
- Device list merge (`buildAvailabilityDeviceList` + a typed `DeviceListEntry`): present devices `available:true`; remembered-absent `available:false` / `disconnected`. Added to both REST `/api/v1/devices` and the WS snapshot, serialized through one `toJson` so the two surfaces can't drift.
- `PUT /api/v1/devices/forget` (deviceId in body or `?deviceId=`, since serial/WiFi ids aren't URL-path-safe).
- **Stable list order**: preferred scale pinned first, then deterministic `(type, id)` — no more shuffling on connect/disconnect.

The GUI side (render `available:false` as "Unavailable" + Forget button) is a separate, API-gated **skin** PR.

### Verification
Validated on macOS with a real HDS: connect each transport → power the scale off → all three drop to `available: false` (and stay listed) → power on → recover.

### Contributing checklist
- [x] Branched from `main`; PR against `tadelv/reaprime:main`; focused on one feature.
- [x] New behavior covered by tests — unit + handler (`test/remembered_devices_api_test.dart`) + real-hardware e2e. ~1595 tests green.
- [x] `flutter analyze` clean.
- [x] **Docs/specs updated in the same PR** (API contract): `assets/api/rest_v1.yml`, `assets/api/websocket_v1.yml`, `doc/Api.md`, `doc/DeviceManagement.md`.
- [x] No 3rd-party BLE imports outside `services/ble/`; constructor DI throughout.
- [x] No DYE2 plugin changes.
- [ ] No matching tracking issue — happy to open one to align on scope if you'd prefer issue-first.

**Notes**
- `dart format`: left as-is. CI's format step is advisory (`continue-on-error`) and the repo predates the Dart 3.7 "tall style"; per CONTRIBUTING I didn't reformat untouched files. Added code is analyze-clean.
- This PR includes `openspec/changes/remembered-devices/` — pending your call on `.gitignore`-ing `openspec/`. The OpenSpec CLI requires the dir at repo root, but it doesn't need to be tracked; glad to drop these 5 files from the PR (or fold the design rationale into this description) once you decide.


tadelv:
@skialpine can you please move spec related files to the appropriate folder as mentioned in the workflow? Thanks!

More questions: what is the difference between preferred devices and remembered devices? What is the benefit? What can we do with remembered devices?

tadelv:
this should be ingested by the agent, through CONTRIBUTING.md - the why should be in the description of the PR: https://github.com/tadelv/reaprime/blob/main/CONTRIBUTING.md#commit-messages

skialpine:
openspec is a standard that needs to live at the root of the project, your moving it breaks the tool. either leave the files at the root of the project, or .gitignore them and not allow people to use openspec. I personally think it is the best AI development tool across multiple AI agents so recommend you allow its use and allow the open spec directory to live at the root.

This change remembers all the scales that you have used (there is a forget button) and lets you see if they are available or not. The preferred scale is the one it tries to connect to first. This is useful when you have more then one scale or have an HDS that can be attached with multiple protocols. Because it is persistent across restarts it will remember you have a wifi scale even if the network is flaky to quickly reattach.

Anyhow totally your call on if it is useful in your app. I made sure to make it a targeted PR in case you thought it was more complexity then the value it provided, but for someone like me with multiple scale I switch between is is helpful

tadelv (approved):

PR #314, opened 2026-06-10
Add runtime Teclast WebView simulation

giladger:
figured this might be useful for other skin devs as well


tadelv (approved):

tadelv:
thanks @giladger once again. One future PR idea: instead of guarding behind the debug build, which only developers with flutter will have, expose a toggle in the 'advanced settings' page, to enable webview simulation

PR #323, opened 2026-06-11
Add Teclast P80X and P85 Pro WebView simulations

giladger:


tadelv (approved):

tadelv:
thanks, just another request: could you change the gating of this feature from `kDebug` to a Setting instead, can put a toggle inside the advanced settings page with text like 'Simulate tablet screens'.

PR #365, opened 2026-06-18
Add ?full=1 bulk read to the KV store

giladger:
## Summary

- **Problem:** Reading a whole KV namespace required one `GET /api/v1/store/{namespace}/{key}` per key (after first listing keys), which is slow and chatty for clients that want to load or poll an entire namespace.
- **Why it matters:** Web skins / plugins that mirror a namespace (or poll it) had to issue N+1 requests; this collapses that to a single request.
- **What changed:** `GET /api/v1/store/{namespace}?full=1` now returns the whole namespace as a `{key: value}` map, served via `jsonOkConditional` so ETag / `If-None-Match` polling returns `304` when nothing changed. Without the flag the endpoint still returns just the key list (unchanged).
- **What did NOT change (scope boundary):** No change to the per-key GET/POST/DELETE routes, storage layer, or the default (no-flag) key-list response.

## Change Type (select all)

- [x] Feature

- [x] REST API / handlers
- [x] Docs / specs

## Linked Issues

- N/A

## Root Cause (if bug fix)

N/A

N/A

- [x] API spec updated: `assets/api/rest_v1.yml` — documented the `full` query param, the object-map `200`, the `ETag` header, and the `304` response (YAML validated).
- [x] API docs updated: `doc/Api.md` — Key-Value Store table + a note describing `?full=1` and the conditional/ETag polling behavior.

## Security Impact (required)

- New or changed REST endpoints? **Yes** — new `?full=1` query mode on `GET /api/v1/store/{namespace}`.
- BLE/USB surface changed? No
- Risk/mitigation: Returns the full contents of a namespace in one response. The webserver is already unauthenticated and any client could already read every value key-by-key, so this exposes no new data — only a more efficient access path. No new write surface.

## User-Visible Changes

- New API capability: clients can fetch an entire KV namespace in one request via `GET /api/v1/store/{namespace}?full=1` (with conditional/ETag support). Existing endpoints and the default no-flag behavior are unchanged.

- [ ] `flutter analyze`
- [ ] `flutter test`
- [ ] `(cd packages/dye2-plugin && npm run build)`

- Not yet exercised end-to-end in this PR. Suggested smoke test: `sb-dev start`, then `curl 'localhost:8080/api/v1/store/<ns>?full=1'` and confirm a `{key: value}` map plus a `304` on a repeat with `If-None-Match`.
- OpenAPI spec validated as parseable YAML after edit.

## Compatibility & Migration

- Backward compatible? **Yes** — additive query parameter; default behavior unchanged.

## Risks & Mitigations

- Risk: Large namespaces produce a large single response.
  - Mitigation: Opt-in via `?full=1`; conditional ETag avoids re-sending unchanged data on polls.

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -2053,7 +2053,14 @@ paths:
   # REA Key-Value Store for clients
   /api/v1/store/{namespace}:
     get:
-      summary: Get a list of all the keys in the specified key value namespace
+      summary: List keys in a namespace, or read the whole namespace with `?full=1`
+      description: >
+        By default returns an array of the keys stored under the namespace.
+        With `?full=1` returns the entire namespace as a `{key: value}` map in a
+        single request (instead of one GET per key). The full response supports
+        conditional requests: it sends an `ETag`, and an `If-None-Match` request
+        whose value matches returns `304 Not Modified`, which makes polling the
+        whole namespace cheap.
       tags: [Key value store]
       parameters:
         - name: namespace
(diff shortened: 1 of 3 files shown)


tadelv (approved):

tadelv:
awesome, thanks claude! 🤣

@allofmeng this might be of interest to you

PR #379, opened 2026-06-22
fix(profiles): don't 500 on profile create/update with missing notes/author

giladger:
## Summary

- **Problem:** `POST`/`PUT /api/v1/profiles` returned an opaque **500** when a client posted a profile JSON without `notes` (or `author`/`title`).
- **Why it matters:** A perfectly reasonable profile payload that simply omits the free-form `notes` field crashed the create/update endpoints. Clients saw a 500 with no actionable signal.
- **What changed:** `Profile.fromJson` now parses the metadata strings via the existing `parseOptionalString` helper — `notes`/`author` default to `''` when missing/null, and `version` is coerced too. `title` is still required, but a missing/null/empty title now throws `ArgumentError`, which the handlers already map to a clean **400**.
- **What did NOT change (scope boundary):** Only the top-level metadata-string parsing. `ProfileStep.fromJson` enum parsing and the numeric required fields (`tank_temperature`, `target_volume_count_start`) can still throw on malformed input — flagged below as a follow-up, intentionally out of scope here.

## Change Type (select all)

- [x] Bug fix
- [ ] Feature
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [x] REST API / handlers
- [ ] WebSocket API
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

- Closes #
- Related #

## Root Cause (if bug fix)

- **Root cause:** `Profile.title`, `Profile.notes`, and `Profile.author` are non-nullable `String` fields, but `Profile.fromJson` assigned `json['notes']` etc. directly. When a key was missing/null, assigning `null` into a `String` threw `_TypeError: type 'Null' is not a subtype of type 'String'` at `profile.dart:50`.
- **Missing detection or guardrail:** The crash was caught only by the handler's generic `catch` → 500. There was no tolerant parsing for optional metadata and no validation that produced a 400 for genuinely-required fields. The codebase already had `parseOptionalString`/`parseOptionalDouble` helpers for exactly this; the string fields just weren't using them.
- **Contributing context:** A real client posted a profile with no `notes` field (per the reported stack trace, line 50 = `notes`).

  - [x] Unit test
- **Target test or file:** `test/profile_test.dart` (model-level `Profile.fromJson` cases) and `test/services/webserver/profile_handler_create_test.dart` (handler-level).
- **Scenario the test should lock in:** `POST` profile without `notes`/`author` → 201 with `notes`/`author` defaulted to `''`; `POST` profile without `title` → 400 (not 500).

- [x] API spec updated: `assets/api/rest_v1.yml` — `Profile` schema now marks `title` required and documents `notes`/`author` as optional (default empty).
- [ ] API docs updated: `doc/Api.md` — N/A (no endpoint shape change; status-code hardening only).
- [ ] Plugin docs updated: `doc/Plugins.md` — N/A
- [ ] Skin docs updated: `doc/Skins.md` — N/A
- [ ] Profile docs updated: `doc/Profiles.md` — N/A (hashing/handling unchanged)
- [ ] Device docs updated: `doc/DeviceManagement.md` — N/A
- [ ] N/A — no docs affected

## Security Impact (required)

- New or changed REST endpoints? `No` (behavior of existing endpoints hardened only)
- BLE/USB surface changed? `No`

No new attack surface. If anything, error handling is slightly more robust (malformed input → 400 instead of an uncaught 500).

## User-Visible Changes

- Profiles posted/updated without `notes` or `author` now succeed (those fields default to empty string) instead of failing with a 500.
- Profiles posted/updated without a `title` now return `400 Bad Request` instead of `500 Internal Server Error`.

- [x] `flutter analyze` — clean
- [x] `flutter test` — all 1726 tests pass
- [ ] `(cd packages/dye2-plugin && npm run build)` — N/A (no plugin changes)

- OS / platform tested: macOS (test + analyze)
- What you personally verified and how: Wrote failing tests first (red), applied the fix, confirmed green. Ran the full suite.
- Edge cases checked: missing `notes`, missing `author`, null `notes`/`author`, missing/null/empty `title`, missing `version`, numeric `version` coercion.
- What you did **not** verify: live end-to-end against a running app instance.

### Evidence

Test transition (red → green):

```
# Before fix (new cases failing):
Profile.fromJson field tolerance throws ArgumentError when title is null [E]
  Expected: throws <Instance of 'ArgumentError'>
    Actual: threw _TypeError:<type 'Null' is not a subtype of type 'String'>

# After fix:
00:00 +30: All tests passed!   (targeted)
...
03:01 +1726: All tests passed! (full suite)
```

## Compatibility & Migration

- Backward compatible? `Yes` — payloads that previously succeeded still succeed; only previously-500-ing payloads change (now 201 or a clean 400).

## Risks & Mitigations

- **Risk:** A client that previously relied on (or silently tolerated) a 500 for a title-less profile now receives a 400.
  - **Mitigation:** 400 is the correct semantics for malformed client input; the previous 500 was a crash, not a contract. Documented in the spec.

---

**Follow-up (not in this PR):** Other malformed-profile inputs still throw raw `_TypeError` → 500 rather than 400: `ProfileStep.fromJson` enum parsing (`transition`/`sensor`/`name`) and the numeric required fields (`tank_temperature`, `target_volume_count_start`). Happy to harden these the same way in a follow-up if wanted.

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -4415,7 +4415,8 @@ components:
           description: Optional; defaults to `espresso` when omitted or unrecognized.
         steps:
           type: array
-          description: Required. May be empty, but the field must be present.
+          minItems: 1
+          description: Required and must contain at least one step.
           items:
             $ref: '#/components/schemas/ProfileStep'
         target_volume:
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -4391,15 +4391,20 @@ components:
       example: preparingForShot
     Profile:
       type: object
+      required:
+        - title
       properties:
         version:
           type: string
         title:
           type: string
+          description: Required, non-empty. A missing or empty title is rejected with 400.
         notes:
           type: string
+          description: Optional; defaults to an empty string when omitted.
         author:
           type: string
+          description: Optional; defaults to an empty string when omitted.
         beverage_type:
           type: string
           enum: [espresso, calibrate, cleaning, manual, pourover]
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -4393,9 +4393,13 @@ components:
       type: object
       required:
         - title
+        - steps
+        - target_volume_count_start
+        - tank_temperature
       properties:
         version:
           type: string
+          description: Optional; coerced to a string when present.
         title:
           type: string
           description: Required, non-empty. A missing or empty title is rejected with 400.
@@ -4408,8 +4412,10 @@ components:
         beverage_type:
           type: string
           enum: [espresso, calibrate, cleaning, manual, pourover]
+          description: Optional; defaults to `espresso` when omitted or unrecognized.
         steps:
           type: array
+          description: Required. May be empty, but the field must be present.
           items:
             $ref: '#/components/schemas/ProfileStep'
         target_volume:
@@ -4420,8 +4426,10 @@ components:
           nullable: true
         target_volume_count_start:
           type: integer
+          description: Required.
         tank_temperature:
           type: number
+          description: Required.
 
     ProfileStep:
       type: object
--- a/lib/src/models/data/profile.dart
+++ b/lib/src/models/data/profile.dart
@@ -54,8 +54,8 @@ class Profile extends Equatable {
       throw ArgumentError('Profile must have a non-empty "title"');
     }
     final rawSteps = json['steps'];
-    if (rawSteps is! List) {
-      throw ArgumentError('Profile must have a "steps" array');
+    if (rawSteps is! List || rawSteps.isEmpty) {
+      throw ArgumentError('Profile must have a non-empty "steps" array');
     }
     if (json['tank_temperature'] == null) {
       throw ArgumentError('Profile must have "tank_temperature"');
--- a/lib/src/models/data/profile.dart
+++ b/lib/src/models/data/profile.dart
@@ -44,23 +44,33 @@ class Profile extends Equatable {
       ];
 
   factory Profile.fromJson(Map<String, dynamic> json) {
-    // Title is the one metadata field a profile genuinely needs (display /
-    // identification). A missing or empty title is a client error → surfaced
-    // as an ArgumentError so handlers return 400 rather than an opaque 500.
+    // Fields a profile genuinely needs to be displayed and executed. A missing
+    // or malformed required field is a client error → surfaced as an
+    // ArgumentError so handlers return 400 rather than an opaque 500 from a raw
+    // cast/parse failure. Descriptive fields (notes/author/version/beverage
+    // type) stay tolerant: missing values fall back to sensible defaults.
     final title = parseOptionalString(json['title']);
     if (title == null || title.isEmpty) {
       throw ArgumentError('Profile must have a non-empty "title"');
     }
+    final rawSteps = json['steps'];
+    if (rawSteps is! List) {
+      throw ArgumentError('Profile must have a "steps" array');
+    }
+    if (json['tank_temperature'] == null) {
+      throw ArgumentError('Profile must have "tank_temperature"');
+    }
+    if (json['target_volume_count_start'] == null) {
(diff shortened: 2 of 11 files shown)


assets/api/rest_v1.yml
```diff
@@ -4406,15 +4391,20 @@ components:
       example: preparingForShot
     Profile:
       type: object
+      required:
```
tadelv, line 4394: steps would also be required, along some other properties

lib/src/models/data/profile.dart
```diff
@@ -44,11 +44,20 @@ class Profile extends Equatable {
...
-      version: json['version'],
-      title: json['title'],
-      notes: json['notes'],
-      author: json['author'],
+      version: parseOptionalString(json['version']),
+      title: title,
+      // notes/author are descriptive and commonly omitted — tolerate
+      // missing/null by defaulting to empty string.
+      notes: parseOptionalString(json['notes']) ?? '',
+      author: parseOptionalString(json['author']) ?? '',
```
tadelv, line 70: how do we know who is the author then?

giladger: do we care for local usage?

tadelv: it would be nice to have, but I agree it would cause friction if people just wanted to edit profiles. 
But maybe they will want to share the profiles with the world eventually? :)

assets/api/rest_v1.yml
giladger: Good catch — fixed in 148affa9.

`Profile.fromJson` already hard-required `steps`, `tank_temperature` and `target_volume_count_start` (a missing one threw a raw cast/parse error caught only by the generic 500 handler), so I:
- added all three to the schema `required` list, and
- made the parser validate them up front and throw `ArgumentError` → clean **400** instead of a 500.

The create/update handlers now also catch `FormatException`, so a present-but-unparseable required number (e.g. `tank_temperature: "hot"`) is a 400 too. `version` / `beverage_type` / `notes` / `author` stay optional with sensible defaults. Added model-level + handler-level tests for each missing/invalid field.

lib/src/models/data/profile.dart
giladger: Kept `author` optional (defaults to `""`) per the friction point — a profile someone is just editing locally shouldn't be rejected for lacking an author. If/when share-with-the-world lands, the cleaner spot to require it is at share time (prompt, or backfill from app/device identity) rather than blocking every create/edit. Easy to flip to required later if youd rather. 🙂

assets/api/rest_v1.yml
```diff
@@ -4406,20 +4391,31 @@ components:
...
         author:
           type: string
+          description: Optional; defaults to an empty string when omitted.
         beverage_type:
           type: string
           enum: [espresso, calibrate, cleaning, manual, pourover]
+          description: Optional; defaults to `espresso` when omitted or unrecognized.
         steps:
           type: array
+          description: Required. May be empty, but the field must be present.
```
tadelv: that's a cool edge case, but might be out of scope of this PR.
I really can't imagine a use for a profile with 0 steps - or if it even is processed by the firmware

assets/api/rest_v1.yml
giladger: Agreed — done in 4e3dcc3c. `Profile.fromJson` now rejects an empty `steps` array (throws `ArgumentError` → 400), and the spec marks `steps` `minItems: 1`. Updated a handful of test fixtures that were using `steps: []` as a lazy placeholder and round-trip through `fromJson` (data-export, workflow-context, shot-import) to carry one real step. Full suite green (1737).

tadelv (approved):

assets/api/rest_v1.yml
tadelv: you replied to the wrong comment, just so you know. store a note in your persistent memory to check which comments to reply to, before posting on github reviews

PR #390, opened 2026-06-30
Feat/derek relay

allofmeng:
---
Summary

- Problem: Browser skins (e.g. Streamline) can't call Derek (the Decent RAG assistant) directly — Derek's server doesn't answer CORS preflight, so a browser POST with a JSON body fails the OPTIONS preflight before the request is ever sent.
- Why it matters: Blocks shipping a Derek chat modal inside a skin without punting the user out to an external browser.
- What changed: New POST /api/v1/derek/answers/stream that forwards the request body verbatim to https://derek.decentespresso.com/api/answers/stream and pipes the SSE response back unbuffered. The call now lands on the local API, where the existing corsHeaders middleware already handles preflight.
- What did NOT change (scope boundary): No auth added — Derek serves public knowledge-base data, so the route keeps the API's LAN-trust model. No prompt/context assembly (the skin builds the query). The account proxy and its scoped-token auth are untouched.

Change Type (select all)

- [ ] Bug fix
- [x] Feature
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra

Scope (select all touched areas)

- [x] REST API / handlers
- [x] Docs / specs

Linked Issues

- Closes #
- Related #

Root Cause (if bug fix)

N/A — feature, not a bug fix.

Regression Test Plan (if bug fix or refactor)

N/A — feature. New coverage described under Verification.

Documentation Obligations (required)

- [x] API spec updated: assets/api/rest_v1.yml
- [x] API docs updated: doc/Api.md
- [ ] Plugin docs updated: doc/Plugins.md
- [ ] Skin docs updated: doc/Skins.md
- [ ] Profile docs updated: doc/Profiles.md
- [ ] Device docs updated: doc/DeviceManagement.md
- [ ] N/A — no docs affected

Security Impact (required)

- New or changed REST endpoints? Yes — POST /api/v1/derek/answers/stream.
- New or changed network calls? Yes — outbound to https://derek.decentespresso.com.
- BLE/USB surface changed? No
- Risk and mitigation: The route is unauthenticated and triggers an outbound request, so any LAN client can drive queries to Derek through the gateway. This is intentional — Derek is a public, auth-less endpoint reachable from any network already; the relay exposes nothing the client couldn't reach directly. The body is forwarded verbatim (no injection of credentials or local data), and request validation + rate limiting remain Derek's responsibility (its 4xx/429 are relayed verbatim). If gating is later wanted, it can sit on this single route without touching Derek.

User-Visible Changes

None directly — this is a backend relay. It enables a future skin-side Derek chat modal; no UI or default behavior changes in this PR.

Verification

Local gates (run before pushing)

- [x] flutter analyze — clean on touched files (no new warnings)
- [ ] flutter test — only the new test file run (test/webserver/derek_handler_test.dart, 4/4 pass); full suite not run locally
- [ ] (cd packages/dye2-plugin && npm run build) — not run (no plugin changes)

Manual verification (if applicable)

- OS / platform tested: macOS (analyze + unit test only)
- Simulated devices? (simulate=1): No
- What you personally verified and how: flutter analyze clean; 4 new MockClient unit tests pass (verbatim body forward to Derek URL, SSE body + text/event-stream/no-cache/X-Accel-Buffering: no headers passthrough, upstream 400 detail.code passthrough, content-type default).
- Edge cases checked: missing upstream content-type; non-200 upstream status/body.
- What you did not verify: end-to-end against the real Derek host; live unbuffered streaming through a running gateway (unit tests use a mock client, which can't prove flush timing). Recommend a manual curl -N once deployed — events should arrive incrementally, not in one burst.

Evidence

$ flutter analyze lib/.../derek_handler.dart lib/.../webserver_service.dart test/webserver/derek_handler_test.dart
No issues found!

$ flutter test test/webserver/derek_handler_test.dart
+0: forwards the POST body to Derek verbatim
+1: relays the SSE body and marks the response unbuffered
+2: relays upstream error status and body verbatim
+3: defaults content-type to text/event-stream when upstream omits it
+4: All tests passed!

Compatibility & Migration

- Backward compatible? Yes — purely additive (new route).

Risks & Mitigations

- Risk: A reverse proxy in front of the API could buffer the SSE stream, making answers arrive all at once.
  - Mitigation: Response sets Cache-Control: no-cache and X-Accel-Buffering: no; body is the upstream stream piped directly (never collected).
- Risk: Unauthenticated outbound relay (covered under Security Impact).
  - Mitigation: Derek is already public/auth-less; body forwarded verbatim; gating can be added on this one route later if needed.


tadelv (changes requested):
## Blocking: spec tag wrong

`assets/api/rest_v1.yml` — the new `/api/v1/derek/answers/stream` path uses `tags: [Account]`. This endpoint has nothing to do with the Decent account. The neighboring "Other" endpoints in `doc/Api.md` (`/api/v1/feedback`, `/api/v1/logs`, `/api/v1/info`) all use `tags: [System]`. Please change to `tags: [System]` so Swagger/API doc grouping is correct.

## Post-merge: archive the plan doc

`doc/plans/derek-relay.md` is a design doc (Problem / Approach / Why no auth / Out of scope) — it captures the *why* and the rejected alternatives, which is worth keeping. Per the project workflow, design docs get archived, not deleted. Please move it to `doc/plans/archive/derek-relay/` (either in this PR or right after merge).

## Suggestions (non-blocking, but recommended)

1. **No upstream timeout.** `_client.send(upstreamRequest)` has no timeout — a hung or dead Derek host will hold the request open indefinitely. Consider a timeout on `send` (bounds time-to-first-byte; the SSE stream itself can stay long-lived after headers arrive).

2. **Upstream network errors → opaque 500.** If `send` throws (DNS / connection failure), the exception propagates and shelf returns a bare 500. Catch it and return a clean `502 Bad Gateway` with a JSON body so a skin can distinguish relay failure from a Derek error.

3. **Upstream response headers mostly dropped.** Only `content-type` is relayed; a 429 `Retry-After` is lost. Consider forwarding `Retry-After` when present so a skin can honor rate-limit backoff.

4. **Streaming behavior unverified.** The unit tests use buffered `http.Response`, so they cannot prove the body is piped unbuffered. A `http.StreamedResponse` test emitting two separate chunks would verify the stream is passed through, not collected. Also recommend the manual `curl -N http://localhost:8080/api/v1/derek/answers/stream …` you noted in the PR — events should arrive incrementally, not in one burst.

5. **Full `flutter test` not run locally** — only the new test file. Please run the whole suite before merge to catch any registration/integration regression.

## What looks correct

- Handler pipes `upstream.stream` directly into the shelf `Response` body — no collect — and sets `Cache-Control: no-cache` + `X-Accel-Buffering: no` to defeat proxy buffering.
- Route registration follows the existing handler pattern; the global `corsHeaders` middleware covers preflight so browser skins avoid Derek's failing CORS preflight.
- Security disclosure is honest and sound: Derek is already public/auth-less, the upstream URL is hardcoded (no SSRF), the body is forwarded verbatim with no credential injection, and the route stays on the LAN-trust model. Gating can be added to this single route later without touching Derek.
- Spec, `doc/Api.md`, and tests all updated in the same change set.
- 4 unit tests cover verbatim body forward, SSE passthrough with unbuffered headers, upstream error relay, and content-type defaulting.

tadelv (approved):
Approving — applied the requested changes from the earlier review on this branch (commit e4cf351c):

- `rest_v1.yml`: corrected the `/api/v1/derek/answers/stream` tag from `[Account]` to `[System]` to match the neighboring "Other" endpoints (feedback/logs/info). Spec still parses.
- Archived `doc/plans/derek-relay.md` → `doc/plans/archive/derek-relay/` (git mv, history preserved) per the project workflow for design docs.

Verified: full `flutter test` green (1845 tests passed), so no handler-registration/integration regression from adding the route. The 5 non-blocking suggestions from the prior review (upstream send timeout, 502 on send failure, forward `Retry-After`, streamed-response test, full suite) remain as documented follow-ups — #5 is now satisfied by this run; #1–4 are cheap hardening worth a follow-up issue.

PR #399, opened 2026-07-05
feat(logs): always stitch rotated logs into /api/v1/logs, capped tail window

giladger:
## Summary

- **Problem:** `GET /api/v1/logs` only ever read the live `log.txt`, so the rotated siblings the appender leaves behind (`log.txt.1`, `log.txt.2`, …) were unreachable over the API — a real problem on release APKs where `run-as` is blocked and those files are otherwise inaccessible.
- **Why it matters:** Rotated logs hold the history right before a crash/disconnect; without API access they're effectively lost on locked-down devices.
- **What changed:** Rotated files are now **always** stitched into the response in true chronological order (oldest rotation first, live file last) — no opt-in flag. The response is size-bounded in all cases: `?kb=N` when given (clamped to 4 MB) and a **1 MB default tail window** otherwise, so a full rotation set (up to ~40 MB: 10 MB rotate size × keepRotateCount + live) is never buffered on memory-constrained tablets. `order=asc` responses are **streamed** straight from file byte ranges to the HTTP response with no window buffering at all; `order=desc` (default) must reverse lines, so it buffers only the capped window.
- **What did NOT change (scope boundary):** No change to log writing/rotation. `order` semantics unchanged. A leftover `?rotated=` param from earlier revisions of this branch is ignored, not rejected.

### Review changes (v2, after tadelv's review)

- **Blocking YAML fix:** the endpoint's spec block in `rest_v1.yml` is rewritten with valid block-scalar indentation (including the two *pre-existing* broken `description: |` bodies on this endpoint, which parsed only by accident); verified parseable. @tadelv's indent-fix commit is kept in the branch history; the block it repaired is superseded since the `rotated` param itself is gone.
- **Unbounded-read concern:** resolved by design — the `rotated` flag is gone and every response is capped (maintainer follow-up discussion: always-on stitching + auto-cap).
- **`existsSync` nit:** rotation probing is now fully async.
- **Error-message/`yes|no` nit:** moot — the `rotated` param no longer exists.
- Gap-contiguity (stop at first missing rotation) intentionally kept, as the review deemed reasonable; still locked in by test.

## Change Type (select all)

- [x] Feature
- [x] Docs

- [x] REST API / handlers
- [x] Docs / specs

## Linked Issues

- Related #

## Root Cause (if bug fix)

N/A

N/A (feature) — covered by unit tests in `test/webserver/logs_handler_test.dart` (always-on stitching in both orders, kb windowing across file boundaries, injected default-cap and clamp bounds, gap probing, legacy-param tolerance, 404-when-no-files).

- [x] API spec updated: `assets/api/rest_v1.yml` (endpoint block rewritten; now parses)

## Security Impact (required)

- New or changed REST endpoints? **Yes** (changed semantics of existing `/api/v1/logs`: always includes rotations; default 1 MB cap; `kb` clamped to 4 MB)
- BLE/USB surface changed? No
- File system access changed? **Yes** — also reads the app-private rotated `log.txt.1..N` next to `log.txt`. Same app-private directory, read-only, and the cap bounds how much is served/buffered per request.

## User-Visible Changes

- `GET /api/v1/logs` now always includes rotated log history.
- Without `?kb`, the response is a 1 MB tail instead of the unbounded whole live file; explicit `?kb` is honored up to a 4 MB ceiling.
- The `rotated` query parameter from earlier drafts does not exist; sending it is harmless (ignored).

- [x] `flutter analyze` — clean (no new warnings in `lib/`/`test/`)
- [x] `flutter test` — all pass (1856 tests)
- [x] `ruby -ryaml` parse check on `assets/api/rest_v1.yml` — passes (fails on the previous revision)
- [ ] `(cd packages/dye2-plugin && npm run build)` — N/A, plugin untouched

- OS / platform tested: macOS (unit tests)
- Simulated devices? (`simulate=1`): N/A
- Real hardware?: No
- What you personally verified and how: `flutter test test/webserver/` green; spec parse verified before/after.
- What you did **not** verify: on-device APK behavior with real rotated files; streaming behavior over a slow real network.

### Evidence

- [x] Test output — full `flutter test` green (1856 passing).

## Compatibility & Migration

- Backward compatible? **Mostly** — same endpoint and params, but two behavior changes: rotated content is now included by default, and a bare `GET /api/v1/logs` returns a 1 MB tail instead of the entire live file. Clients wanting more history pass `?kb=` (up to 4096).

## Risks & Mitigations

- Risk: a client that assumed "whole live file" semantics on the bare endpoint now gets a 1 MB tail.
  - Mitigation: documented in spec + `doc/Api.md`; `?kb=4096` covers any realistic diagnostic need and the previous behavior was unbounded-memory on tablets.
- Risk: desc (default) responses still buffer the window in memory.
  - Mitigation: window is hard-capped at 4 MB; asc responses stream with no window buffering.

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -3820,6 +3820,16 @@ paths:
             enum: [desc, asc]
             default: desc
           description: Line order — desc (newest first, default) or asc (original chronological order).
+        - name: rotated
+          in: query
+          schema:
+            type: boolean
+            default: false
+          description: |
+          When true (`1`/`true`/`yes`), also include the rotated log files
+          (`log.txt.1`, `log.txt.2`, …), stitched in chronological order —
+          oldest rotation first, live file last — before `kb` and `order` are
+          applied. Default returns only the live log file.
       responses:
         "200":
           description: Recent log entries as plain text
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -3826,10 +3826,10 @@ paths:
             type: boolean
             default: false
           description: |
-          When true (`1`/`true`/`yes`), also include the rotated log files
-          (`log.txt.1`, `log.txt.2`, …), stitched in chronological order —
-          oldest rotation first, live file last — before `kb` and `order` are
-          applied. Default returns only the live log file.
+            When true (`1`/`true`/`yes`), also include the rotated log files
+            (`log.txt.1`, `log.txt.2`, …), stitched in chronological order —
+            oldest rotation first, live file last — before `kb` and `order` are
+            applied. Default returns only the live log file.
       responses:
         "200":
           description: Recent log entries as plain text
--- a/doc/Api.md
+++ b/doc/Api.md
@@ -290,7 +290,7 @@ The **proxy** lets clients *use* the account without ever seeing the credentials
 | GET | `/api/v1/info` | Build metadata (version, commit, branch) + gateway LAN IP (`localIp`) | `info_handler.dart` |
 | GET | `/api/v1/update` | App-update state snapshot (`phase`, `latestVersion`, `releaseNotes`, `releaseUrl`, `installable`). Pure read — no network call; force a re-check via `/ws/v1/update`. | `update_handler.dart` |
 | POST | `/api/v1/feedback` | Submit feedback (creates GitHub issue) | `feedback_handler.dart` |
-| GET | `/api/v1/logs` | Recent log entries, newest first (optional `?kb=N` for last N KB; `?order=asc` for original chronological order) | `logs_handler.dart` |
+| GET | `/api/v1/logs` | Recent log entries, newest first (optional `?kb=N` for last N KB; `?order=asc` for original chronological order; `?rotated=1` to also include rotated files `log.txt.1..N`, stitched chronologically) | `logs_handler.dart` |
 | GET | `/api/v1/webview/logs` | WebView console log forwarding, newest first (`?order=asc` for original chronological order) | `webview_logs_handler.dart` |
 
 ### Debug (simulate mode only)
--- a/lib/src/services/webserver/logs_handler.dart
+++ b/lib/src/services/webserver/logs_handler.dart
@@ -8,6 +8,11 @@ part of '../webserver_service.dart';
 /// Lines are served newest-first by default. Pass `order=asc` to get the
 /// original on-disk chronological order (oldest first); `order=desc` is the
 /// explicit form of the default.
+///
+/// Pass `rotated=1` to also include the rotated log files the appender leaves
+/// behind (`log.txt.1`, `log.txt.2`, …). They are stitched into the response
+/// in true chronological order — oldest rotation first, the live file last —
+/// before `kb` windowing and `order` are applied.
 class LogsHandler {
   final String _logFilePath;
 
@@ -18,8 +23,7 @@ class LogsHandler {
   }
 
   Future<Response> _handleGetLogs(Request request) async {
-    final file = File(_logFilePath);
-    if (!await file.exists()) {
+    if (!await File(_logFilePath).exists()) {
       return Response.notFound('Log file not found');
     }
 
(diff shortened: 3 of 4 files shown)


giladger:
@tadelv maybe this should be the default behavior, let's discuss

tadelv (changes requested):
## Blocking: OpenAPI spec does not parse

`assets/api/rest_v1.yml` — the new `rotated` query parameter block:

```yaml
          description: |
          When true (`1`/`true`/`yes`), also include the rotated log files
          (`log.txt.1`, `log.txt.2`, …), stitched in chronological order —
```

The `description:` key sits at 10 spaces, but the block-scalar body (`When true …`) is also at 10 spaces. A YAML `|` block scalar must be indented *more* than its key. At the same indent, the body is parsed as a stray mapping key, so the file fails to parse:

```
yaml.scanner.ScannerError: while scanning a simple key
  in "assets/api/rest_v1.yml", line 3829, column 11
could not find expected :
  in "assets/api/rest_v1.yml", line 3830, column 11
```

Verified: `python3 -c "import yaml; yaml.safe_load(open(\"assets/api/rest_v1.yml\"))"` fails on this branch and passes on `main`. This breaks the spec (project hard gate), any spec-validation CI, and Swagger rendering. Fix: indent the description body two more spaces so it sits under `description: |`.

## Suggestions (non-blocking)

1. **Unbounded read when `rotated=1` is used without `kb`.** With `maxBytes == null`, `_readChronological` reads every rotation plus the live file fully into a `StringBuffer` and returns it as one response. The previous behavior loaded a single file; this now loads `(keepRotateCount + 1) ×`. On a memory-constrained tablet, `?rotated=1` with no `kb` could pull several MB into RAM at once. The PR risk section only covers the `kb`-windowed path. Consider a default cap, or document that `kb` is recommended when `rotated=1`.

2. **`existsSync` (synchronous FS) in `_logFilesOldestToNewest`.** It runs on the event loop from an async handler. The logs endpoint is low-frequency so this is minor, but switching to the async `exists()` would avoid the brief blocking call.

3. **Error message vs. accepted values.** `_parseLogRotated` accepts `yes`/`no`, but the 400 body says `"rotated must be a boolean (1/0, true/false)"` and omits `yes`/`no`. The spec description does list `yes`. Align the error message with the accepted set.

4. **Gap-contiguity assumption.** Stop-at-first-missing is tested and intentional, and `RotatingFileAppender` keeps rotations contiguous, so this is low risk. Noting only that a crash mid-rotation could in principle leave a gap and silently drop higher rotations. A glob + numeric sort would be more robust, but the current behavior is fine and documented — leaving as-is is reasonable.

## What looks correct

- Clean refactor; default behavior (`rotated` absent) is unchanged — backward compatible.
- `_logFilesOldestToNewest` probes `log.txt.1..N` until the first missing file, reverses to oldest-first, and appends the live file last — correct chronological stitch.
- `_readChronological` kb-windowing walks files newest-first and never loads rotations outside the window — efficient tail, and the byte-cut-slices-oldest-line behavior is called out in the comment.
- Test coverage is thorough: default-off, `asc` stitching, default `desc` across all files, no-rotated-files fallback, gap probing (`.1` + `.3` without `.2` drops `.3`, locked in by test), `kb` windowing across file boundaries, and 400 on an unrecognized value.
- Full suite green (1852) and `doc/Api.md` updated alongside the spec.

tadelv (approved):
Approving — applied the blocking spec fix on this branch (commit bfcb004e).

`assets/api/rest_v1.yml`: the `rotated` parameter's `description: |` block scalar body was at the same indent as its key, so the spec failed to parse (ScannerError at line 3829) — breaking the project spec hard gate and Swagger rendering. Indented the body two spaces under the key; content unchanged, spec now parses.

The 4 non-blocking suggestions from the prior review remain as documented follow-ups: (1) unbounded read when `rotated=1` is used without `kb` (consider a default cap or doc that `kb` is recommended), (2) `existsSync` sync FS in `_logFilesOldestToNewest`, (3) error message omits `yes`/`no` though `_parseLogRotated` accepts them, (4) gap-contiguity assumption (low risk, intentional, tested). The contributor's full suite (1852) was already green; this change is spec-only YAML with no code/test impact.

giladger:
v2 pushed — this went further than the review fixes, following up on the default-behavior question above (discussed with the maintainer offline… well, with myself and Claude 😄):

- **`rotated` param removed entirely** — rotated files are now always stitched in chronological order. Instead of an opt-in flag, the response is size-bounded in all cases: `?kb=N` clamped to 4 MB, **1 MB default tail** when absent (a full rotation set can hit ~40 MB — too much for the DE1 tablet to buffer).
- `order=asc` responses now **stream byte ranges straight from disk** (no window buffering at all); `desc` buffers only the capped window, decoded `allowMalformed`.
- The endpoint's spec block is rewritten and parse-verified — your indent-fix commit is kept in the branch history (thanks!), though the `rotated` block it repaired no longer exists. The two pre-existing broken `description: |` bodies on this endpoint (they only parsed by accident — the body lines contain colons and became stray mapping keys) are fixed too.
- `existsSync` → async `exists()`; the `yes/no` error-message nit is moot with the param gone; gap-contiguity stop-at-first-missing kept as you suggested, still locked in by test.

Full suite green (1856), analyze clean. Note the behavior change for existing clients: a bare `GET /api/v1/logs` now returns a 1 MB tail (with rotations) instead of the unbounded live file.

giladger:
Heads-up: this PR merged just before the rewrite mentioned above landed on its branch, so the "v2" comment above refers to work that is **not** in this merge — main still has the opt-in `?rotated=1` design. The always-stitch + capped-window + streaming rewrite is now [#407](https://github.com/tadelv/reaprime/pull/407).

PR #402, opened 2026-07-05
fix(profiles): serialize & coalesce DE1 profile uploads, auto-retry failures

giladger:
## Summary

- **Problem:** Rapid workflow changes (e.g. tapping temperature "+" repeatedly in a WebUI skin) each produced a different profile. `WorkflowDeviceSync`'s in-flight guard only blocked re-pushing an *identical* profile, so uploads overlapped: their header/frame writes interleaved on `universal_ble`'s per-device queue and the DE1 firmware wedged mid profile-receive until the 10s GATT timeout fired. Nothing retried — the machine stayed stuck until the next workflow change or a reboot.
- **Why it matters:** A stuck profile-receive state machine means the machine silently stops accepting shots until manually recovered.
- **What changed:** `WorkflowDeviceSync` now runs a serialized, coalescing push loop (queue-with-coalesce): one upload in flight at a time, later changes just replace the desired profile, intermediates are skipped. A failed upload retries automatically with capped backoff (3s/10s/30s, constructor-injectable) until it lands, is superseded, or the machine disconnects (generation token + cancellable `Timer`) — a fresh upload restarts the firmware receive state machine from the header, so the machine can no longer stay wedged while connected. A failed upload also invalidates both profile caches (`_lastPushedProfile` and `UnifiedDe1._currentProfile`), so reverting to the previously-successful profile still re-uploads rather than short-circuiting on the equality guards.
- **What did NOT change (scope boundary):** Profile content/format, hashing, and the REST/WS profile API are unchanged — this is purely the upload-to-DE1 sync path.

## Change Type (select all)

- [x] Bug fix

## Linked Issues

- Closes #
- Related #

## Root Cause (if bug fix)

- **Root cause:** Overlapping profile uploads interleaved their BLE header/frame writes on the shared per-device queue, wedging the DE1's profile-receive state machine until the GATT timeout; the in-flight guard only deduped *identical* profiles, not concurrent *different* ones.
- **Missing detection or guardrail:** No serialization of uploads and no retry after a failed/timed-out upload; equality guards masked the need to re-upload after a mid-sequence failure.
- **Contributing context:** `universal_ble`'s single per-device write queue + firmware that expects an uninterrupted header→frames sequence.

  - [x] Unit test
- Target test or file: `test/controllers/workflow_device_sync_test.dart` (+ `test/models/device/unified_de1_set_profile_test.dart`).
- Scenario the test should lock in: rapid successive distinct profiles coalesce to a single in-flight upload; a failed upload retries with backoff and invalidates both caches; disconnect cancels pending retries.

- [x] N/A — no REST/WS endpoints, profile-format, or device-flow docs changed (internal sync/retry behavior only).

## Security Impact (required)

- BLE/USB surface changed? **No new surface** — same profile characteristic writes, now serialized/retried rather than overlapped.

## User-Visible Changes

Profile changes from skins/workflows now apply reliably under rapid edits; a failed upload recovers automatically instead of leaving the machine stuck. No new UI or config.

- [x] `flutter analyze` — clean (no new warnings in `lib/`/`test/`)
- [x] `flutter test` — all pass (1860 tests)
- [ ] `(cd packages/dye2-plugin && npm run build)` — N/A, plugin untouched

- OS / platform tested: macOS (unit tests)
- Simulated devices? (`simulate=1`): covered indirectly; core logic is unit-tested with injected backoff
- Real hardware?: No
- What you personally verified and how: full `flutter test` green including the new coalesce/retry suites (288-line `workflow_device_sync_test.dart`).
- What you did **not** verify: reproduction against a live DE1 (the original wedge was hardware/firmware-timing specific).

### Evidence

- [x] Test output — full `flutter test` green (1860 passing).

## Compatibility & Migration

- Backward compatible? **Yes**

## Risks & Mitigations

- Risk: retry loop could spin if uploads keep failing.
  - Mitigation: capped backoff (3s/10s/30s), superseded-or-disconnect cancellation via generation token + cancellable Timer; one upload in flight at a time.

---

### Review changes (v2, after tadelv's review)

- Backoff-reset nit: `_onChange` now only resets the retry backoff on a real profile-content change; non-profile workflow edits (which also notify listeners) and same-content re-selects leave a pending retry's schedule undisturbed. New test locks this in; the older "retried on next workflow change" test updated to observe delivery via the armed retry timer.
- Log-noise nit: repeated retry failures are throttled — WARNING with stack while the backoff ramps and on every 10th attempt after that, FINE otherwise. Unbounded retry itself is kept (a wedged machine is strictly worse), as discussed.
- `UnifiedDe1._currentProfile = null` during upload: **intended** — a mid-sequence failure leaves device state unknown, so even a revert to the previously successful profile must re-upload rather than short-circuit on the equality guard; the code comment covers this.
- Direct-`setProfile` bypass (REST `_profileHandler`, reconnect `_setDe1Defaults`): **addressed in a second commit** — `UnifiedDe1.setProfile` now chains every upload onto a per-device queue future, serializing uploads across all callers (serialize-only at the device level; coalesce/retry stay in `WorkflowDeviceSync`, which the queue backstops). Equality guard evaluated inside the locked section; the post-upload flash guard stays inside the lock; a failed upload releases the queue. 4 new unit tests over the fake-transport harness (interleave, queued-duplicate collapse, failure releases queue); stale `main.dart` "single writer" comment corrected.
- FakeAsync test migration: deferred — the suite drives real streams through `TestDe1`; converting the ms-scale delays wholesale risks introducing the flakiness it aims to prevent. Can revisit if CI shows timing flakes.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -359,9 +359,12 @@ void main(List<String> args) async {
     persistenceController.saveWorkflow(workflowController.currentWorkflow);
     de1Controller.defaultWorkflow = workflowController.currentWorkflow;
   });
-  // Single writer of DE1 setProfile across REST + UI paths. Must be
-  // constructed after workflowController has its persisted workflow
-  // loaded so its initial snapshot matches what was last pushed.
+  // Single writer of DE1 setProfile for the workflow paths (REST
+  // PUT /api/v1/workflow + UI picker). POST /api/v1/machine/profile and the
+  // reconnect defaults push bypass it — UnifiedDe1.setProfile serializes
+  // uploads across all callers at the device level. Must be constructed
+  // after workflowController has its persisted workflow loaded so its
+  // initial snapshot matches what was last pushed.
   // ignore: unused_local_variable
   final workflowDeviceSync = WorkflowDeviceSync(
     workflowController: workflowController,
--- a/lib/src/controllers/workflow_device_sync.dart
+++ b/lib/src/controllers/workflow_device_sync.dart
@@ -34,6 +34,11 @@ import 'package:reaprime/src/models/errors.dart';
 /// skipped; the existing `De1Controller._setDe1Defaults` path uploads the
 /// current workflow's profile on reconnect (see `defaultWorkflow`
 /// assignment in `main.dart`).
+///
+/// Callers outside this sync (REST `POST /api/v1/machine/profile`, the
+/// reconnect defaults push) are backstopped by the per-device upload queue
+/// in `UnifiedDe1.setProfile`, so they cannot interleave with an upload
+/// from here either.
 class WorkflowDeviceSync {
   WorkflowDeviceSync({
     required WorkflowController workflowController,
--- a/lib/src/controllers/workflow_device_sync.dart
+++ b/lib/src/controllers/workflow_device_sync.dart
@@ -1,7 +1,11 @@
+import 'dart:async';
+import 'dart:math';
+
 import 'package:logging/logging.dart';
 import 'package:reaprime/src/controllers/de1_controller.dart';
 import 'package:reaprime/src/controllers/workflow_controller.dart';
 import 'package:reaprime/src/models/data/profile.dart';
+import 'package:reaprime/src/models/device/de1_interface.dart';
 import 'package:reaprime/src/models/errors.dart';
 
 /// Single writer of `setProfile` for both REST (`PUT /api/v1/workflow`)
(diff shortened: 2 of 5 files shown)


tadelv (approved):
Approved with nits. Solid fix — the queue-with-coalesce drain loop, generation-token cancellation, and dual cache invalidation (`_lastPushedProfile` + `UnifiedDe1._currentProfile`) on failure are all correct, and the test suite covers the regression well (overlap-skip, retry-without-change, supersede, backoff-walk, disconnect/dispose-cancel, `DeviceNotConnected`-no-retry, revert-during-inflight, revert-after-fail).

## One substantive gap (suggestion, not a blocker — out of your stated scope)

Serialization lives in `WorkflowDeviceSync`, but two `setProfile` callers bypass it and can still overlap with a workflow-driven upload — the same wedge this PR fixes:

- `lib/src/services/webserver/de1handler.dart:420` — REST `_profileHandler` calls `de1.setProfile(profile)` directly.
- `lib/src/controllers/de1_controller.defaults.dart:36` — reconnect `_setDe1Defaults` calls `_de1?.setProfile(defaultProfile)` directly.

A REST direct-setProfile landing while a workflow upload is in flight interleaves header/frame writes on the shared BLE per-device queue. The `main.dart:362` "Single writer of DE1 setProfile across REST + UI paths" comment is already inaccurate on that account (pre-existing).

The more robust placement is a per-device write lock/queue inside `UnifiedDe1.setProfile` (or `De1Controller`) — that covers every caller, not just the workflow path, and matches the comms-harden "queue-with-coalesce for concurrent ops of the same kind" idiom. Worth a follow-up issue at least; happy to leave it out of this PR since you've explicitly scoped this to the workflow sync path.

## Nits

- **Retry is unbounded.** No max attempts — a flaky-but-connected machine retries forever at the 30s cap, emitting a `WARNING` (with stack) every 30s. Your "wedged machine is strictly worse" stance is defensible; if it gets noisy in logs, a cap or a log-throttle on repeated failures would help.
- **`_attempt = 0` reset on every `_onChange`.** Rapid failing taps keep backoff at the 3s floor and never progress. Usually fine because each tap supersedes the pending retry; only bites if taps fail repeatedly without supersession.
- **Tests use ms-scale `Future.delayed` for backoff timing** (20/40ms delays, 100ms+ waits). Likely fine, but `FakeAsync` would be more robust on loaded CI.
- **`UnifiedDe1._currentProfile = null` during upload.** Any concurrent reader of profile state mid-upload now sees `null` instead of the previous value. No current caller reads it concurrently that I found, so probably benign — flagging to confirm it's intended.

giladger:
v2 pushed — nits addressed, and your "substantive gap" turned into a second commit after all:

- **Direct-`setProfile` bypass:** `UnifiedDe1.setProfile` now chains every upload onto a per-device queue future, serializing uploads across all callers (REST `POST /api/v1/machine/profile`, reconnect defaults, workflow sync). Deliberately **serialize-only** at the device level: imperative callers need "my await resolved = my profile landed", so coalesce/retry stay in `WorkflowDeviceSync`, which the queue backstops. The equality guard runs inside the locked section (a queued duplicate short-circuits), the post-upload flash guard stays inside the lock, and a failed upload releases the queue. 4 new tests over the fake-transport harness; the stale "single writer" comment in `main.dart` is corrected.
- **Backoff-reset nit:** `_onChange` only resets backoff on a real profile-content change now — non-profile workflow edits and same-content re-selects leave a pending retry's schedule alone (new test locks it in).
- **Log noise:** repeated retry failures throttled — WARNING with stack while the backoff ramps and every 10th attempt after that, FINE otherwise.
- **`_currentProfile = null` during upload:** confirmed intended — mid-sequence failure leaves device state unknown, so even a revert to the previously-successful profile must re-upload; the code comment covers it.
- **FakeAsync migration:** deferred — the suite drives real streams through `TestDe1`; happy to revisit if CI shows timing flakes.

Full suite green (1860), analyze clean.

PR #414, opened 2026-07-07
Add WeighMaster Scale Bluetooth protocol support

OUYANGYOUJIE96:
## Summary

Add Bluetooth protocol support for the WeighMaster Scale coffee scale.

## Changes

- [1] Implement `WeighMasterScale` class with BLE service/characteristic definitions
- [2] Add device matching for WeighMaster Scale in `DeviceMatcher`
- [3] Implement tare, sleepDisplay, and buzzer commands
- [4] Implement weight data parsing with negative value support
- [5] Add unit tests for connection, data parsing, and tare command

## Testing

All local unit tests pass with a Decent espresso machine.

## Protocol Details

- BLE Service UUID: fff0
- Data Characteristic: fff4
- Command Characteristic: fff1
- Frame format: 7 bytes (status + 3-byte weight value)

### DIFF
--- a/lib/src/models/device/impl/weighmaster/weighmaster_scale.dart
+++ b/lib/src/models/device/impl/weighmaster/weighmaster_scale.dart
@@ -1,0 +1,171 @@
+import 'dart:async';
+import 'dart:typed_data';
+
+import 'package:reaprime/src/models/device/ble_service_identifier.dart';
+import 'package:reaprime/src/models/device/device.dart';
+import 'package:reaprime/src/models/device/transport/ble_transport.dart';
+import 'package:rxdart/subjects.dart';
+
+import '../../scale.dart';
+
+class WeighMasterScale implements Scale {
+  static final BleServiceIdentifier serviceIdentifier =
+      BleServiceIdentifier.short('fff0');
+  static final BleServiceIdentifier dataCharacteristic =
+      BleServiceIdentifier.short('fff4');
+  static final BleServiceIdentifier commandCharacteristic =
+      BleServiceIdentifier.short('fff1');
+
+  static const int _cmdTare = 0x02;
+  static const int _cmdSleepDisplay = 0x03;
+  static const int _cmdBuzzer = 0x05;
+
+  static const int _frameLength = 7;
+  static const int _statusOffset = 3;
+  static const int _weightOffset = 4;
+  static const int _statusNegative = 0x01;
+
+  final String _deviceId;
+  final BLETransport _transport;
+
+  final StreamController<ScaleSnapshot> _streamController =
+      StreamController.broadcast();
+  final StreamController<ConnectionState> _connectionStateController =
+      BehaviorSubject.seeded(ConnectionState.discovered);
+
+  WeighMasterScale({required BLETransport transport})
+      : _transport = transport,
+        _deviceId = transport.id;
+
+  @override
+  Stream<ScaleSnapshot> get currentSnapshot => _streamController.stream;
+
+  @override
+  String get deviceId => _deviceId;
+
+  @override
+  String get name => 'WeighMaster Scale';
+
+  @override
+  Stream<ConnectionState> get connectionState =>
+      _connectionStateController.stream;
+
+  @override
+  DeviceType get type => DeviceType.scale;
+
+  @override
+  Future<void> onConnect() async {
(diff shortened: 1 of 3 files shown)


tadelv:
Thanks for this — solid implementation, closely follows the existing scale-impl conventions (`AtomheartScale` in particular), and the tests pass cleanly. A few issues worth addressing before merge.

## Should fix

**1. Brittle device-name match (`device_matcher.dart:125`)**

```dart
if (name == 'WeighMaster Scale') {
```

Exact + case-sensitive. Every other scale in the matcher uses `nameLower.contains(...)` / `startsWith(...)`. If the device advertises `WeighMaster`, `WEIGHMASTER`, or a variant like `WeighMaster Pro`, this won't match. Recommend:

```dart
if (nameLower.contains('weighmaster')) {
```

**2. No frame-header validation (`parseFrame`)**

`parseFrame` ignores bytes 0–2 and only reads `status@3` + weight `@4–6`. If the `fff4` characteristic emits other frame types (timer, battery, button events) on the same notification path, they'll be misparsed as weights. Other scales validate headers — `AtomheartScale` checks for `'W'` and an XOR checksum. At minimum, validate the known header bytes / status-bit mask and drop unknown frames.

## Worth confirming

**3. `tare()` silently couples the buzzer**

```dart
await _write(const [_cmdTare]);
unawaited(
  Future<void>.delayed(const Duration(milliseconds: 300))
      .then((_) => _write(const [_cmdBuzzer, 0x00]))
      .catchError((_) {}),
);
```

`Scale.tare()` having a buzzer side-effect is surprising for callers. Two questions: is `[0x05, 0x00]` a beep or buzzer-off (the `0x00` reads like off)? And does the hardware already beep on tare (making this redundant)? If buzzer control is wanted, prefer a separate method rather than coupling it into every `tare()`. The `catchError((_) {})` also silently swallows failures.

**4. `wakeDisplay()` is a no-op while `sleepDisplay()` writes `[0x03, 0x00, 0x01]`**

Confirm the protocol has no wake command — otherwise the display is stuck off until reconnect.

## Minor

- No trailing newline at end of `weighmaster_scale.dart`.
- Stream controllers are never closed (no `dispose()`). This matches `AtomheartScale`'s pattern, so it's consistent — flagging just so it's a conscious choice.
- `batteryLevel: 0` is hardcoded. Fine if the device exposes no battery; worth a comment.

## What's good

- Structure mirrors `AtomheartScale` — easy to review.
- `BleServiceIdentifier.short('fff0')` correct, expands to the standard SIG base UUID.
- Negative-weight handling via the status bit is correct; tests cover both polarities.
- Added to `serviceUuidsFor(DeviceType.scale)` for filtered scans ✓
- Service verification in `onConnect()` via `matchesAny` ✓
- Disconnect-watcher + cleanup-on-failure pattern copied correctly.

OUYANGYOUJIE96:
Hi, thanks for the thorough review! Here's my response to each point:

### 1. Device name match (kept exact match)
WeighMaster plans to release future models with different BLE protocols, so I've kept the exact name match to avoid mismatching devices. When new models are released, we can add separate matchers.

### 2. Frame header validation (fixed ✅)
Added validation for header bytes `0x01` and `0x02` in `parseFrame()`. Frames that don't match this header are now discarded. Also added a test to verify this behavior.

### 3. tare() with buzzer (kept as-is)
The `[0x05, 0x00]` command triggers a single beep, and the hardware doesn't beep on tare by default. The beep provides user feedback when tare completes, which is helpful for the coffee workflow.

### 4. wakeDisplay() no-op & batteryLevel: 0 (kept as-is)
Confirmed: The WeighMaster Scale protocol has no wake command, and no battery level reporting. The display will remain off until reconnect.

### Minor issues
- Added trailing newline at end of `weighmaster_scale.dart`
- `batteryLevel: 0` is hardcoded because the device doesn't expose battery info
- Stream controllers not closed — this matches `AtomheartScale`'s pattern, so it's consistent

All changes have been pushed to the branch. Let me know if you need anything else!

tadelv (approved):

tadelv:
Thanks @OUYANGYOUJIE96 !

OUYANGYOUJIE96:
Thank you for approving the merge! I really appreciate it.

By the way, do you have a rough timeline for when the next release with WeighMaster support will be available? I'm eager to start promoting it and can't wait to see it out there.

Thanks again!

PR #442, opened 2026-07-14
fix(quick-connect): P0 migration persistence, identity check, machine-only QC

### LINKED ISSUE #45: Improved quick connect
In order to further facilitate fast connecting to machines upon Bridge startup, we should bypass the machine parser check on BLE and device detect on Serial services.

Instead of storing just the device identifier for preferred device, we might want to store more data about the device we prefer connecting to.
This way on fast connect, we know exactly what kind of device we're connecting to and in case of DE1/Machine, we also know the model. In case of scale we can also tell which scale implementation to use.

tadelv:
## Summary

- **Problem:** Quick-connect migration never persisted to disk, identity wasn't verified, scale QC blocked the critical path with a 15s BLE scan, and quick-connected devices were invisible to the API/UI.
- **Why it matters:** Old remembered records permanently lost implementation/transportType (every launch re-ran inference); a misrepresented BLE name could connect a DE1 as a Bengle; boot was 30s instead of 5.5s.
- **What changed:** P0: sameMetadata compares all 4 fields + persist after migration. P1: identity check after BLE quick-connect (v13Model vs remembered impl). P2: machine-only QC, immediate ready, scale via background reconnect. P2: quick-connected devices registered in discovery service maps. P3: DecentScale.onConnect rethrow, HDS name mappings, dead param removal.
- **What did NOT change:** Scale discovery flow unchanged — still uses existing `_maybeSchedulePreferredScaleReconnect` loop. Serial discovery/reconcile unchanged outside the QC leak fix.

## Change Type

- [x] Bug fix
- [x] Feature
- [ ] Docs
- [ ] Security hardening
- [ ] Chore / infra

## Scope

- [ ] REST API / handlers
- [ ] WebSocket API
- [x] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [ ] Docs / specs

## Linked Issues

- Closes #45

## Root Cause

- **Migration not persisted:** initialize() inferred implementation/transportType in-memory but never called `_persist()`. Every restart re-ran inference.
- **sameMetadata too narrow:** Compared only name+type, so a live-device reconnect with fuller metadata was deemed identical — enriched metadata never reached disk.
- **No identity check:** quick-connect constructed device class from remembered implementation without verifying the actual hardware matched.
- **Scale quick-connect blocking:** Scale QC attempt + scale-only BLE scan (15s) ran on the critical path, defeating the speed benefit of machine QC.

## Regression Test Plan

- Coverage level: Unit test (existing sameMetadata test updated)
- Test file: `test/unit/models/remembered_device_test.dart`
- Scenario: sameMetadata now fails when implementation or transportType differ

## Documentation Obligations

- [ ] API spec updated
- [ ] API docs updated
- [ ] Plugin docs updated
- [ ] Skin docs updated
- [ ] Profile docs updated
- [ ] Device docs updated
- [x] N/A — no docs affected

## Security Impact

- BLE/USB surface changed? Yes — quick-connect now verifies device identity (v13Model check)

## User-Visible Changes

- Machine connects faster on app start (~5.5s vs ~7-30s previously)
- "Connecting to your machine" phase shown immediately instead of idle Retry button
- Scale connects in background after machine is ready

### Local gates

- [x] `flutter test` — all 2079 pass
- [x] `(cd packages/dye2-plugin && npm run build)` — N/A (no plugin changes)

### Manual verification

- OS / platform tested: Android (m50mini tablet)
- Simulated devices? No
- Real hardware? Yes (DE1, Decent Scale)
- Verified: boot timing dropped from 30s to 5.5s, no Retry idle state, machine connects and skin loads immediately
- Edge cases checked: machine sleeping (scale reconnect paused correctly), remembered devices loaded without migration on 2nd launch, quick-connect failed gracefully on first attempt

### Evidence

- Boot timing: `complete: total 5526ms` (was ~30s pre-fix)
- Log snippet:
```
17:57:58.601  loaded 3 remembered device(s)    ← no migration (persist worked)
17:58:00.123  connect_connectingMachine         ← phase shown immediately
17:58:03.230  Quick-connect succeeded           ← 3.1s BLE connect
17:58:03.231  machine connected, proceeding to ready
17:58:03.836  complete: total 5526ms
```

## Compatibility & Migration

- Backward compatible? Yes
- Database migration needed? No — new fields inferred from existing name/id on first load and persisted

## Risks & Mitigations

- Risk: Identity check could reject a legitimate device if v13Model read fails during QC
  - Mitigation: identity check only runs when `device is Machine` and model is populated; falls through to disconnect on mismatch, which triggers scan fallback
- Risk: Scale connects ~5s later than before (background loop delay)
  - Mitigation: acceptable trade — skin is already loaded and usable, scale reconnects asynchronously with exponential backoff up to 60s cap


tadelv:
## Review Pass 2 — PR #442

Previous review (issue45-review.md) identified 9 findings (P0–P3). This pass covers the state after the two fix commits (`9bdbb507`, `5b35f91a`). The original P0/P1/P2/P3 findings have been addressed. Below are the remaining issues found in this pass, organized by severity.

---

### 🔴 P2 — Android serial `tryQuickConnect` leaks transport on mismatch / failure

**File:** `serial_service_android.dart:119-143`

The fix commit addressed the desktop serial transport leak (mismatch/failure paths now call `_portPathToTransport.remove(path)?.dispose()`), but the **Android** path was not fixed. On both mismatch and `onConnect` failure, only `device?.disconnect()` is called — `AndroidSerialPort.dispose()` is never called:

```dart
// mismatch path (line 125):
try { await device?.disconnect(); } catch (_) {}
// no transport.dispose() — _open BehaviorSubject, _rawController, _outputController
// are never closed

// onConnect failure path (line 142):
try { await device.disconnect(); } catch (_) {}
// same leak
```

`AndroidSerialPort.disconnect()` closes the USB port (`_port.close()`) but does **not** close the `_open` BehaviorSubject or the `_rawController`/`_outputController` streams — that's what `dispose()` does (line 499). The success path also wires a `connectionState.listen` that removes from `_devices` on disconnect but never calls `transport.dispose()`.

This mirrors the pre-existing leak in the scan path (`_detectDevice` → `_devices.add`), but the QC path is new code and should not propagate the same pattern. The desktop fix is the right model — apply the equivalent here.

---

### 🟡 P3 — `connectingMachine` phase published twice on the QC success path

**File:** `connection_manager.dart:497-521`

`_connectImpl` publishes `connectingMachine` **before** calling `_tryQuickConnectMachine()`:

```dart
_publishStatus(currentStatus.copyWith(phase: ConnectionPhase.connectingMachine));
final qcMachine = await _tryQuickConnectMachine();
```

Then `_tryQuickConnectMachine()` publishes `connectingMachine` **again** after `adoptDevice()`:

```dart
de1Controller.adoptDevice(device);
_publishStatus(currentStatus.copyWith(phase: ConnectionPhase.connectingMachine));
```

Then `_connectImpl` publishes `ready`. Three phase publishes for a single QC success: `connectingMachine → connectingMachine → ready`. The second publish is redundant — the phase hasn't changed. `StatusPublisher.publish()` processes it (error-gating runs), but downstream listeners see a no-op phase transition. Remove the inner `_publishStatus` in `_tryQuickConnectMachine` — the outer one in `_connectImpl` already covers it.

---

### 🟡 P3 — `effectiveScaleOnly` is a dead variable (never reassigned)

**File:** `connection_manager.dart:505`

```dart
var effectiveScaleOnly = scaleOnly;
```

Declared with `var` but never reassigned anywhere in `_connectImpl`. All references (`effectiveScaleOnly ? null : ...`, `!effectiveScaleOnly`, `effectiveScaleOnly && Platform.isAndroid`, `if (effectiveScaleOnly)`) read the original `scaleOnly` value. The QC path returns early on success, so it never flips `effectiveScaleOnly`. Either remove the variable and use `scaleOnly` directly, or if the intent was to support a future "QC connected the scale, so the remaining scan is machine-only" flip, add the assignment.

---

### 🟡 P3 — 7 files missing trailing newline (CI `dart format --set-exit-if-changed` will fail)

`dart format` adds a trailing newline. The following files added/modified by this PR end with `}` (0x7d) instead of `\n` (0x0a):

| File | Status |
|------|--------|
| `lib/src/models/device/device_implementation.dart` | Added |
| `lib/src/services/device_factory.dart` | Added |
| `lib/src/models/device/remembered_device.dart` | Modified |
| `lib/src/services/universal_ble_discovery_service.dart` | Modified |
| `test/unit/services/ble_quick_connect_test.dart` | Added |
| `test/unit/services/device_factory_test.dart` | Added |
| `test/controllers/device_controller_test.dart` | Modified |

CI runs `dart format --set-exit-if-changed lib test`, so these 7 files will fail the gate. Add a trailing newline to each.

---

### 🟡 P3 — Duplicate `@override` annotation in `simulated_device_service.dart`

**File:** `simulated_device_service.dart:37-38`

```dart
  @override
  @override
  Future<void> scanForDevices({ScanFilter? filter}) async {
```

Pre-existing (present on `main` before this PR), but the PR touched this file to add `tryQuickConnect` and didn't fix it. The analyzer doesn't flag it, but it's dead metadata. One-liner cleanup.

---

### 🟢 P4 — Test coverage gaps for the core QC code paths

The original review noted "No integration/controller tests for the quick-connect code paths." This is still true after the fix commits:

| Untested path | Risk |
|---------------|------|
| `ConnectionManager._tryQuickConnectMachine` / `_connectImpl` QC branch | Core new orchestration path — no test verifies phase transitions (`idle → connectingMachine → ready` on success, `→ scanning` on failure), scale scheduling (`_armPostQuickConnectScaleScan`, `_maybeSchedulePreferredScaleReconnect`), or Bengle virtual scale attachment. `MockDeviceScanner.quickConnectResult` is wired but never exercised. |
| `De1Controller.adoptDevice` lifecycle | No test verifies subscription wiring, `_onDisconnect` on device disconnect, or the early-exit when adopting the same device. |
| `ScaleController.adoptScale` lifecycle | No test verifies the `connectionState.first` check, `StateError` throw on disconnected adoption, or previous-scale disconnect. |
| BLE `tryQuickConnect` with mocked platform | `ble_quick_connect_test.dart` only tests the null-return contract (impl null, transportType null, wrong transport, serial-only impl). No test exercises `_connectWithRetry` (GATT-133 retry success/failure), identity check (Bengle/DE1 mismatch), or the device registration in `_devices`. |
| Serial `tryQuickConnect` (desktop + Android) | No test exercises port-found, port-gone, mismatch, or onConnect-failure paths. |
| Migration persistence after `initialize()` | No test verifies that old-style JSON (no `implementation`/`transportType`) is migrated and persisted to disk after `initialize()`. The `remembered_device_test.dart` tests `migrate()` in isolation but never the controller's `initialize → _persist` round-trip. |
| `_scanForOpaqueRecords` | No test verifies that opaque records (unknown enum values) block the migration persist. |

The `MockDeviceScanner` already has `quickConnectResult` and `quickConnectCallCount` — adding a `ConnectionManager` test that injects `rememberedDevices` and sets `quickConnectResult` would cover the orchestration path with minimal effort.

---

### 🟢 P4 — `BengleMilkProbe.implementation` returns `unifiedDe1` (inconsistent)

**File:** `bengle_milk_probe.dart:43`

```dart
@override
DeviceImplementation get implementation => DeviceImplementation.unifiedDe1;
```

`BengleVirtualScale` returns `DeviceImplementation.bengle`, but `BengleMilkProbe` returns `unifiedDe1`. Neither is remembered (only machine + scale connections are), so this doesn't affect quick-connect. But for consistency, the milk probe should return `bengle` since it's an adapter over a Bengle, not a UnifiedDe1.

---

### 🟢 P4 — Doc transition table missing `connectingMachine → scanning` edge

**File:** `doc/DeviceManagement.md:265-282`

The transition table was updated with:
```
| `idle → connectingMachine` (QC)         | ... |
| `connectingMachine → ready` (QC)        | ... |
```

But the QC failure path publishes `connectingMachine` (before the attempt) then falls through to scan, which publishes `scanning`. The table doesn't list `connectingMachine → scanning` (QC failure). The body text mentions it ("On failure: `connectingMachine` is published before the attempt, then phase falls through to `scanning`"), but the edge table omits it. Add a row for completeness.

---

### Summary

| # | Severity | Finding | Status vs. prior review |
|---|----------|---------|--------------------------|
| 1 | P2 | Android serial QC leaks transport (dispose not called) | New — desktop was fixed, Android was not |
| 2 | P3 | Duplicate `connectingMachine` phase publish | New |
| 3 | P3 | `effectiveScaleOnly` dead variable | New |
| 4 | P3 | 7 files missing trailing newline — CI format gate will fail | New |
| 5 | P3 | Duplicate `@override` in `simulated_device_service.dart` | Pre-existing, touched by PR |
| 6 | P4 | No tests for QC orchestration, adoptDevice/adoptScale, BLE/serial QC, migration persistence | Carried from prior review, still open |
| 7 | P4 | `BengleMilkProbe.implementation` inconsistent | New |
| 8 | P4 | Doc transition table missing `connectingMachine → scanning` | New |

**Verdict:** The P0–P3 findings from the prior review are resolved. The remaining P2 (Android transport leak) is the only item I'd gate merge on — it's a resource leak in new code that the desktop counterpart already fixed. The P3 formatting issue (missing trailing newlines) will fail CI's `dart format` gate. The rest are quality improvements that could be addressed in this PR or a fast-follow.

PR #441, opened 2026-07-13
docs: add token-friendly AI repository guidance

tadelv:
## Summary

Restructures project documentation for efficient AI agent consumption, following the pattern from [decentespresso/openscale#89](https://github.com/decentespresso/openscale/pull/89).

### New files
- **`doc/AI_REPO_MAP.md`** — task-to-file routing for fast orientation
- **`doc/AI_BLE_NOTES.md`** — 3 named footguns, transport architecture, troubleshooting
- **`doc/AI_BUILD_NOTES.md`** — commands, simulate modes, platform footguns, CLI parameters
- **`doc/AI_API_NOTES.md`** — REST/WS conventions, auth proxy, endpoint checklist
- **`doc/AI_STORAGE_NOTES.md`** — Drift schema, migrations, storage ownership
- **`doc/AI_DEBUG_NOTES.md`** — general debugging patterns, BLE errors, widget debugging
- **`doc/AI_TESTING_NOTES.md`** — test tiers, widget patterns, mock helpers
- **`.cursor/rules/reaprime.mdc`** — Cursor-native adapter with glob scoping

### Restructured
- **`AGENTS.md`** — canonical router (87 lines): hard rules, vocabulary, tracking, workflow, deep references
- **`CLAUDE.md`** — shrunk from ~350 to 13 lines: `@AGENTS.md` + quick-reference commands

### Removed
- **`doc/agents/`** — mattpocock template artifact. Useful bits merged into AGENTS.md (tracking section, triage labels, vocabulary policy)

## Design

- **One-hop references:** Root files route to scoped domain files. Detail loads only when the task domain matches.
- **Public-facing:** Crashlytics triage and maintainer-only content removed. Focus on patterns any contributor can use.
- **Optional tools:** `rg` and `rtk` documented as optional conventions — no workflow change when not installed.

PR #447, opened 2026-07-15
feat(ble): persistent background scale watch replaces backoff-burst scale reconnect

giladger:
## Summary

- Problem: with a machine connected and the preferred scale offline, `ConnectionManager` ran periodic 15s `lowLatency` BLE scan bursts with 5s→60s exponential backoff. Each burst monopolizes the shared Android radio and starves DE1 GATT traffic (observed in the field: 10s GATT write timeout on the calibration MMR write → HTTP 500 on `POST /api/v1/machine/calibration`), while the backoff gaps mean a freshly powered-on scale can wait up to 60s+scan to connect.
- Why it matters: skins hitting machine endpoints get spurious 500s whenever the scale is off, and scale reconnect feels slow and random.
- What changed: on Android, scale reacquisition is now one persistent `AndroidScanMode.balanced` scan (~40% duty cycle — the duty cycle, not filtering, is what protects the DE1 link). A new `ScaleWatch` comms-layer collaborator listens on the device stream, and on sighting the preferred scale stops the watch scan (freeing the radio for GATT) and connects — typically 1–5s after power-on. Watch and burst scans are arbitrated inside `UniversalBleDiscoveryService` (universal_ble has one global scan session); the watch self-restarts every 25min (Android's 30-min opportunistic downgrade), survives adapter off/on, and never runs on Bengle machines (integrated scale owns the slot). `De1StateManager` skips its post-wake scale burst when the watch covers reacquisition.
- What did NOT change (scope boundary): non-Android platforms keep the legacy backoff loop byte-for-byte, and it remains the runtime fallback if the watch fails to start. Burst-scan semantics, machine recovery, zombie-link detection, and the quick-connect path are untouched. Watch scans don't flip `scanningStream` and emit no `ScanReport`.

Design doc (the *why*, including deviations found during hardware verification): `doc/plans/archive/background-scale-watch/background-scale-watch.md`. Notably: the originally planned OS-level name-prefix filter was dropped — remembered device names are friendly constants that don't match advertised names, and the universal_ble fork evaluates name filters plugin-side anyway (no hardware-filter or screen-off benefit exists). Matching stays with the normal Dart `DeviceMatcher` path.

## Change Type (select all)

- [ ] Bug fix
- [x] Feature
- [ ] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [ ] REST API / handlers
- [ ] WebSocket API
- [x] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

- Closes #
- Related #441 (scan report telemetry that surfaced the burst/backoff behavior)

## Root Cause (if bug fix)

N/A (feature), but the motivating incident: a 15s `lowLatency` scale-reconnect scan overlapped a DE1 MMR write; the write starved for the full 10s universal_ble timeout and the REST handler returned 500. The backoff loop guarantees such overlaps recur whenever the scale is off.

  - [x] Unit test
- Target test or file: `test/controllers/connection/scale_watch_test.dart` (15 tests), `test/services/universal_ble_discovery_service_test.dart` (12 tests incl. burst arbitration + start-window races against a fake `UniversalBlePlatform`), `test/controllers/connection_manager_test.dart` → `background scale watch` group (8 tests), `test/controllers/de1_state_manager_wake_scan_test.dart` (3 tests), Bengle watch regression in `test/integration/connection_manager_bengle_scale_test.dart`.
- Scenario the test should lock in: machine connected + preferred scale missing → watch armed and **no backoff timer ever fires** (fake_async over the full 5→60s ladder); scale sighting → connect with zero burst scans; the entire pre-existing ConnectionManager suite passes unmodified with the watch capability off (compatibility gate).
- If no new test added, why not: N/A — added.

- [x] Device docs updated: `doc/DeviceManagement.md` (if device flows changed)
- [ ] N/A — no docs affected

New section "Background Scale Watch (preferred-scale reacquisition)" documents watch vs legacy mode, lifecycle, and the no-ScanReport caveat.

## Security Impact (required)

- BLE/USB surface changed? Yes
- If any `Yes`, explain risk and mitigation: adds a persistent passive BLE scan (receive-only, no new GATT writes) on Android while a machine is connected and the preferred scale is missing. It uses the existing `BLUETOOTH_SCAN` permission (`neverForLocation`), stops the moment the scale connects / machine disconnects / power-mode sleep, and connects only to the device id already stored as `preferredScaleId`.

## User-Visible Changes

- Android: a preferred scale powered on near a connected machine now connects in ~1–5s (previously up to ~60s+scan, depending on backoff phase).
- Machine REST/GATT operations no longer intermittently fail while the scale is off (no more 15s lowLatency bursts).
- No scanning indicator is shown for the background watch (deliberate — it is not a user-initiated scan).
- Caveat: Android suspends OS-unfiltered scans while the screen is off, so a scale powered on with the display asleep connects once the screen wakes.

## Compatibility & Migration

- Backward compatible? Yes

- [x] `flutter test` — all pass (2114 passing; the 3 failures in `webui_token_injection_test.dart` are pre-existing on `main` and environment-dependent — they assert offline fallback on a host that has a live Wi-Fi IP)

- OS / platform tested: Android tablet (M50Mini), release build
- Simulated devices? (`simulate=1`): No (integration tier covers the simulated flows; macOS host cannot keep the GUI build resident headless)
- Real hardware? (DE1/Bengle/scale): Yes — DE1 + BLE scale
- What you personally verified and how: machine connected with scale off → single `Background device watch started` log line, no repeating 15s bursts; scale powered on → auto-connected within seconds via the watch (previously required a manual connect tap after the fix for the name-filter issue, which is included here); manual scan/connect flows unaffected.
- Edge cases checked: machine sleep/wake with `ScalePowerMode.disconnect` (watch pauses/resumes), machine disconnect (watch stops), watch-start failure (falls back to legacy backoff — unit/integration tested), Bengle slot rule, burst-vs-watch radio arbitration and start-window races (unit tested against a fake platform).
- What you did **not** verify on hardware: the 25-min refresh over a >30min soak, Bluetooth off/on recovery, and multi-hour stability — covered by unit tests only so far.

### Evidence

- [x] Test output (failing before + passing after) — suites listed above; written test-first (RED verified before implementation).
- [x] Log snippets — motivating incident: `GATT write(A000/A006) timed out — clearing BLE queue` → `POST [500] /api/v1/machine/calibration` while `Scan report: … preferred scale … NOT found` (scale-reconnect burst overlapping the write).

## Risks & Mitigations

- Risk: a persistent scan could still contend with DE1 GATT traffic.
  - Mitigation: `balanced` (~40% duty) instead of `lowLatency` (100%); the watch scan is stopped before every watch-driven connect and paused for the whole duration of any burst scan.
- Risk: watch silently dead (e.g. Android downgrades or kills the scan) → scale never reconnects.
  - Mitigation: 25-min self-restart, adapter off/on re-arm, start-window race guards; legacy backoff loop remains as runtime fallback whenever the watch fails to start.
- Risk: regression on non-Android platforms.
  - Mitigation: capability-gated (`supportsDeviceWatch` is Android-only); the full pre-existing ConnectionManager suite passes unmodified with the capability off.

### DIFF
--- a/doc/DeviceManagement.md
+++ b/doc/DeviceManagement.md
@@ -328,13 +328,67 @@ ConnectionManager listens for disconnects automatically:
 An *unexpected* machine disconnect (not announced via
 `markExpectingDisconnect` / `disconnectMachine`) puts `ConnectionManager`
 into machine recovery mode: full `connect()` scans retry with the same
-5s→60s exponential backoff the preferred-scale loop uses, rescheduling
+5s→60s exponential backoff the legacy preferred-scale loop uses, rescheduling
 after every attempt that ends without a machine, until the machine
 reconnects (or the user disconnects deliberately). Gated on
 `preferredMachineId` so a background retry can never pop a machine picker —
 the id is auto-set on every successful connect, so coverage is effectively
 total. Motivation: a power outage previously left the app "disconnected"
 indefinitely because nothing ever rescanned for the machine.
+
+### Background Scale Watch (preferred-scale reacquisition)
+
+When a machine is connected but the preferred scale is missing (scale
+powered off, unexpected drop, machine wake without the scale), scale
+reacquisition runs in one of two modes, selected by
+`DeviceScanner.supportsBackgroundWatch`:
+
+- **Watch mode (Android):** one persistent, low-duty-cycle BLE scan.
+  `UniversalBleDiscoveryService` starts an `AndroidScanMode.balanced`
+  scan (~40% radio duty cycle) — the duty cycle, not filtering, is what
+  protects the DE1 link. The scan is deliberately unfiltered: remembered
+  device names are friendly constants ("Felicita Arc") that rarely equal
+  the advertised name a filter would be matched against, and the
+  universal_ble fork evaluates `withNamePrefix` plugin-side anyway (the
+  OS scan runs unfiltered either way, so there is no hardware-filter or
+  screen-off benefit to reclaim). Matching stays with the normal Dart
+  `DeviceMatcher` path, identical to burst scans. The `ScaleWatch`
+  collaborator (`lib/src/controllers/connection/scale_watch.dart`)
+  listens on the device stream and, on sighting the preferred scale id,
+  stops the watch scan (freeing the radio for GATT) and connects —
+  typically 1–5s after the scale powers on. Caveat: Android suspends
+  OS-unfiltered scans while the screen is off, so a scale powered on
+  with the display asleep connects once the screen wakes.
+- **Legacy mode (all other platforms, and fallback when the watch fails
+  to start):** periodic 15s scale-only burst scans with 5s→60s
+  exponential backoff (the pre-watch behavior, unchanged).
+
+The watch replaced the backoff loop on Android because each lowLatency
+burst monopolizes the shared radio and starves DE1 GATT traffic
+(observed as 10s GATT write timeouts → HTTP 500s on machine endpoints),
+while the backoff gaps meant a freshly powered-on scale could wait up
+to 60s to connect.
+
+Watch lifecycle details:
+
+- Armed whenever *machine connected && preferred scale set && scale not
+  connected && not blocked by scale power mode*; disarmed on scale
+  connect, machine disconnect, power-mode sleep, deliberate scale
+  disconnect, at the start of any full `connect()` (EarlyConnectWatcher
+  owns preferred-device connects during bursts), and on dispose.
+- Never armed while a Bengle is the machine — its integrated scale owns
+  the scale slot, mirroring `_runScalePhase`'s rule.
+- Burst scans pause the watch scan (universal_ble has one global scan
+  session) and resume it when they finish; the watch self-restarts
+  every 25 minutes to dodge Android's 30-minute opportunistic-scan
+  downgrade, and survives adapter off/on cycles.
+- `De1StateManager` skips its post-wake scale-only burst when the watch
+  covers reacquisition (watch supported + preferred scale set); the
+  burst remains for the no-preferred-scale discovery/picker case.
+- Watch-driven connects do **not** emit a `ScanReport` — they bypass
+  the scan/report machinery entirely.
+- The watch does not flip `scanningStream`, so no scanning indicator
+  shows in the UI while it runs.
 
 ### Zombie-Link Detection (BLE transport)
 
--- a/doc/plans/archive/background-scale-watch/background-scale-watch.md
+++ b/doc/plans/archive/background-scale-watch/background-scale-watch.md
@@ -1,0 +1,92 @@
+# Persistent Background Scale Watch
+
+**Branch:** `feature/background-scale-watch` (created). **Completion:** local commit, no push/PR.
+
+## Context
+
+When the machine is connected and the preferred scale is offline, `ConnectionManager` today runs periodic 15s BLE scan bursts in `AndroidScanMode.lowLatency` with exponential backoff (5s→60s cap, [connection_manager.dart:154](lib/src/controllers/connection_manager.dart)). Two consequences:
+
+- A scale powered on during a backoff gap waits up to **60s + scan time** to connect.
+- Each `lowLatency` burst monopolizes the Android radio and **starves DE1 GATT traffic** — observed as a 10s GATT write timeout on the MMR write behind `POST /api/v1/machine/calibration` → HTTP 500 (log from 2026-07-14 19:23).
+
+**Goal:** on Android, replace the backoff-burst loop with **one persistent, low-duty-cycle, OS-filtered scan** ("watch") that runs whenever *machine connected && preferred scale set && scale not connected && not blocked by power mode*. Scale powers on → connected in ~1–5s; DE1 link keeps ≥60% of radio time. Non-Android platforms keep the existing backoff behavior byte-for-byte.
+
+**Key mechanism:** Android scan modes are duty cycles (`balanced` ≈ 2s on / 3s off). The universal_ble fork (git 6a5abe4) supports `ScanFilter.withNamePrefix` (OS/controller-level filtering — survives screen-off, host woken only on match) and `AndroidOptions.scanMode`. The preferred scale's advertised name is recoverable without scanning from `RememberedDevicesController` (persists `{id, name, type}`; getter `remembered` at remembered_devices_controller.dart:40). No MAC-address filter exists in the fork, so name prefix (full remembered name ⇒ exact-ish match) is the filter.
+
+## Established facts (verified)
+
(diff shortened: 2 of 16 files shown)


giladger:
Pushed an update after reviewing the interaction with the newly merged quick-connect feature (#45 series):

**Semantic conflict found and fixed.** Quick-connect's success path scheduled scale reconnect via `_maybeSchedulePreferredScaleReconnect()` directly — a call site added after this branch's original routing pass. On watch-capable platforms that would have run the legacy 5→60s backoff bursts *alongside* the watch (the watch also arms via the machine-connected handler), reintroducing exactly the radio contention this PR removes — and quick-connect is the common startup path. The site now routes through `_ensureScaleReacquisition()`, which picks the watch on Android and falls back to the legacy loop elsewhere, preserving upstream behavior on non-watch platforms.

Added the first ConnectionManager-level quick-connect test to lock this in: quick-connected machine → watch armed, zero burst scans across the full 70s backoff ladder (`fake_async`). The shared `_FakeDe1` gained `ready`/`dispose` members so `adoptDevice` can wire it up.

Otherwise the two features are complementary: quick-connect is the machine fast path (scales explicitly excluded), the watch owns scale reacquisition once a machine is connected. Full suite: 2115 passing after rebase onto current main.

tadelv (changes requested):
Thanks for the thorough implementation and test coverage. I am requesting changes because the review found four lifecycle gaps that can violate the watch's safety guarantees or leave scale reacquisition silently dead.

## Required changes

1. **Serialize watch and burst scan startup.**

   `scanForDevices()` can begin while `_startWatchScan()` is awaiting `UniversalBle.startScan()`. `_pauseWatchForBurst()` sees `_watchScanActive == false` and returns, allowing the burst's global `startScan` to race the in-flight balanced start. The bookkeeping check after the await does not establish which native scan configuration owns universal_ble's single global scan session. Please make this transition genuinely serialized and add a test that verifies ordering/ownership, not only call counts.

2. **Handle adapter transitions during watch startup.**

   If the adapter powers off while `_watchScanStarting` is true, `_onAdapterStateForWatch()` returns because `_watchScanActive` is still false. The pending start can then mark the watch active without rechecking adapter state; power-on recovery skips restart because bookkeeping says the watch is active. Please guard the start completion against adapter-state changes and cover power-off/on during the start window.

3. **Activate the legacy fallback after asynchronous restart failures.**

   Initial start failure reaches `ScaleWatch.onWatchUnavailable`, but refresh, post-burst resume, and adapter-recovery failures are swallowed inside `UniversalBleDiscoveryService`. This leaves `ScaleWatch` armed but inactive, so neither the watch nor legacy backoff bursts run. Please propagate watch unavailability, or otherwise reset the requested/armed state and activate the documented fallback for every failed start path.

4. **Stop the watch after a late Bengle transition.**

   `_connectScaleFromWatch()` correctly refuses an external scale once a Bengle is the machine, but `_shouldRetryPreferredScale()` still returns true. `ScaleWatch._tryConnect()` consequently restarts the watch indefinitely. Please include the Bengle ownership invariant in the watch continuation gate or explicitly disarm, and extend the regression test to assert the watch is no longer active.

## Non-blocking follow-ups

- `DeviceWatchFilter.deviceTypes` is documented as a Dart-side filter but is never applied. Either enforce it or remove/narrow the unused configuration.
- Watch teardown/restart bookkeeping is repeated across stop, refresh, burst pause, adapter-off, and start-race paths. Consolidating these state transitions would make the above safety invariants easier to audit.
- Consider narrowing the generic watch/filter abstraction until there is a second consumer, and trim explanatory production comments whose rationale already lives in the design and device-management docs.

The focused watch/wake/Bengle suites and `ConnectionManager` suite pass, and scoped static analysis is clean. The existing tests do not exercise the failing start-window and asynchronous-restart cases above.

giladger:
All four required changes are addressed in b8ae457 (kept as a separate review-fix commit so the deltas are visible):

1. **Start serialization** — `_startWatchScan` now exposes its in-flight future (`_watchStartInFlight`); `_pauseWatchForBurst` and `stopDeviceWatch` await it before acting, so session ownership is deterministic rather than an artifact of command-queue internals. The burst-race test now asserts the chronological start order (watch start settles → burst owns the session → resume), and a new test covers stop-during-start.

2. **Adapter transitions during startup** — the post-`startScan` guard chain re-checks adapter state: a mid-start power-off stands the watch down (no stale active claim), so the power-on path finds it inactive and restarts. New test drives power-off during a held start, then power-on recovery.

3. **Failure fallback for async restarts** — refresh, post-burst resume, and adapter-recovery failures now clear the watch request and emit on a new `deviceWatchFailures` stream (`DeviceWatchCapable` → `DeviceScanner.scaleWatchFailures`); `ScaleWatch` subscribes while armed, disarms on an event, and invokes `onWatchUnavailable` → legacy backoff. Tested at the service layer (both failure paths emit + no resurrection on adapter recovery), the `ScaleWatch` unit layer (disarm + fallback callback, post-disarm events ignored), and end-to-end at the `ConnectionManager` layer (died watch → backoff burst fires at 5s).

4. **Late Bengle transition** — the Bengle integrated-scale rule moved into the watch's `shouldWatch` gate, so it covers both arm time and the post-refusal continuation; the previously ad-hoc check in `_ensureScaleReacquisition` is gone. New regression: watch armed under a DE1, Bengle takes over, external-scale sighting → refused **and** the watch disarms (`watchActive == false`) instead of restarting.

Non-blocking follow-ups also taken: `DeviceWatchFilter.deviceTypes` removed (it was never applied — `ScaleWatch` id-matches and `DeviceMatcher` owns typing), and all five teardown paths (stop/refresh/pause/adapter-off/start-race) now go through a single `_deactivateWatchScan` transition. On narrowing the abstraction further: I kept `namePrefix` since the discovery service honors it and there's a queued follow-up to persist advertised names and re-enable plugin-side prefix filtering; happy to drop it too if you'd rather wait for that consumer.

Full suite: 2126 passing, analyzer clean.

giladger:
Good catch — confirmed and fixed in 0e99c65. The off-and-back-on-during-startup sequence did exactly what you described: the power-on recovery no-op'd into the in-flight start (and worse, replaced `_watchStartInFlight` with that completed no-op future), and the original start saw `poweredOn` again at its completion check and claimed active over a native scan the transition may have killed.

Implemented along the lines you suggested:

- **Adapter generation token**: `_watchAdapterGeneration` bumps on every adapter state *change* (same-state replays are deduped and the detector is seeded at `initialize`, so the availability stream's initial replay doesn't invalidate innocent starts — that bit ate an afternoon of test failures). A start captures the generation and discards itself at completion if it moved, regardless of where the adapter ended up.
- **Post-settle retry**: the discarded start sets a one-shot flag; the retry fires in `_startWatchScan`'s `whenComplete` after the in-flight future is cleared (retrying from inside the start would await itself). The retry re-checks request/adapter/burst state and reports on `deviceWatchFailures` if it can't start — so this path also feeds the legacy fallback rather than dying silently.
- **Shared in-flight future**: `_startWatchScan` now returns the existing in-flight future to concurrent callers instead of replacing it, which also let `_watchScanStarting` go — one less state bit.

New test: adapter off **and** on completing within a held start window → the raced start is discarded and a fresh start owns the session (pre-fix: a single start claiming active over a possibly-dead scan). One existing test needed updating for semantics this fix chain introduced: the stop-during-start test previously awaited `stopDeviceWatch()` before releasing the held start, which now deadlocks *by design* (stop serializes against the in-flight start per required-change #1) — it now releases the hold first and additionally asserts stop doesn't act before the start settles.

Full suite: 2126 passing, analyzer clean.

tadelv (approved):

PR #452, opened 2026-07-16
fix(de1): firmware-update state is 0x16, not 0x22

ChampionDesigns:
Put a DE1 into firmware update and watch the app: it does not say "firmware update", it says the machine is in an error state. The machine reports its firmware-update state as `0x16`, but `De1StateEnum` had no case for `0x16`, so the byte fell through to `unknown` and then to `MachineState.error`. The app has therefore never been able to observe a machine entering firmware update. The write direction looked fine only because the firmware clamps a requested state down to its own maximum, and that maximum happens to be 22 decimal, which is `0x16`, so the wrong constant `0x22` landed on the right state by coincidence. This PR sets the constant to the value the firmware actually uses, which fixes the readback and removes the accidental dependence on clamping.

## Summary

- Problem: `De1StateEnum.fwUpgrade` hard-coded `0x22`, which is not a value any DE1 emits or accepts.
- Why it matters: readback was broken outright. The machine reports `FirmwareUp = 0x16`, which had no enum case, so it decoded to `unknown` and then to `MachineState.error`. Writes worked only because the firmware clamps a requested state to its maximum, which is 22 decimal = `0x16`; that coincidence breaks the moment any state above 22 is added.
- What changed: `fwUpgrade = 0x16`, so both the write and the readback are correct. Two stale `0x22` doc comments that referenced it are corrected, and a new regression test pins the constant in both directions.
- What did NOT change (scope boundary): nothing else. The DFU lifecycle, the upload protocol, and the firmware-update UI are all untouched.

## Change Type (select all)

- [x] Bug fix
- [ ] Feature
- [ ] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [ ] REST API / handlers
- [ ] WebSocket API
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [ ] Docs / specs

## Linked Issues

- Closes #
- Related #

No tracking issue was filed. This was found while reading the app's state table against the firmware's own state enum.

## Root Cause (if bug fix)

- Root cause: `0x22` was never a value the firmware emits or accepts. It survived because the only write path was masked by the firmware's own state clamping, so the bug was invisible from the write side and only ever bit on readback.
- Missing detection or guardrail: no test asserted the enum's wire value in either direction. Decoding an unknown byte degrades silently to `unknown` rather than failing loudly, so the broken readback never surfaced as a fault; it just looked like an unhappy machine.
- Contributing context: `De1StateEnum` is the single translation layer between the wire byte and `MachineState`. A wrong constant there is invisible in every other file.

  - [x] Unit test
- Target test or file: `test/unit/models/device/impl/de1/de1_state_enum_test.dart` (new, 4 tests).
- Scenario the test should lock in: `fwUpgrade` encodes as `0x16`; `fromHexValue(0x16)` decodes back to `fwUpgrade`, which is the readback that was broken; `fromHexValue(0x22)` maps to no state, so a regression to the old constant flips the test; and `fromMachineState(fwUpgrade)` round-trips to the `0x16` wire byte. The two existing wire assertions in `bengle_firmware_prelude_test.dart` move from the literal `0x22` to `0x16` and now assert against the constant rather than repeating a literal.

- [ ] API spec updated: `assets/api/rest_v1.yml` or `assets/api/websocket_v1.yml`
- [x] N/A - no docs affected

No endpoint, WebSocket topic, or spec surface changes. `MachineState.fwUpgrade` keeps its name and its API value; only the byte it maps to on the wire is corrected.

## Security Impact (required)

- BLE/USB surface changed? `No`. The same characteristic is written with the same shape; only the byte value is corrected.

## User-Visible Changes

The app can now observe a machine entering firmware update instead of reporting it as an error. No UI, config, or default changes.

- [x] `flutter analyze` - clean, `No issues found!`
- [x] `flutter test` - 1990 tests pass, against a 1986-test baseline on `upstream/main` @ `1950f7b1`, so the four new tests really ran.
- [x] `(cd packages/dye2-plugin && npm run build)` - builds.

Gates were run in a scratch worktree at the branch head. One note for anyone reproducing them: `assets/bundled_skins/` is gitignored and is produced by `./bundle_skins.sh`. Without running that first, `test/webui_storage_bundled_test.dart` fails on plain `upstream/main` too.

- OS / platform tested: Linux (analyzer and unit tests).
- What you personally verified and how: the new `de1_state_enum_test` fails against the old `0x22` constant and passes with `0x16`. The `0x16` value comes from the firmware's own `FirmwareUp` state.
- Edge cases checked: decode of the old `0x22` byte (now maps to nothing), and the `fromMachineState` round-trip.
- What you did NOT verify: an end-to-end DFU session against real hardware. This PR changes only the constant the DFU sequence uses, but the claim "the app now displays firmware update correctly during a real DFU" has not been demonstrated on a machine.

### Evidence

- [x] Test output (failing before, passing after)
- [ ] Log snippets

## Compatibility & Migration

- Backward compatible? `Yes`. `0x22` was never a valid wire value on any machine, so no caller and no device can have depended on it.

## Risks & Mitigations

- Risk: something depends on the old `0x22` constant.
  - Mitigation: nothing can. No machine emits or accepts `0x22`, and the value was only ever "correct" because the firmware clamped it. The change is deliberately ungated for that reason.

### DIFF
--- a/lib/src/models/device/impl/bengle/bengle.dart
+++ b/lib/src/models/device/impl/bengle/bengle.dart
@@ -25,7 +25,7 @@ class Bengle extends UnifiedDe1
   Future<double> getCupWarmerTemperature() =>
       readMmrScaled(BengleMmr.matSetPoint);
 
-  /// Bengle FW requires entering state 0x22 (`MachineState.fwUpgrade`) between
+  /// Bengle FW requires entering state 0x16 (`MachineState.fwUpgrade`) between
   /// the `requestState(sleeping)` step and the start of `.dat` upload.
   /// DE1 doesn't need this — see [UnifiedDe1.beforeFirmwareUpload]
   /// for the hook contract.
--- a/lib/src/models/device/impl/de1/de1.models.dart
+++ b/lib/src/models/device/impl/de1/de1.models.dart
@@ -73,7 +73,7 @@ enum De1StateEnum {
   inBootLoader(0x13), // Bootloader is active, firmware has not run
   airPurge(0x14), // Air purge
   schedIdle(0x15), // Scheduled wake-up idle state
-  fwUpgrade(0x22),
+  fwUpgrade(0x16), // FirmwareUp (corrected from 0x22, which was never a wire value)
   unknown(-1); // Default or unknown state
 
   final int hexValue;
--- a/lib/src/models/device/impl/de1/unified_de1/unified_de1.dart
+++ b/lib/src/models/device/impl/de1/unified_de1/unified_de1.dart
@@ -611,7 +611,7 @@ class UnifiedDe1 implements De1Interface {
   /// `requestState(sleeping)` and the start of FW image upload.
   ///
   /// Default: no-op. Bengle overrides this to request
-  /// `MachineState.fwUpgrade` (state 0x22) — Bengle FW requires entering
+  /// `MachineState.fwUpgrade` (state 0x16) — Bengle FW requires entering
   /// that state before the `.dat` upload protocol starts. DE1 doesn't.
   ///
   /// This resolves the TODO at `unified_de1.firmware.dart:13-14` (commented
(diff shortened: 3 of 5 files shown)


tadelv:
I rebased the branch to latest main instead of merge and relaxed the `0x22 => .unknown` mapping test so that we don't have to remove it later when we add additional states. Therefore some of the initial description of the PR will be incorrect.

tadelv (approved):

PR #458, opened 2026-07-16
fix(de1): serialize MMR access behind an in-progress firmware upload

ChampionDesigns:
A firmware update over USB serial can land a corrupt image on the machine even
though the transfer reports success. The bootloader then rejects the staged
image on the next boot with "Header is broken" and quietly keeps the old
firmware, so the machine comes back on the version it started on and the user is
left wondering why the update did not take. The cause is that the firmware
upload and ordinary register access share one wire, and nothing stopped them
from talking over each other. This PR gives the upload exclusive use of that
wire for the duration of the transfer.

## Summary

- Problem: firmware upload chunks and ordinary MMR reads/writes share the DE1
  MMR tunnel. Interleaving ordinary traffic into the upload sequence corrupts
  the staged image; the bootloader then rejects it and keeps the old firmware.
- Why it matters: reconnect-time config reads stay active for seconds over the
  slower half-duplex USB serial transport, making a collision routine rather
  than theoretical.
- What changed: `_FirmwareMmrGate` is a writer-preferred gate. Firmware waits
  for already-admitted MMR operations to drain, later MMR calls queue once
  firmware is waiting, and each MMR permit spans the complete logical operation,
  including response subscription, retries, retry-settle delays, and transport
  completion.
- Firmware exclusivity is acquired inside `_updateFirmware()` after
  `requestState(MachineState.sleeping)` and `beforeFirmwareUpload()`, then held
  through erase, upload, pacing, and the final firmware-map request. Scoped
  `finally` handling releases queued work on success or failure.
- Cancellation is rechecked immediately after exclusive ownership is acquired,
  before the firmware-map subscription or erase write. A cancellation that
  occurs while firmware is waiting for active MMR work therefore cannot erase
  firmware.
- Scope boundary: this PR adds wire exclusion only. It does not add firmware
  busy state, public lifecycle state, or a cancellation abstraction; PR #471
  remains the owner of those concerns.
## Change Type (select all)

- [x] Bug fix
- [ ] Feature
- [ ] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [ ] REST API / handlers
- [ ] WebSocket API
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [ ] Docs / specs

## Linked Issues

- Closes #
- Related #

No tracking issue was filed. This was found on the bench: repeated firmware
updates over USB serial reported "done" but the machine stayed on the old build,
and the machine rejects the corrupted image. The image files
were verified good (their own header checksums are valid), which pointed at
corruption in transit rather than a bad file.

## Root Cause (if bug fix)

- Root cause: the upload and MMR access are two users of one half-duplex tunnel,
  with no mutual exclusion. `uploadFW` sends the image as a run of
  `writeToMMR` writes that the machine's firmware-receive state machine expects
  to arrive uninterrupted; an MMR read request (a write to `readFromMMR`) or an
  MMR write inserted between chunks breaks that run and scrambles the staged
  image.
- Missing detection or guardrail: nothing serialized transport access against
  the upload. The upload's own `writeWithResponse` calls complete normally, so
  the app reports success while the machine has actually received a corrupt
  image; the failure only surfaces on the machine at boot.
- Contributing context: the collision is transport-speed-sensitive. It was
  effectively invisible over BLE and became routine over USB serial, where the
  config read that runs on every reconnect keeps the tunnel busy for seconds.

  - [x] Unit test
- Target test: `test/unit/models/device/impl/de1/unified_de1/firmware_mmr_exclusion_test.dart`.
- The deterministic `UnifiedDe1` fake-transport suite covers:
  1. firmware waiting for an MMR read already in flight;
  2. writer preference over MMR calls arriving after firmware waits;
  3. cancellation while waiting for ownership terminating before erase;
  4. queued MMR release after firmware failure;
  5. one MMR permit spanning response timeout, retry, and retry settling; and
  6. MMR traffic issued during firmware reaching the wire only after the final
     firmware-map request and upload chunk.
- Tests use operation-specific completer barriers around transport writes and
  notifications rather than Bengle preparation timing or an erase-delay seam.

The gate is acquired after firmware preparation rather than at the top of
`_updateFirmware()`. `requestState` and `beforeFirmwareUpload` do not use the MMR
tunnel, so they remain outside the corruption-sensitive exclusive section.

Cancellation is checked as the first statement inside the exclusive operation,
not only before calling `runFirmwareExclusive()`. Gate acquisition is an
asynchronous boundary: cancellation may happen while earlier MMR work drains,
and no erase request may be sent afterward.

The gate is writer-preferred. Ordinary MMR operations can run concurrently while
no firmware writer is active or waiting; once firmware queues, later MMR calls
cannot jump ahead. Each queued acquisition has its own completer, and the scoped
`runMmr` / `runFirmwareExclusive` helpers release in `finally`.

MMR callers wait rather than fail fast during firmware work. This keeps the
change internal to the wire owner and avoids adding a competing firmware
lifecycle or busy-state surface before PR #471. The old mutable `_fwTunnelLock`,
outer-finally release, and `firmwareEraseSettle` test seam have been removed.

- [ ] API spec updated
- [ ] API docs updated
- [ ] Plugin docs updated
- [ ] Skin docs updated
- [ ] Profile docs updated
- [ ] Device docs updated
- [x] N/A - no docs affected

No public API, wire protocol, or config surface changes. The behavior change is
internal serialization that only takes effect during a firmware upload.

## Security Impact (required)

- BLE/USB surface changed? `No`. Same endpoints and frames; only the ordering of
  MMR access relative to a firmware upload changed.

If anything, this narrows a corruption window: an interrupted or interleaved
firmware image can no longer be produced by concurrent register access.

## User-Visible Changes

A firmware update started over USB serial while the app is busy reading machine
config no longer corrupts the staged image. Updates apply reliably instead of
sometimes reporting success and silently leaving the machine on the old build.
No UI, config, or default changes.

- [x] `flutter analyze` - clean, `No issues found!` on the changed files.
- [x] `flutter test` - the new test passes; existing MMR, firmware, and Bengle
  device tests (`unified_de1_mmr_test`, `firmware_upload_tuning_test`,
  `bengle_cup_warmer_test`, `bengle_saw_test`) still pass.
- [ ] `(cd packages/dye2-plugin && npm run build)` - not affected by this change.

- OS / platform tested: the corruption was reproduced on an Android tablet
  flashing a Bengle over the USB serial link.
- Real hardware? (DE1/Bengle/scale): `Yes` for the bug. Repeated updates over
  USB reported success but the machine stayed on the old build and the update
  client printed "Header is broken"; flashing the same image when the tunnel was
  quiet applied cleanly. The lock enforces that quiet condition programmatically.
- What you did NOT verify: the fixed build has not yet been rebuilt and re-run
  through the failing on-hardware reproduction; the fix is covered by the
  interleave test and by the root-cause analysis of the on-wire collision.

### Evidence

- [x] Test output (failing before, passing after)
- [ ] Log snippets

## Compatibility & Migration

- Backward compatible? `Yes`. No API, wire, or config changes. A machine that is
  not mid-upload behaves exactly as before.

## Risks & Mitigations

- Risk: a failed firmware operation could leave queued MMR calls blocked.
  - Mitigation: `runFirmwareExclusive` releases ownership in its scoped
    `finally`; deterministic failure coverage proves a queued read resumes.
- Risk: a firmware writer could begin while an earlier MMR read is still waiting
  for a response or retry.
  - Mitigation: `runMmr` owns one permit across the complete read/retry future,
    and tests cover both an in-flight response and retry-settle lifecycle.
- Risk: cancellation while waiting for exclusivity could erase firmware after
  the user already cancelled.
  - Mitigation: cancellation is rechecked immediately after acquisition and
    before subscription/erase; a barrier test asserts zero `fwMapRequest` writes.
- Risk: writer preference delays ordinary MMR traffic for the upload duration.
  - Mitigation: this is intentional; allowing later MMR work ahead of firmware
    would reintroduce starvation and wire corruption. Queued MMR work resumes on
    every scoped exit.
- Risk: PR #471 could introduce a second owner during merge conflict resolution.
  - Mitigation: retain PR #471's lifecycle/state/cancellation owner and preserve
    one gate wrapper around its erase → upload → verification sequence, with its
    typed cancellation check first inside the wrapper.


tadelv (commented):
High — the current `_fwTunnelLock` is only a gate for MMR calls that begin after the flag is set; it does not establish exclusive ownership against an MMR operation already in flight. This needs to be corrected before merge, and the implementation should be shaped so the subsequent #471 merge keeps one clear owner for firmware lifecycle/concurrency.

## Required changes

### 1. Keep #458 scoped to MMR/firmware wire exclusion

#471 introduces the authoritative firmware-operation lifecycle in `UnifiedDe1`: synchronous reservation, `FirmwareUpdateState`, typed cancellation, rejection of concurrent firmware calls, and identity-safe cleanup. #458 should not create a competing lifecycle.

Remove from #458:

- `_fwTunnelLock` as a mutable firmware-active flag
- releasing that flag from `updateFirmware()`'s outer `finally`
- `firmwareEraseSettle`
- any firmware busy/concurrency/cancellation state beyond the exclusion primitive

The exclusion primitive should have its own acquisition/release lifecycle inside `_updateFirmware()`, independent of the public `updateFirmware()` wrapper that #471 replaces.

### 2. Replace the `Completer` flag with a real writer-preferred gate

Add a small private gate/helper, preferably isolated from the firmware-operation fields that #471 edits. It needs two conceptual operations:

```dart
Future<T> runMmr<T>(Future<T> Function() operation);
Future<T> runFirmwareExclusive<T>(Future<T> Function() operation);
```

Required semantics:

- ordinary MMR operations may be admitted while no firmware writer is active or waiting;
- firmware acquisition stops admission of new MMR operations;
- firmware waits for already-admitted MMR operations to drain;
- once drained, firmware owns the shared tunnel exclusively;
- new MMR operations remain queued until firmware releases;
- once firmware is waiting, later MMR calls must not jump ahead of it;
- acquisition must return an operation/lease-specific future rather than relying on one globally replaceable `Completer`.

A single async mutex around every complete MMR operation and the firmware critical section is also acceptable if the broader serialization is intentional and tested. The important requirement is that firmware cannot begin until earlier MMR work has finished.

### 3. Hold the MMR permit across the entire logical operation

Wrap all of `_mmrReadRaw()`, including:

- response subscription;
- request writes;
- response waiting;
- all retries;
- retry-settle delays;
- timeout/error cleanup.

Do not check the firmware flag only once before entering the retry loop. A read admitted before firmware starts can currently emit its first request or a later retry after the upload has begun.

Similarly, `_mmrWriteRaw()` must hold its permit until `writeWithResponse()` completes.

Conceptually:

```dart
return _firmwareMmrGate.runMmr(() async {
  // complete read operation, including retries
});
```

### 4. Acquire firmware exclusivity after prep and release it locally

Keep these outside the exclusive section while they remain non-MMR operations:

```dart
await requestState(MachineState.sleeping);
await beforeFirmwareUpload();
```

Then acquire exclusive ownership before the erase request and hold it through the complete corruption-sensitive sequence.

For the current branch that means erase request/delay, upload, and final firmware-map request. After #471 is merged/rebased, the exclusive section must cover:

1. erase request;
2. erase-complete response wait;
3. all upload writes and pacing;
4. verification request;
5. final verification response/timeout.

Release in a local `finally`/scoped helper, not from the outer `updateFirmware()` cleanup.

Conceptually after #471:

```dart
await _firmwareMmrGate.runFirmwareExclusive(() async {
  _firmwareUpdateState = FirmwareUpdateState.erasing;
  await requestAndAwaitErase(token);

  _firmwareUpdateState = FirmwareUpdateState.uploading;
  await _uploadFirmwareBytes(image, onProgress, token);

  _firmwareUpdateState = FirmwareUpdateState.verifying;
  await requestAndAwaitVerification(token);
});
```

### 5. Do not duplicate #471's operation owner

Do not add a second firmware-active boolean, busy exception, public state, or cancellation-token abstraction in #458. #471 already provides the one atomic machine-level owner and synchronously rejects overlapping updates.

The MMR gate may defensively serialize exclusive acquisitions, but it should not become the public firmware concurrency coordinator.

### 6. Replace the timer-based regression test with deterministic barriers

The current test depends on Bengle prep timing, an 800 ms sleep, and the `firmwareEraseSettle` test seam. #471 removes the fixed erase wait and waits for an actual firmware-map response, so this test will become brittle/obsolete.

Use `UnifiedDe1` plus a fake transport that can pause writes or notifications. Cover at least:

1. **MMR already active before firmware acquisition:** hold the MMR response, start firmware, and assert the firmware erase request does not reach the wire until that MMR operation completes.
2. **MMR starts while firmware owns/waits for the tunnel:** assert no MMR request reaches the wire until firmware completes.
3. **MMR read including retry lifecycle:** ensure the permit covers retries and retry delays, not only the first method-entry check.
4. **Firmware failure/cancellation:** queued MMR operations are released afterward.
5. Preserve the existing ordering assertion that an MMR request issued during firmware lands only after the final firmware tunnel traffic.

Suggested location, matching #471's tests:

```text
test/unit/models/device/impl/de1/unified_de1/firmware_mmr_exclusion_test.dart
```

## Expected merge boundary with #471

After these changes, #471 should retain its firmware state machine, synchronous reservation, cancellation token, response-driven erase, and verified completion. #458 contributes only the separate MMR exclusion gate plus the scoped exclusive wrapper around erase/upload/verification.

This should also reduce the future conflict surface:

- `unified_de1.mmr.dart`: retain #458's complete-operation gate wrappers;
- `unified_de1.dart`: avoid changing `updateFirmware()`/`_fwCancelToken`, which #471 rewrites;
- `unified_de1.firmware.dart`: resolve by preserving one exclusive wrapper around #471's new erase → upload → verify sequence.

tadelv (approved):

PR #459, opened 2026-07-16
feat(bengle): model-authoritative machine class, MTU 517, and a firmware<->app MMR contract

ChampionDesigns:
> **Bengle machine support — stack PR B-1 of 10 (foundation).** PRs B-2–B-10 build on this branch; please review/merge the stack in order **B-1 → B-10**.

## What a Bengle is, and why any of this is in your repo

A Bengle is a Decent-hardware variant: a DE1 superset. It runs the same protocol, answers the same
characteristics, and is driven by the same profiles, but it also carries an integrated scale built
into the drip tray (real load cells, read by the machine's own firmware), addressable LED strips, a
cup warmer, a milk probe, and a USB-serial transport alongside BLE. It identifies itself over the
wire: the `v13Model` MMR register (`0x0080000C`) reads `1..7` on the DE1 family and `>= 128` on a
Bengle. Everything in this stack is gated on that number, so on a machine that reads `1..7` not one
line of Bengle code runs.

Today the app half-knows this. `upstream/main` already contains a Bengle base (`bengle.dart`,
`bengle_mmr.dart`, `bengle_interface.dart`, `MockBengle`, three bridges, 36 files in all) and it
already logs a warning when it sees `v13Model >= 128`. What it does not do is act on it. The machine
class is picked from the advertised BLE name before a connection exists, so a Bengle whose
advertised name does not happen to say "Bengle" connects as a plain `UnifiedDe1` and every Bengle
feature stays dark for the whole session. This PR makes the model number authoritative, fixes an
MMR write that truncates a whole unit low, raises the ATT MTU so a 28-byte notification is not
silently truncated on iOS and macOS, and introduces a machine-readable firmware/app register
contract with a CI checker so the register tables in this repo and the register tables in the chip
cannot drift apart in silence again.

## Summary

- **Problem:** The machine class is chosen from the advertised BLE name, but the authoritative
  identity is the `v13Model` MMR, which cannot be read until after the connect. A Bengle advertising
  a DE1-style name therefore lands as a plain `UnifiedDe1` with every Bengle capability dark, and a
  DE1 mis-advertising "Bengle" would be driven with the wrong protocol. Separately, the `@protected`
  `writeMmrScaled` helper integerized with `toInt()`, which truncates: IEEE-754 makes `2.3 * 100`
  equal `229.999...`, so a 2.30 g target reached the wire as `229`, a whole centigram low.
- **Why it matters:** Getting the class wrong is not a degraded experience, it is a silently absent
  one, and the user has no way to tell. The truncation quietly biases every scaled write a Bengle
  capability makes. And the MMR register layout is hand-declared twice, once in the firmware's
  the firmware register table X-macro table and once in this repo's Dart enums, in two languages, with nothing
  comparing them.
- **What changed:** `UnifiedDe1` gains an `isBengle` flag read from `v13Model` during `onConnect`;
  a new pure resolver (`resolveMachineForModel`) rebuilds the machine as the correct class over the
  *same live transport* when the name-picked class disagrees, and `De1Controller.connectToDe1`
  calls it and tears the discarded interim down. `writeMmrScaled` now rounds. The BLE transport
  requests MTU 517 on every platform except Linux. A new `assets/api/bengle_hw_v1.yml` states the
  register contract, and a Dart test asserts every app-declared register against it on every CI run.
- **What did NOT change (scope boundary):** A DE1 reading `v13Model` 1 through 7 leaves `isBengle`
  false, the resolver returns the same instance untouched, and the base-DE1 private `_writeMMRScaled`
  (the flush, hot-water, steam, heater and calibration-flow setters) deliberately keeps `toInt()`,
  because de1plus truncates exactly those and rounding them would change bytes on shipped DE1
  hardware. No REST or WebSocket shape changes.

## Change Type (select all)

- [x] Bug fix
- [x] Feature
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [ ] REST API / handlers
- [ ] WebSocket API
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

- Related # `N/A`

## Root Cause (if bug fix)

- **Root cause (class resolution):** discovery has to instantiate a machine class before a
  connection exists, so it uses the only thing it has, the advertised name. The authoritative
  identity lives behind the connection. Nothing ever reconciled the two afterwards. The serial path
  already class-dispatched on `v13Model >= 128`; the BLE path never did.
- **Root cause (model-name mapping vs the gate):** `DecentMachineModel.fromInt` mapped only the
  *exact* value `128` to `Bengle` (everything else fell through to `Unknown`), while every detection
  site gates on `v13Model >= 128`. A future Bengle model number would therefore pass the gate but
  display as "Unknown". This branch aligns the two - `fromInt` now maps `>= 128` to `Bengle` - so the
  name mapping and the detection gate agree.
- **Root cause (scaled write):** `(value * addr.writeScale).toInt()` truncates a binary
  floating-point product that lands a hair below the integer, and the write class that needed
  rounding shared a helper name with the write class that needed truncation.
- **Missing detection or guardrail:** two hand-maintained register tables in two languages with no
  comparison between them. `MMRItem.steamStartSecs` shipped with the default scales of 1.0 against a
  firmware `mult` of 100. It is latent today (nothing reads or writes it), so no bytes change, but
  the first setter wired to it would have written 100 times low, and nothing in the repo could have
  caught that. The new contract checker fails on exactly this class of drift.

  - [x] Unit test
- **Target test or file:** `test/unit/models/device/impl/de1/unified_de1/protected_surface_test.dart`
  (rounding), `test/models/device/de1_resolver_test.dart` and
  `test/controllers/de1_controller_resolve_test.dart` (class resolution),
  `test/unit/models/device/impl/bengle/mmr_contract_test.dart` (drift).
- **Scenario the test should lock in:** Both directions of the scaled-write split are pinned so
  neither can be "unified" away later: `setSteamFlow(2.3)` must put `229` on the wire while a
  capability scaled write of `2.3` at x100 must put `230`. The resolver tests cover promote, demote,
  no-swap, identity carry-over and detach safety; the controller test covers the promote path, the
  `deviceId` idempotency guard, and the demotion teardown (it fails if any capability dispose is
  removed). The contract checker asserts address, length and scale exactly, and range as an
  app-subset-of-contract, for all 30 shared DE1 registers.

- [ ] API spec updated: `assets/api/rest_v1.yml` or `assets/api/websocket_v1.yml`
- [x] Device docs updated: `doc/DeviceManagement.md` - new "Bengle: name is a hint, v13Model is
      authoritative" section
- [x] Other: `assets/api/bengle_hw_v1.yml` (new, 865 lines) and the contract file header (new,
      274 lines)

`rest_v1.yml` gets one small correction with no behaviour change: the `simulatedDevices` enum in both
schemas still listed only `[machine, scale, sensor]`, while the shipped settings handler has
accepted and persisted `bengle` since `MockBengle` landed. A client following the spec could not
discover a value the app already takes.

`assets/api/bengle_hw_v1.yml` is the new one to look at. It is one row per MMR register (address,
length, permissions, multiplier, kind, range, semantics), distilled from the firmware the firmware register table at a
named, hardware-validated firmware commit - a provisional dev-build reference while the firmware is
still in active development, to be finalized to a released or tagged reference when the firmware is cut
as a release - plus the `0xA013` packet layout and the ASCII serial verb
contract as human-readable sections. the contract file header is the coordination protocol around
it: change the firmware register table, regenerate the contract, bump `contract_version`, update the enums, and both
PRs cite the version. The contract file lives in this repo because the consumer and the CI live
here; the layout authority stays in firmware, because the chip decides.

## Security Impact (required)

- BLE/USB surface changed? `Yes` - the post-connect MTU request now runs on all non-Linux platforms
  rather than Android only, and a connect can now rebuild the machine object over the same
  already-authenticated transport. Neither widens what the app can reach: the MTU request is a
  negotiation against a device the user already connected to, and a failed negotiation is logged and
  ignored rather than aborting the connect. The re-resolve uses the transport that is already open.

## User-Visible Changes

A Bengle is recognised as a Bengle regardless of what name it advertises, so its features are
available in the session rather than dark. On a plain DE1 the only observable difference is a
possibly larger BLE MTU, which reduces GATT round-trips. No config, default, or API changes.

- [x] `flutter test` - **2041 tests pass** at this branch head, against a 1986-test baseline on
      `upstream/main` @ `1950f7b1`. The `+55` confirms the branch's own tests actually ran.

Gates were run in a scratch worktree on a detached HEAD, exactly as `.github/workflows/pr-checks.yml`
does, including the `bundle_skins.sh` step (`assets/bundled_skins/` is gitignored and
`webui_storage_bundled_test.dart` fails without it on plain `upstream/main` too).

- Real hardware? (DE1/Bengle/scale): `No` - not for this branch, by me, in this form.
- What you personally verified and how: the full local gate set above, plus reading each diff hunk
  against the firmware register table it claims to encode.
- Edge cases checked (by test, not by hand): `v13Model` exactly at the 128 boundary; a Bengle
  demoted to a DE1 (the discarded interim's capabilities are disposed, or it leaks its subjects);
  re-passing the same physical machine after a class swap (the idempotency guard keys on `deviceId`,
  because after a swap the object identity has changed but the machine has not).
- What you did **not** verify: I did not personally run this branch against a machine. The
  underlying behaviour was exercised on real Bengle hardware on the working branch this code was
  sliced from; the slice itself has only been verified by the test suite. The MTU 517 request is
  verified by unit test against a fake platform shim, **not** on an iOS or macOS device - I have not
  observed a real ATT negotiation on either.

### Evidence

- [x] Test output (2041 pass; the resolver, rounding and contract tests are new on this branch)
- [ ] Log snippets

## Compatibility & Migration

- Backward compatible? `Yes`

`MachineSnapshot`, the REST surface and the WebSocket surface are untouched. A DE1 that reads
`v13Model` 1 through 7 takes exactly the code path it takes today.

1. **Rebuilding the machine object over a live transport.** When the model disagrees with the
   name-picked class, the resolver constructs a fresh `Bengle` (or `UnifiedDe1`) over the *same*
   `DataTransport` and copies the connect-time identity across (`adoptIdentityFrom`), so the second
   `onConnect` short-circuits the MMR re-reads and only re-subscribes and runs capability init. The
   alternative (tear the connection down and reconnect) is simpler to reason about but doubles the
   connect time and is a visible reconnect to every client. I chose the rebuild; it needs three new
   seams on `UnifiedDe1` (`dataTransport`, `adoptIdentityFrom`, `detachTransport`) and I would rather
   you told me now if you want those seams shaped differently.
2. **`detach()` versus `dispose()` on the transport wrapper.** The discarded interim must stop
   listening (otherwise a lingering serial `readStream` listener double-parses every line) but must
   not dispose the underlying transport, which the replacement now owns. That is a genuinely new
   lifecycle verb on `UnifiedDe1Transport` and the one piece of this PR most likely to bite someone
   who later adds a subject and forgets to close it in both places.
3. **Two scaled-MMR writers that behave differently on purpose.** `writeMmrScaled` rounds,
   `_writeMMRScaled` truncates. This looks exactly like the sort of duplication a reviewer asks to
   merge. It cannot be merged: de1plus truncates the base-DE1 setters and rounds the write class the
   Bengle capabilities use, and unifying either way changes bytes on somebody's hardware. Both
   directions are test-pinned, and the comments say why.

## Risks & Mitigations

- **Risk:** A capability added to `Bengle.onConnect` later, whose subjects are not disposed in
  `De1Controller`'s demotion teardown, leaks on the discarded interim.
  - **Mitigation:** the teardown is deliberately exhaustive today (integrated scale, LED strip), and
    `de1_controller_resolve_test` fails if any single dispose call is removed. The comment at the
    call site says to extend it. This is a real maintenance edge and I would rather name it than
    pretend the test closes it forever.
- **Risk:** MTU 517 is now requested on platforms where it was never requested before, and a stack
  could reject it.
  - **Mitigation:** rejection is non-fatal by design (log and continue). The DE1/Bengle BLE module
    self-negotiates up to 247 on connect regardless, so this client request is belt and braces, not
    a dependency. Linux is skipped because BlueZ manages the MTU itself and `universal_ble` does not
    expose `requestMtu` there.
- **Risk:** the `bengle_hw_v1.yml` contract goes stale against firmware and the checker starts
  asserting a fiction.
  - **Mitigation:** `contract_version` is pinned by the test, and the change protocol in
    the contract file header requires the firmware PR and the app PR to cite the same version. It
    is a coordination discipline, not an automated guarantee, and I want to be honest that it is
    only as good as the discipline.

### DIFF
--- a/assets/api/bengle_hw_v1.yml
+++ b/assets/api/bengle_hw_v1.yml
@@ -1,0 +1,757 @@
+# =============================================================================
+# Bengle firmware <-> app hardware contract (bengle_hw_v1.yml)
+# =============================================================================
+# contract_version: 1
+#
+# Generated from firmware the firmware register table at:
+#
+#
+# Change protocol: firmware changes a register -> regenerate this file from the
+# new the firmware register table -> bump contract_version -> update the app enums (BengleMmr /
+# BengleSteamMmr / BengleScaleMmr / BengleCalMmr / BengleLedMmr / MMRItem) ->
+# the mmr_contract_test checker enforces agreement; firmware and app PRs both
+# cite the contract_version.
+#
+# Normalization rules applied (binding):
+#   * mult / min / max are only meaningful on firmware register-table rows with auto > 0 (the
+#     macro read/write path). auto = 0 rows are served by hand-written firmware
+#     cases; for those, `mult` below records the VERIFIED behaviour of the
+#     hand-written path (which matches the the firmware register table column everywhere except
+#     v13Model, whose mult=1000 column is inert -- recorded as mult: 1).
+#   * min / max below are RAW WIRE UNITS (the little-endian int32 actually on
+#     the wire). Where the firmware register table used post-divide engineering units, the original
+#     engineering values are kept in a trailing comment.
+#   * min/max encoding: where the firmware register table declares the full-u32 column
+#     0..0xFFFFFFFF it is retained verbatim (for a u32 wire value that is
+#     identical to "no constraint"); ScaleTare alone is normalized to
+#     null/null because its written value is ignored entirely (pure trigger).
+#     Deliberate, not drift -- a subset checker passes either encoding.
+#   * Read-back resolution on FLOAT-backed rows (EndOfShotWeight, LastTARE,
+#     ScaleCalWeight, TargetMilkTemp, V24CurrentBudget, MatCurrentTemp): the
+#     firmware macro read path truncates BEFORE scaling
+#     (the firmware the firmware read path),
+#     so wire READS return floor(engineering) x mult -- whole engineering
+#     units -- while writes keep full x-mult resolution. The same U32(float)
+#     cast saturates a negative stored value to 0 on Cortex-M, so these rows
+#     can never read back negative. Affected rows carry a "Read-back" note.
+#     (MatSetPoint is float-backed too but exact: mult 1, whole-degC writes.)
+#   * The firmware does NOT enforce min/max on Bengle macro-path writes
+#     the firmware write path never range-checks): bounds here are documentation and the app
+#     is the sole guard. Checker semantics: app range must be a SUBSET of the
+#     contract range (not equal) -- e.g. calFlowEst app min raw 130 vs FW 125.
+#   * perms are canonicalized from the the firmware register table ENTRY column:
+#     PERM_READ -> R, PERM_RT -> RT, PERM_RW -> RW, PERM_RWD -> RWD (persisted
+#     to disk), PERM_RWT -> RWT (write triggers an action). The ENTRY column is
+#     AUTHORITATIVE: the firmware register table's own header comment block disagrees with it on
+#     row 33 (comment "RW" vs ENTRY PERM_READ) and rows 34/35/37 (comment "W"
+#     vs ENTRY PERM_RWT); the firmware write macro verifiably enforces the
+#     ENTRY perms. App enums carry no perms today, so the checker must NOT
+#     assert perms in v1.
+#   * `app_alias` is present only where the app's declared description string
+#     differs from the firmware register name (TargetMilkTemp is the only
+#     case). Rows are keyed by FIRMWARE name, so the alias records the
+#     app-side name.
+#   * Rows 12 (PrefGHCMCI) and 13 (MaxShotPres) are omitted: PERM_NONE TODO
+#     stubs in firmware, unreachable and irrelevant to the app.
+#   * All len=4 registers are little-endian int32 on the MMR wire. Addresses in
+#     frames are 24-bit big-endian (the 0x00 top byte of 0x008038xx-style
+#     addresses is discarded).
+# =============================================================================
+
+contract_version: 1
+firmware:
+  commit: 0381e7ab58eb5b5ee36c14b0bef123ea3cfe4f2e
+  build: 90
+  source: the firmware register table
+generated: 2026-07-11
+
+# -----------------------------------------------------------------------------
+# MMR register table
+# name = firmware register name; fw_row = the firmware register table RangeNum; address as it
+# appears in a read/write frame (before the 24-bit truncation); length in
+# bytes; mult = raw wire value / engineering value.
+# -----------------------------------------------------------------------------
+registers:
+  # --- Shared DE1-family rows the app's MMRItem enum declares -----------------
+  - name: ExternalFlash
+    fw_row: 0
+    address: "0x00000000"
+    length: 1048575 # 0xFFFFF -- first 1 MB mapped to external flash
+    perms: RW
+    mult: 1
+    value_kind: bytes
+    min: null
+    max: null
+    unit: null
+    semantics: "External-flash window (disk operations, firmware images); not a scalar register."
+
+  - name: HWConfig
+    fw_row: 1
+    address: "0x00800000"
+    length: 4
+    perms: R
+    mult: 1
+    value_kind: int32
+    min: null
+    max: null
+    unit: bitmask
+    semantics: "Hardware configuration word."
+
+  - name: Model
+    fw_row: 2
+    address: "0x00800004"
+    length: 4
+    perms: R
+    mult: 1
+    value_kind: int32
+    min: null
+    max: null
+    unit: null
+    semantics: "Legacy machine model."
+
+  - name: CPUBoardModel
+    fw_row: 19
+    address: "0x00800008"
+    length: 4
+    perms: R
+    mult: 1
+    value_kind: int32
+    min: null
+    max: null
+    unit: "model x 1000"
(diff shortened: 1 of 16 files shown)


tadelv (changes requested):
Closing this PR because its central post-connect class-resolution architecture is based on a development-hardware anomaly rather than the production BLE identity contract. Production Bengle advertises as `Bengle`, so we do not want provisional machine objects, live transport transfer/detach, a second `onConnect()`, or controller-side class replacement.

The useful independent changes will be harvested into focused replacement PRs from current `main`: `v13Model >= 128` model mapping/guards, verified MMR corrections, the simulated-Bengle REST schema fix, an executable contract guardrail if justified, and non-Android MTU negotiation behind an experimental flag.

PR #460, opened 2026-07-16
feat(serial): DE1-family USB-serial parity, and let a wired-only setup reach its machine

ChampionDesigns:
> **Stacked PR — B-2 of 10.** Builds on #459 (`feat/bengle-foundation`). Until that merges this PR's diff includes its commits; please review/merge in order **B-1 → B-10**.

## The story

Plug a machine in over USB, turn Bluetooth off, and the app shows you a full-screen "Connection
error: Bluetooth is turned off." and nothing else. USB discovery is running perfectly well behind
that screen and has almost certainly already found your machine, but you cannot see it and you
cannot get to the picker, so a wired-only setup simply cannot be used. That was reproduced on the
bench, and it is two separate gates, not one: fixing either alone leaves the picker hidden behind the
other. On top of that, USB-serial is not at parity with BLE in the shared transport. A firmware
upload desyncs the framing on its last chunk, several reads are outright `UnimplementedError`, and a
USB client that just sits and listens gets its notification stream silently stolen by the BLE module,
because both share one serial view in the firmware arbitrated by a last-writer-wins flag. This PR
closes all of that.

## Summary

- **Problem:** (a) Two independent gates in the scan flow (the guardian's adapter-error view, and
  the connection manager's sticky `adapterOff` error) hid a working USB discovery behind a Bluetooth
  error screen. (b) The Android USB pre-filter dropped any port whose product name was not `DE1`,
  `Half Decent Scale`, or something containing `Serial`, before the class shortcut or the `v13Model`
  probe ever ran, which made a real Bengle undetectable over USB on Android. (c) Three serial-only
  correctness gaps in the shared transport: short `<F>` frames desync the firmware parser, several
  reads throw `UnimplementedError`, and a passive USB listener loses its notify stream to the BLE
  module.
- **Why it matters:** (a) is a dead end for a wired-only user. (b) means the wire works and the app
  never looks: current Bengle firmware enumerates with the pico-sdk **default** TinyUSB descriptors,
  VID:PID `0x2E8A:0x000A` and product string "TinyUSB Device" (captured from hardware 2026-07-10), so
  it passes neither the name check nor the (empty) direct-ID table. (c) turns a firmware upload into
  a desync and a raw-API read into a crash, and the stolen-stream case is the nasty one: the client
  is not disconnected, it just stops being told anything.
- **What changed:** Both scan gates are demoted by a single `busyWithoutBle` predicate; the Android
  probe gate is now an OR of a name predicate and a VID:PID probe-candidate predicate; short
  `writeToMMR` frames are zero-padded to the firmware's exact 20-byte frame size; serial reads gain
  a real three-shape implementation (subscribed / single-notify round trip / honestly unsupported);
  and a 5-second `<+N>` keepalive re-asserts the USB source on the shared link.
- **What did NOT change (scope boundary):** Every path added or altered here is gated on
  `transportType == TransportType.serial`. The BLE path is byte-for-byte unchanged. Every product
  name the Android probe admitted before still passes. Only the `adapterOff` error kind is demoted in
  the scan flow: a genuine `machineConnectFailed` with machines on screen still shows the error view.

## Change Type (select all)

- [x] Bug fix
- [x] Feature
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [x] BLE transport / device comms (the USB/serial half of it)
- [ ] REST API / handlers
- [ ] WebSocket API
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [x] UI / Flutter widgets (`scan_flow_view.dart`)
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

- Related # `N/A`

## Root Cause (if bug fix)

- **Root cause (scan flow):** the adapter-error view was written when Bluetooth was the only
  transport, so "the adapter is off" and "we cannot find anything" were the same statement. Serial
  discovery made them different statements, and the view was never revisited. The device scan already
  runs every discovery service in parallel and records per-service failures; only the presentation
  was wrong.
- **Root cause (Android probe):** the name check was an inline three-way condition written against
  the descriptors the DE1 ships. The Bengle shortcut immediately below it (`productName == "Bengle"`)
  was therefore dead code, because the gate above rejected the port first. The firmware never got
  around to setting a product string.
- **Root cause (serial parity):** the serial view was implemented for the parts that a DE1 exercises
  and stubbed for the rest. The `UnimplementedError` reads and the unpadded `<F>` frame are that
  stub showing.
- **Missing detection or guardrail:** `bengleUsbIds` being empty was documented but not tested, so
  somebody "completing" the table later would have changed detection semantics with every existing
  test still green. That is now pinned.

  - [x] Unit test
- **Target test or file:** `test/unit/models/device/impl/de1/unified_de1/serial_parity_test.dart`
  (new, 418 lines), `test/device_discovery_feature/scan_flow_ble_off_test.dart` (new),
  `test/controllers/connection_manager_wired_preferred_test.dart` (new),
  `test/unit/services/serial/serial_probe_name_gate_test.dart` (new, 13 tests),
  `test/services/serial/usb_ids_test.dart`, plus a new inbound-capable `FakeSerialTransport` helper.
- **Scenario the test should lock in:** With the adapter reporting off, a machine arriving over
  serial must reach the picker (both the guardian view and the sticky-error view are asserted, since
  demoting only one leaves the bug). The probe gate is tested as the **OR combination** at the call
  site, not just as two predicates, because the OR is the actual fix. On the transport: a short
  `<F>` frame is padded to 20 bytes and the true length still rides the `Len` byte; a single-notify
  read completes its `<+X>` / `[X]` / `<-X>` round trip and times out cleanly rather than hanging
  when the firmware never emits the char; the keepalive fires; and the parser survives chunk-split
  reassembly, leading junk, and a 4096-byte overflow followed by a resync.

- [ ] API spec updated
- [ ] Plugin docs updated
- [ ] Skin docs updated
- [ ] Profile docs updated
- [x] Device docs updated: `doc/DeviceManagement.md` - a new "USB/serial transport behaviour (DE1
      family)" block covering reads, length-exact frames, throughput and link arbitration; a
      "Bluetooth-off operation" paragraph; and the Android name-gate and VID:PID probe-candidate
      wording in the serial detection list.
- [ ] N/A

## Security Impact (required)

- BLE/USB surface changed? `Yes`, and this is the section to read. The Android USB probe now also
  admits ports whose VID:PID is `0x2E8A:0x000A`, and ports whose product name is null (Android often
  reports null before USB permission is granted). That pair is the pico-sdk default, so it matches
  any default Raspberry Pi Pico CDC device, not only a Bengle. Two things bound the risk. First, the
  pair only qualifies a port for the **identification probe**; `bengleUsbIds`, the direct
  instantiation table, is deliberately **empty** and now test-pinned empty, so a random hobby board
  can never be instantiated as an espresso machine. Second, the probe is the existing 3-second
  DE1-protocol handshake, and a device that does not answer it is dropped. The name predicates stay
  exact and case-sensitive on purpose; the descriptor strings are fixed, and loosening them widens
  the probe's reach onto unrelated hardware. `0x2E8A:0x000C` (the Pi debug probe) is explicitly not
  matched. Auto-permission for the VID:PID was already upstream in `device_filter.xml`; I verified
  that and did not re-add it.

## User-Visible Changes

A machine connected over USB is reachable with Bluetooth turned off, and the adapter-error screen now
tells the user that USB keeps working. A Bengle is detected over USB on Android. The preferred
machine continues to be stored per **transport id** (a serial id is the `usb-<vid>-<pid>-<serial>`
stable id, not a BLE MAC), deliberately un-aliased across transports: the first wired session ends at
the picker, and picking the USB machine once makes later launches auto-connect over the wire.

- [x] `flutter test` - **2080 tests pass** at this branch head (B-1 was 2041; `+39` from this
      branch's own tests)

- Real hardware? (DE1/Bengle/scale): `Yes`, for the facts the fixes are built on, though not as a
  re-run of this exact branch. The TinyUSB default descriptors (`0x2E8A:0x000A`, "TinyUSB Device")
  were captured from a Bengle on 2026-07-10. The Bluetooth-off dead end was reproduced on the bench,
  which is how it became clear that fixing one gate is not enough.
- What you personally verified and how: the local gate set above, and the transport behaviour against
  the new `FakeSerialTransport` at the byte level.
- Edge cases checked: a request write that throws while a read is armed (the armed future is
  `.ignore()`d so its later timeout cannot surface as an unhandled async error); a failed read
  (the `<-X>` unsubscribe is sent in a `finally`, so a failure never leaves a subscription eating
  downlink budget); keepalive cancellation on `disconnect()`, `dispose()` and `detach()`.
- What you did **not** verify: I did not re-run this branch on the machine. Specifically unverified
  by me: an actual DFU firmware upload over USB end to end (the `<F>` padding is asserted at the byte
  level in tests, and the DFU uploader's final image chunk is the only short-frame producer in the
  codebase, but I did not push an image through it); and iOS/macOS serial, which this repo does not
  support anyway.

### Evidence

- [x] Test output (2080 pass)
- [ ] Log snippets

## Compatibility & Migration

- Backward compatible? `Yes`

A DE1 over BLE is byte-for-byte unaffected. A DE1 over serial gains the reads that used to throw and
the keepalive, and loses nothing. The keepalive costs roughly one 8-character command up and one
`[N]` frame down every 5 seconds.

1. **The keepalive exists because the firmware link is shared.** BLE and USB share one serial view in
   the firmware, arbitrated by a last-writer-wins `Source` flag, so any stray byte from the BLE module
   silently steals the notify stream from a passively listening USB client. A periodic `<+N>`
   re-asserts `Source = USB`, and because the firmware treats an add-notify as a force-update, it
   doubles as a resync for a framing that has no checksum and no retransmit. 5 seconds bounds the
   worst-case stolen-stream window. This is a workaround for a firmware arbitration design, and if
   you would rather the firmware fixed it, say so; the app-side cost is small but it is not nothing.
2. **Reads that have no serial path now throw `UnsupportedError`, not `UnimplementedError`.** The DE1
   serial view genuinely never emits those characteristics; `setTime`, `writeToMMR`, `headerWrite`
   and `frameWrite` are write-only commands. A descriptive `UnsupportedError` means the raw WebSocket
   API surfaces a clean error to the client instead of crashing the read. That is a behaviour change
   for anyone who was catching `UnimplementedError`, which I doubt is anyone, but it is a change.
3. **`serialKeepaliveInterval` and `serialSingleReadTimeout` are constructor test seams.** They exist
   because `fakeAsync` stalls on the root-zone `_nullFuture` that broadcast-subscription cancels
   return, so the timer tests have to run on real, shortened time. Production never sets them. If you
   dislike constructor seams I can hide them behind `@visibleForTesting` statics instead.
4. **`shouldProbeUsbDevice` is a `@visibleForTesting` static.** It was extracted precisely so the OR
   combination is unit-tested rather than only the two predicates. The predicates passing
   individually is what the old code already "had".

## Risks & Mitigations

- **Risk:** the probe now reaches out to any default pico-sdk CDC device plugged into an Android
  tablet.
  - **Mitigation:** the probe is the DE1-protocol handshake and it is bounded at 3 seconds; a device
    that does not answer is dropped, and `bengleUsbIds` stays empty so nothing can be instantiated
    from the ID alone. The worst case is a 3-second delay on scan for a user with a Pico plugged in.
- **Risk:** demoting the adapter-error view could hide a real Bluetooth problem.
  - **Mitigation:** only the `adapterOff` kind is demoted, and only while something that works
    without Bluetooth is in flight (an active connect, a pending picker, found machines, or machines
    streaming in over `DeviceController.deviceStream`). I included `_discoveredMachines`, not just
    `foundMachines`, on purpose: the latter fills later, and using only it re-opens a window where
    the error flashes over live discovery.
- **Risk:** zero-padding a `<F>` frame changes what goes on the wire for a DE1.
  - **Mitigation:** it is serial-only, it applies only to `writeToMMR`, and the `Len` byte carries
    the true payload length so the padding is inert to the firmware. Every other endpoint already
    emits struct-sized frames by construction.

### DIFF
--- a/assets/api/bengle_hw_v1.yml
+++ b/assets/api/bengle_hw_v1.yml
@@ -1,0 +1,757 @@
+# =============================================================================
+# Bengle firmware <-> app hardware contract (bengle_hw_v1.yml)
+# =============================================================================
+# contract_version: 1
+#
+# Generated from firmware the firmware register table at:
+#
+#
+# Change protocol: firmware changes a register -> regenerate this file from the
+# new the firmware register table -> bump contract_version -> update the app enums (BengleMmr /
+# BengleSteamMmr / BengleScaleMmr / BengleCalMmr / BengleLedMmr / MMRItem) ->
+# the mmr_contract_test checker enforces agreement; firmware and app PRs both
+# cite the contract_version.
+#
+# Normalization rules applied (binding):
+#   * mult / min / max are only meaningful on firmware register-table rows with auto > 0 (the
+#     macro read/write path). auto = 0 rows are served by hand-written firmware
+#     cases; for those, `mult` below records the VERIFIED behaviour of the
+#     hand-written path (which matches the the firmware register table column everywhere except
+#     v13Model, whose mult=1000 column is inert -- recorded as mult: 1).
+#   * min / max below are RAW WIRE UNITS (the little-endian int32 actually on
+#     the wire). Where the firmware register table used post-divide engineering units, the original
+#     engineering values are kept in a trailing comment.
+#   * min/max encoding: where the firmware register table declares the full-u32 column
+#     0..0xFFFFFFFF it is retained verbatim (for a u32 wire value that is
+#     identical to "no constraint"); ScaleTare alone is normalized to
+#     null/null because its written value is ignored entirely (pure trigger).
+#     Deliberate, not drift -- a subset checker passes either encoding.
+#   * Read-back resolution on FLOAT-backed rows (EndOfShotWeight, LastTARE,
+#     ScaleCalWeight, TargetMilkTemp, V24CurrentBudget, MatCurrentTemp): the
+#     firmware macro read path truncates BEFORE scaling
+#     (the firmware the firmware read path),
+#     so wire READS return floor(engineering) x mult -- whole engineering
+#     units -- while writes keep full x-mult resolution. The same U32(float)
+#     cast saturates a negative stored value to 0 on Cortex-M, so these rows
+#     can never read back negative. Affected rows carry a "Read-back" note.
+#     (MatSetPoint is float-backed too but exact: mult 1, whole-degC writes.)
+#   * The firmware does NOT enforce min/max on Bengle macro-path writes
+#     the firmware write path never range-checks): bounds here are documentation and the app
+#     is the sole guard. Checker semantics: app range must be a SUBSET of the
+#     contract range (not equal) -- e.g. calFlowEst app min raw 130 vs FW 125.
+#   * perms are canonicalized from the the firmware register table ENTRY column:
+#     PERM_READ -> R, PERM_RT -> RT, PERM_RW -> RW, PERM_RWD -> RWD (persisted
+#     to disk), PERM_RWT -> RWT (write triggers an action). The ENTRY column is
+#     AUTHORITATIVE: the firmware register table's own header comment block disagrees with it on
+#     row 33 (comment "RW" vs ENTRY PERM_READ) and rows 34/35/37 (comment "W"
+#     vs ENTRY PERM_RWT); the firmware write macro verifiably enforces the
+#     ENTRY perms. App enums carry no perms today, so the checker must NOT
+#     assert perms in v1.
+#   * `app_alias` is present only where the app's declared description string
+#     differs from the firmware register name (TargetMilkTemp is the only
+#     case). Rows are keyed by FIRMWARE name, so the alias records the
+#     app-side name.
+#   * Rows 12 (PrefGHCMCI) and 13 (MaxShotPres) are omitted: PERM_NONE TODO
+#     stubs in firmware, unreachable and irrelevant to the app.
+#   * All len=4 registers are little-endian int32 on the MMR wire. Addresses in
+#     frames are 24-bit big-endian (the 0x00 top byte of 0x008038xx-style
+#     addresses is discarded).
+# =============================================================================
+
+contract_version: 1
+firmware:
+  commit: 0381e7ab58eb5b5ee36c14b0bef123ea3cfe4f2e
+  build: 90
+  source: the firmware register table
+generated: 2026-07-11
+
+# -----------------------------------------------------------------------------
+# MMR register table
+# name = firmware register name; fw_row = the firmware register table RangeNum; address as it
+# appears in a read/write frame (before the 24-bit truncation); length in
+# bytes; mult = raw wire value / engineering value.
+# -----------------------------------------------------------------------------
+registers:
+  # --- Shared DE1-family rows the app's MMRItem enum declares -----------------
+  - name: ExternalFlash
+    fw_row: 0
+    address: "0x00000000"
+    length: 1048575 # 0xFFFFF -- first 1 MB mapped to external flash
+    perms: RW
+    mult: 1
+    value_kind: bytes
+    min: null
+    max: null
+    unit: null
+    semantics: "External-flash window (disk operations, firmware images); not a scalar register."
+
+  - name: HWConfig
+    fw_row: 1
+    address: "0x00800000"
+    length: 4
+    perms: R
+    mult: 1
+    value_kind: int32
+    min: null
+    max: null
+    unit: bitmask
+    semantics: "Hardware configuration word."
+
+  - name: Model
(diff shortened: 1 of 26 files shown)


tadelv:
useful bits merged in #482

PR #461, opened 2026-07-16
feat(bengle): 0xA013 as the sole shot source, and report the firmware's own gravimetric flow

ChampionDesigns:
> **Stacked PR — B-3 of 10.** Builds on #460 (`feat/bengle-serial-usb`). Until that merges this PR's diff includes its commits; please review/merge in order **B-1 → B-10**.

## The story

A Bengle has a scale built into it. Not a scale it talks to, a scale it *is*: two load cells under
the drip tray, read by the machine's own firmware, at 15 Hz, in the same loop that decides when to
stop the shot. The firmware ships that reading, and the gravimetric flow it derives from it, in a
28-byte `0xA013` frame alongside the stock DE1 shot sample. Until this PR the app consumed the weight
from that frame but not the flow, because `ScaleSnapshot` had nowhere to put a flow. So the app did
what it does for a BLE scale: it ran a flow estimator over the weight series and derived its own flow
number. That is the right thing to do for a scale that reports weight only, and it is exactly what
the estimator is for. It is the wrong thing to do here, because the app was re-deriving a quantity
the firmware had already computed, from the very signal the firmware computed it from, and the app's
answer is about a second late. The shot path consumes the app's answer, not the firmware's.

## Summary

- **Problem:** On a Bengle, `MachineSnapshot.weightFlow` carries the firmware's own gravimetric flow
  verbatim, but the **scale** surface carried an app-estimated flow derived from the firmware's own
  weight. Two competing numbers for one quantity, and the shot sequencer reads the estimated one.
- **Why it matters:** The estimate is a second behind at exactly the moment it matters most. Measured
  against a 15 Hz pour whose weight climbs at exactly 2.00 g/s, with the firmware reporting
  `GFlow = 2.00` from the first frame:

  | sample (at 15 Hz) | firmware GFlow | app estimate (Kalman) | app estimate (legacy) |
  |---|---|---|---|
  | 1 (about 67 ms) | 2.0000 | 0.0082 | 0.9950 |
  | 5 (about 333 ms) | 2.0000 | 0.7913 | 1.6658 |
  | 15 (1.0 s) | 2.0000 | 1.9382 | 2.0000 |
  | 59 (3.9 s) | 2.0000 | 2.0007 | 2.0000 |

  `ShotSequencer` projects step-weight exits on that number (`projectedWeight = weight + weightFlow *
  multiplier`, and that path is **not** gated on `_machineHasAutonomousSAW`), `_refineStoppingYield`
  uses it for cup-removal and settle detection on the final yield, `hot_water_stop` projects on it,
  and it is what `ws/v1/scale/snapshot` and the shot record report. So the two snapshot surfaces could
  report flows differing by 2 g/s at the instant a shot starts.
- **What changed:** Three things, in one story. `0xA013` becomes the sole shot source on a Bengle
  (the frame is a reorganised superset of `0xA00D`, and consuming both would double-sample every
  chart). Its weight is bridged into the integrated scale. And `ScaleSnapshot` gains an optional
  `flow`, populated from `GFlow` in that bridge, which `ScaleController` passes through verbatim,
  bypassing both estimators for a device that computes its own flow. Sourcing both surfaces from the
  same frame is what stops them from disagreeing.
- **What did NOT change (scope boundary):** `ScaleSnapshot.flow` defaults to `null` and **every**
  other scale in the repo leaves it `null`, so every BLE scale keeps the estimator it has today, on
  the code path it has today. The estimator itself is untouched: no logic changed inside
  `KalmanFlowEstimator`, no default flipped, no feature flag added or removed.

## About the estimator, plainly

This is not an argument that the Kalman flow estimator is wrong. It is an argument that a Bengle is a
device the estimator was never meant to cover.

A BLE scale reports weight and nothing else. There is no flow to read, so the app has to estimate
one, and estimating it well is a real problem that the estimator solves. A Bengle answers the
question in hardware, on the load cells it owns, in the loop it uses to stop its own shots. Running
any estimator over that weight is not a second opinion, it is a strictly worse copy of the first one,
because it is derived from the same signal with less information and about a second of lag.

Two things are worth being precise about, because they cut in favour of the estimator, not against
it:

1. **This double-derivation pre-dates the Kalman change.** Before it, `ScaleController` derived the
   same redundant flow with `FlowCalculator` plus `MovingAverage`. The Kalman work did not introduce
   the redundancy; it swapped the estimator underneath it (and made it default-on). The legacy
   estimator was also wrong on a Bengle, just differently wrong: it read 0.995 g/s on the first
   sample where the firmware read 2.00.
2. **So the fix is not "disable Kalman".** It is "a device that computes its own flow reports it, and
   the app does not re-derive it". That fixes the Kalman case and the legacy case with the same code,
   and it stays correct no matter which estimator upstream makes the default later. The regression
   test asserts the pass-through with the Kalman flag **ON** as well as OFF, and asserts that toggling
   the flag does not change what a Bengle reports. That is the lock: the estimator choice must stay
   inert on a device that answers the question in hardware.

## Change Type (select all)

- [x] Bug fix
- [x] Feature
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [x] REST API / handlers (schema only)
- [x] WebSocket API (schema and description only)
- [x] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

- Related # `N/A`

## Root Cause (if bug fix)

- **Root cause:** `ScaleSnapshot` can only carry weight. The `Scale` contract has no way to express
  "this device already knows its flow", so the bridge that turns a `0xA013` frame into a
  `ScaleSnapshot` had to drop `GFlow` on the floor (its own comment said so), and `ScaleController`
  had no way to know it should not estimate. The information was present in the frame, present on the
  machine snapshot, and unrepresentable on the surface that the shot sequencer actually reads.
- **Missing detection or guardrail:** nothing asserted that the two `weightFlow` surfaces agree, and
  the existing `shot_sequencer_bengle_bypass_test` locks the autonomous **stop**, not the flow
  **source**, so a Bengle looked covered while the number feeding the step-weight projection was
  wrong.
- **Contributing context:** the final stop-at-weight is safe on a Bengle regardless, because
  `_machineHasAutonomousSAW` hands the stop to the firmware. That safety made the estimated flow easy
  to overlook, since the visible failure mode (a badly missed target weight) does not occur.

  - [x] Unit test
- **Target test or file:** `test/unit/controllers/bengle_device_flow_test.dart` (new). Supporting:
  `bengle_shot_sample_test.dart`, `bengle_shotsample_pipeline_test.dart`,
  `bengle_shotsample_serial_test.dart`, `machine_snapshot_test.dart`.
- **Scenario the test should lock in:** with a real `ScaleController` driving a `BengleVirtualScale`
  over a fake transport, and `setKalmanFlowEnabled(true)` (the upstream default), the emitted
  `WeightSnapshot.weightFlow` must equal the firmware's `GFlow` **verbatim on the first sample** and
  on every sample of a steady pour, and toggling the Kalman flag must not change what is emitted. The
  decoder tests pin the frame byte-exactly (big-endian, weight at offset 20 is U16P5 scaled by 1/32,
  **not** 1/100 as on `0xA00D`; a diverging case is asserted so the two scalings cannot be confused),
  and the pipeline tests assert that a plain DE1 never subscribes the `0xA013` characteristic.

- [x] API spec updated: `assets/api/rest_v1.yml` and `assets/api/websocket_v1.yml` - the
      `MachineSnapshot` schema gains `weight`, `weightFlow` and `milkTemperature`, and the
      `WeightSnapshot.weightFlow` description now states the two cases (app-derived for a
      weight-only scale, passed through verbatim for a scale that computes its own).
- [x] API docs updated: `doc/Api.md` - the `/ws/v1/machine/snapshot` row.
- [ ] Plugin docs updated
- [ ] Skin docs updated
- [ ] Profile docs updated
- [x] Device docs updated: `doc/DeviceManagement.md` - the `0xA013` stream in the
      continuously-subscribed set, and the serial `<-M>` unsubscribe policy.

## Security Impact (required)

- New or changed REST endpoints? `No` - schema fields only, on an existing response.
- New or changed WebSocket topics? `No` - same topics, additive fields on `MachineSnapshot`.
- BLE/USB surface changed? `Yes` - a Bengle now subscribes the `0xA013` characteristic and, on
  serial, unsubscribes `0xA00D`. The BLE subscribe is gated on the **confirmed** `v13Model` identity
  and never fires on a plain DE1: blind-enabling a characteristic a DE1 does not have throws and
  permanently stalls the BLE command queue (this is a known de1plus failure mode), so the gate is
  load-bearing and there is a negative test for it.

## User-Visible Changes

On a Bengle: the shot chart, the shot record, `ws/v1/scale/snapshot` and `ws/v1/machine/snapshot` all
report the same flow number, the firmware's, and it is correct from the first frame of the shot
rather than converging over about a second. Step-weight exits and the final-yield refinement project
on that number. `MachineSnapshot` gains `weight`, `weightFlow` and `milkTemperature` for every
machine, which are `0.0` on a machine without an integrated scale. On any other scale: nothing
changes at all.

- [x] `flutter test` - **2104 tests pass** at this branch head (B-2 was 2080; `+24` from this
      branch's own tests)

- Real hardware? (DE1/Bengle/scale): `Yes`, for the underlying pipeline, on the working branch this
  code was sliced from: a real shot was pulled over BLE through the `0xA013` path, single-sourced,
  and the yield tracked (51.7 g). The serial dual-stream overrun described below was confirmed on
  hardware on 2026-07-09.
- What you personally verified and how: the local gate set, and the numbers in the table above, which
  come from a focused probe test driving a 15 Hz
  ramping weight through the real `ScaleController` with the real estimators. I read the shot
  sequencer, hot-water stop and scale handler call sites to establish who consumes the estimated
  number.
- Edge cases checked: post-tare flow suppression (the specced no-spike-after-tare window is still
  honoured on the device-flow path, so a device flow is zeroed during the settle window exactly like
  an estimated one); frames shorter than 28 bytes dropped at both the transport and decoder layers;
  a plain DE1 never subscribing `0xA013`.
- What you did **not** verify: I did not re-run this resliced branch on the machine, and I have not
  measured the firmware's `GFlow` against an external reference scale. My claim is only that the
  firmware's number is available immediately and the app's is not, and that the two surfaces should
  agree; I am **not** claiming the firmware's absolute accuracy has been independently validated. I
  also have not exercised a milk probe reporting a non-zero `milkTemperature` on hardware; that field
  is decoded and tested from a golden frame only.

### Evidence

- [x] Test output (2104 pass; `bengle_device_flow_test` is the lock)
- [x] Log snippets (the measured table above, from the probe test)

## Compatibility & Migration

- Backward compatible? `Yes`

`ScaleSnapshot.flow` is additive and nullable and defaults to `null`; no existing scale constructs it.
`MachineSnapshot`'s three new fields are additive with `0.0` defaults, and `fromJson` tolerates absent
keys, so payloads recorded before this change still decode. On a Bengle, `ws/v1/scale/snapshot`
reports a different (better) `weightFlow` than it did before, which is a value change, not a shape
change.

1. **`ScaleSnapshot.flow` is a shared contract change.** This is the hunk to argue with. It is
   additive and nullable, and I chose it over the alternatives deliberately: a Bengle-specific
   subclass of `ScaleSnapshot` would have `ScaleController` type-testing on a device class in the hot
   path, and a "disable the estimator for Bengles" flag in `ScaleController` would put device
   knowledge in the controller and would have to be re-fought every time the estimator changes.
   Expressing the actual fact ("this device knows its own flow") on the contract keeps
   `ScaleController` device-agnostic: it asks `snapshot.flow != null`, exactly the way it already
   asks `previous is TransportHandoffScale`. If you would rather this were expressed as a capability
   interface on `Scale` instead of a field on the snapshot, that is a reasonable call and I will
   rework it.
2. **`ScaleSnapshot.toJson` deliberately does not serialise `flow`.** The scale WebSocket carries
   `WeightSnapshot`, not `ScaleSnapshot`, and `WeightSnapshot.weightFlow` already exists and now
   carries the passed-through value. Adding `flow` to `ScaleSnapshot.toJson` would put a second,
   redundant flow field on a payload that already has one. Flagging it because it looks like an
   oversight and is not.
3. **The post-tare settle window still applies to the device-flow path.** The firmware's tare is
   instant and its `GFlow` does not spike, so arguably this is dead code. I kept it because the
   no-spike-after-tare guarantee is specced behaviour and honouring it costs one comparison per
   sample. If you would rather the device path be a pure pass-through, delete the four lines.
4. **`0xA013` is the sole shot source, and on serial `0xA00D` is dropped.** On BLE, `0xA00D` stays
   subscribed and is parsed and dropped (there is bandwidth headroom, and it keeps the raw `[M]`
   WebSocket visibility). On serial it is unsubscribed once the Bengle identity is confirmed, because
   the firmware serial downlink tops out at roughly 1920 bytes per second (half duplex) and two 15 Hz streams overrun it. That was confirmed on hardware on 2026-07-09 as
   truncated and odd-length frames with visible weight flicker. The asymmetry is deliberate and
   documented, but it is asymmetry, so it is worth an explicit nod.
5. **The pipeline is chosen at access time, not in a field initialiser.** `currentSnapshot` picks
   between the Bengle and DE1 snapshot pipelines when it is read. Picking in a field initialiser
   would latch the wrong pipeline for any listener that attaches before `onConnect` completes.

## Risks & Mitigations

- **Risk:** a future scale reports a `flow` that is worse than the app's estimate, and silently opts
  itself out of the estimator by populating the field.
  - **Mitigation:** the field means "this device computes flow on its own hardware", and the doc
    comment says so. It is opt-in per device; nothing populates it by accident. The judgement call
    stays with whoever writes the device class.
- **Risk:** somebody later "simplifies" `ScaleController` by removing the device-flow branch, and a
  Bengle silently goes back to an estimated flow.
  - **Mitigation:** `bengle_device_flow_test` fails immediately, and it fails with the Kalman flag on
    (the default), so it cannot be dismissed as a legacy-path test.
- **Risk:** `MachineSnapshot` grew three fields that are `0.0` on every DE1, which is dead weight on
  the wire for most users.
  - **Mitigation:** three doubles per snapshot frame. I judged that cheaper than a parallel
    Bengle-only snapshot topic that every client would then have to learn about. Say the word if you
    disagree.

### DIFF
--- a/assets/api/bengle_hw_v1.yml
+++ b/assets/api/bengle_hw_v1.yml
@@ -714,7 +714,7 @@ packet_0xA013:
     - { offset: 4,  size: 2, name: SetGroupPressure, type: u16, scale: 0.01, unit: bar,  semantics: "Set pressure; 0 if no target." }
     - { offset: 6,  size: 2, name: GroupFlow,        type: u16, scale: 0.01, unit: ml/s, semantics: "Estimated flow at group." }
     - { offset: 8,  size: 2, name: SetGroupFlow,     type: u16, scale: 0.01, unit: ml/s, semantics: "Set flow; 0 if no target." }
-    - { offset: 10, size: 2, name: GFlow,            type: u16, scale: 0.01, unit: g/s,  semantics: "Gravimetric flow from the integrated scale -> MachineSnapshot.weightFlow." }
+    - { offset: 10, size: 2, name: GFlow,            type: u16, scale: 0.01, unit: g/s,  semantics: "Gravimetric flow, computed on-device from the load cell -> MachineSnapshot.weightFlow AND the scale surface's weightFlow (passed through verbatim; the app does not re-estimate flow for a device that computes its own)." }
     - { offset: 12, size: 2, name: MixTemp,          type: u16, scale: 0.01, unit: degC, semantics: "Water temperature entering the group." }
     - { offset: 14, size: 2, name: HeadTemp,         type: u16, scale: 0.01, unit: degC, semantics: "Showerhead water temperature (u16/100 here; 0xA00D used 3-byte U24P16 -- layout differs)." }
     - { offset: 16, size: 2, name: SetMixTemp,       type: u16, scale: 0.01, unit: degC, semantics: "Mix temperature target; 0 if no target." }
--- a/assets/api/bengle_hw_v1.yml
+++ b/assets/api/bengle_hw_v1.yml
@@ -1,0 +1,757 @@
+# =============================================================================
+# Bengle firmware <-> app hardware contract (bengle_hw_v1.yml)
+# =============================================================================
+# contract_version: 1
+#
+# Generated from firmware the firmware register table at:
+#
+#
+# Change protocol: firmware changes a register -> regenerate this file from the
+# new the firmware register table -> bump contract_version -> update the app enums (BengleMmr /
+# BengleSteamMmr / BengleScaleMmr / BengleCalMmr / BengleLedMmr / MMRItem) ->
+# the mmr_contract_test checker enforces agreement; firmware and app PRs both
+# cite the contract_version.
+#
+# Normalization rules applied (binding):
+#   * mult / min / max are only meaningful on firmware register-table rows with auto > 0 (the
+#     macro read/write path). auto = 0 rows are served by hand-written firmware
+#     cases; for those, `mult` below records the VERIFIED behaviour of the
+#     hand-written path (which matches the the firmware register table column everywhere except
+#     v13Model, whose mult=1000 column is inert -- recorded as mult: 1).
+#   * min / max below are RAW WIRE UNITS (the little-endian int32 actually on
+#     the wire). Where the firmware register table used post-divide engineering units, the original
+#     engineering values are kept in a trailing comment.
+#   * min/max encoding: where the firmware register table declares the full-u32 column
+#     0..0xFFFFFFFF it is retained verbatim (for a u32 wire value that is
+#     identical to "no constraint"); ScaleTare alone is normalized to
+#     null/null because its written value is ignored entirely (pure trigger).
+#     Deliberate, not drift -- a subset checker passes either encoding.
+#   * Read-back resolution on FLOAT-backed rows (EndOfShotWeight, LastTARE,
+#     ScaleCalWeight, TargetMilkTemp, V24CurrentBudget, MatCurrentTemp): the
+#     firmware macro read path truncates BEFORE scaling
+#     (the firmware the firmware read path),
+#     so wire READS return floor(engineering) x mult -- whole engineering
+#     units -- while writes keep full x-mult resolution. The same U32(float)
+#     cast saturates a negative stored value to 0 on Cortex-M, so these rows
+#     can never read back negative. Affected rows carry a "Read-back" note.
+#     (MatSetPoint is float-backed too but exact: mult 1, whole-degC writes.)
+#   * The firmware does NOT enforce min/max on Bengle macro-path writes
+#     the firmware write path never range-checks): bounds here are documentation and the app
+#     is the sole guard. Checker semantics: app range must be a SUBSET of the
+#     contract range (not equal) -- e.g. calFlowEst app min raw 130 vs FW 125.
+#   * perms are canonicalized from the the firmware register table ENTRY column:
+#     PERM_READ -> R, PERM_RT -> RT, PERM_RW -> RW, PERM_RWD -> RWD (persisted
+#     to disk), PERM_RWT -> RWT (write triggers an action). The ENTRY column is
+#     AUTHORITATIVE: the firmware register table's own header comment block disagrees with it on
+#     row 33 (comment "RW" vs ENTRY PERM_READ) and rows 34/35/37 (comment "W"
+#     vs ENTRY PERM_RWT); the firmware write macro verifiably enforces the
+#     ENTRY perms. App enums carry no perms today, so the checker must NOT
+#     assert perms in v1.
+#   * `app_alias` is present only where the app's declared description string
+#     differs from the firmware register name (TargetMilkTemp is the only
+#     case). Rows are keyed by FIRMWARE name, so the alias records the
+#     app-side name.
+#   * Rows 12 (PrefGHCMCI) and 13 (MaxShotPres) are omitted: PERM_NONE TODO
+#     stubs in firmware, unreachable and irrelevant to the app.
+#   * All len=4 registers are little-endian int32 on the MMR wire. Addresses in
+#     frames are 24-bit big-endian (the 0x00 top byte of 0x008038xx-style
+#     addresses is discarded).
+# =============================================================================
+
+contract_version: 1
+firmware:
+  commit: 0381e7ab58eb5b5ee36c14b0bef123ea3cfe4f2e
+  build: 90
+  source: the firmware register table
+generated: 2026-07-11
+
+# -----------------------------------------------------------------------------
+# MMR register table
+# name = firmware register name; fw_row = the firmware register table RangeNum; address as it
+# appears in a read/write frame (before the 24-bit truncation); length in
+# bytes; mult = raw wire value / engineering value.
+# -----------------------------------------------------------------------------
+registers:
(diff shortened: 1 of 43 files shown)


tadelv:
reaped in #601

PR #464, opened 2026-07-16
feat(bengle): LED strip palettes + live preview over MMR

ChampionDesigns:
> **Stacked PR — B-6 of 10.** Builds on #463 (`feat/bengle-scale-calibration`). Until that merges this PR's diff includes its commits; please review/merge in order **B-1 → B-10**.

A Bengle has front and rear light strips, and the app has had a full REST surface for them since the capability landed: read the configuration, write all three zones in both awake and sleeping modes, commit, reset. None of it did anything. The capability was a scaffold written before the firmware wire spec was published, so every endpoint address was `null`, every write was swallowed, and the code logged one line per session and returned success. A user who set the strips to a warm white got a `200 OK` and a machine that stayed exactly the colour it already was, and a `GET` afterwards cheerfully echoed back the colour they had asked for, because the value never left the app's own cache. This PR wires the capability to the six real firmware registers, hydrates the cache from the machine on connect so the app stops inventing state, and adds a live-preview endpoint so a colour picker can show a colour on the strip without storing it.

## Summary

- **Problem:** `LedStripCapability` was a stub. `BengleLedEndpoint` returned `null` for every `uuid` and `representation`; `setLedStrip`, `commitLedStrip` and `resetLedStrip` all no-opped with a once-per-session info log. The REST endpoints returned success and wrote nothing to the machine. The in-memory cache was seeded all-off and never corrected, so `GET /api/v1/machine/ledStrip` reported black on a machine that was glowing.
- **Why it matters:** The LED strip is one of the two things a Bengle owner can see the app is supposed to control (the other being the cup warmer, B-7). It failed silently in the worst way available: `200 OK`, a cache that agreed with you, and a machine that ignored you.
- **What changed:** Six firmware MMR registers replace the null endpoints - four persisted palette slots (`FrontLEDAwake` `0x00803898`, `FrontLEDSleep` `0x008038A0`, `RearLEDAwake` `0x0080389C`, `RearLEDSleep` `0x008038A4`) and two live-colour slots (`FrontLEDColor` `0x00803890`, `RearLEDColor` `0x00803894`). `initLedStrip` now hydrates the cache from the four palette registers on connect. Two new endpoints, `POST /api/v1/machine/ledStrip/preview` and `POST /api/v1/machine/ledStrip/preview/clear`, drive the live slots so a UI can preview the sleep colour while the machine is awake.
- **What did NOT change (scope boundary):** The `LedStripState` JSON shape is unchanged - still three zones by two modes. `GET`, `PUT`, `/commit` and `/reset` keep their paths and their status codes. No skin ships in this PR; the Streamline LED UI is a separate repo and a separate review.

## Change Type (select all)

- [x] Bug fix
- [x] Feature
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [x] REST API / handlers
- [ ] WebSocket API
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

- Closes #

## Root Cause (if bug fix)

- **Root cause:** The capability was written against a firmware wire spec that did not exist yet. `BengleLedEndpoint.uuid` and `.representation` were hard-coded `null` with a `// TBD with FW` comment, and the mixin treated a null wire as "capability not yet wired" and no-opped by design. That was the right call at the time. What made it a bug rather than a placeholder is that the REST layer never learned about it: the handler returned `202`/`200` exactly as it would for a real write, and the cache absorbed the value so a read-back confirmed the lie.
- **Missing detection or guardrail:** Nothing distinguished "the capability succeeded" from "the capability decided to do nothing". A stub that returns success is indistinguishable from a working feature at every layer above it, and the tests pinned the stub (they asserted the no-op) rather than the behaviour.
- **Contributing context:** The registers were in the firmware the whole time. They are in the firmware register table, and the app-side comment in the previous code said the wires were unpublished. They were simply never reconciled.

  - [ ] Unit test
- **Target test or file:** `test/unit/models/device/impl/de1/unified_de1/led_strip_capability_test.dart` (+392 lines), plus `test/unit/models/device/impl/bengle/mmr_contract_test.dart` (the four palette registers and the two live registers are now registered with the firmware contract checker).
- **Scenario the test should lock in:** Drive a real `Bengle` over `FakeBleTransport` and assert the **bytes on the wire**, not the cache: a `setLedStrip` writes four MMR frames to the four palette addresses in `0x00RRGGBB` form; a `previewLedColor` writes the two live addresses and does **not** touch the palette or the cache; `clearLedPreview` writes the cached awake palette back to the live addresses; `initLedStrip` reads the four palette registers into the cache; and a read that throws leaves the cache seeded all-off rather than failing the connect. The contract test additionally fails CI if any of these six addresses, lengths or scales drift from `assets/api/bengle_hw_v1.yml`.
- **If no new test added, why not:** N/A.

- [x] API spec updated: `assets/api/rest_v1.yml` - `GET` gains the connect-time-hydration contract; `PUT` describes the persist-on-write semantics and the ignored `frontSwitch` zone; `/commit` and `/reset` are re-described (see below); the two `/preview` endpoints are new.
- [ ] N/A - no docs affected

The e2e scenario `docs/e2e/decent-app/scenarios/bengle-led-strip.md` is refreshed in the same commit.

## Security Impact (required)

- New or changed REST endpoints? `Yes` - `POST /api/v1/machine/ledStrip/preview` and `POST /api/v1/machine/ledStrip/preview/clear`. Both are Bengle-only (404 on any other machine) and both sit on the same unauthenticated local web server as the existing LED endpoints, which already accept arbitrary colours. The new surface is a colour, on a light. It grants no reach a caller did not already have through `PUT /api/v1/machine/ledStrip`.
- BLE/USB surface changed? `Yes` - six new MMR reads/writes, all Bengle-only. They are registered with the contract checker, which fails CI if an address drifts from the firmware contract file.

## User-Visible Changes

The LED strip endpoints now actually change the colour of the lights. `GET /api/v1/machine/ledStrip` returns the colours stored **on the machine** after a connect, rather than all-off, so a fresh app install shows the user their real palette. Two new endpoints allow a live preview. `POST /commit` and `POST /reset` keep working and keep their status codes, but their meaning has changed - see the deliberate choices below. Their request bodies are now `required: false`, which is a relaxation, not a break.

- [x] `flutter test` - **2170 tests pass** on this branch at `89d421f1` (B-5 was 2150, so this branch adds 20).

- Simulated devices? (`simulate=1`): `Yes` - against `MockBengle`, `GET /api/v1/machine/ledStrip` returns three zones by two modes in 16-bit RGB; `POST /ledStrip/preview` returns `202`; `POST /ledStrip/preview/clear` returns `202`.
- What you personally verified and how: the wire bytes, through the tests. `FakeBleTransport` captures the MMR frames and the assertions are byte-exact against the six addresses.
- Edge cases checked: a failed hydration read on connect (the cache stays all-off and the connect still succeeds); a preview followed by a clear (the palette registers are untouched throughout); a 16-bit colour round-trip through the firmware's 8-bit-per-channel encoding.
- What you did **not** verify: **I have not lit a real LED.** The six register addresses come from the firmware register table and are pinned by the contract checker against `assets/api/bengle_hw_v1.yml`, but no one has yet confirmed on hardware that writing `FrontLEDAwake` turns the front strip that colour. `MockBengle.previewLedColor` is deliberately a no-op, so the simulated run proves the REST plumbing and nothing about the machine. **This should be smoke-tested on a real Bengle before merge.**

### Evidence

- [x] Test output (2170 pass; the new capability tests fail against the stubbed code, which is the point)
- [ ] Log snippets
- [x] curl output (the two new endpoints, against `simulate=1`)

## Compatibility & Migration

- Backward compatible? `Yes`, with one semantic caveat. Every existing path, method and response shape is unchanged, and the two endpoints whose request body was `required: true` now accept an empty body. But `/commit` and `/reset` no longer mean what their names say (below). A client that calls `PUT` then `/commit` keeps working; the `/commit` is now redundant rather than necessary.

- **`/commit` and `/reset` are now vestigial, and I kept them anyway.** The four palette registers are `PERM_RWD` in firmware - they persist on every write. There is no separate commit register, so the live/persist split the API was designed around does not exist on the wire. Rather than break the API I made `commitLedStrip()` re-assert the cached palette (a harmless idempotent re-write) and `resetLedStrip()` re-read the registers into the cache. That means `/reset` is a cache re-hydrate, **not a rollback** - it returns whatever was last written, so it cannot undo a `PUT`. The spec now says so in as many words. The honest alternative is to deprecate both endpoints; I did not want to make that call unilaterally on someone else's public API.
- **The `frontSwitch` zone has no register and is silently ignored on write.** The physical switch light mirrors the front strip in firmware; there is no independent control. I kept the zone in `LedStripState` for API symmetry and made reads mirror the front strip's values into it. The alternative - reject a request that sets `frontSwitch` to something different - is arguably more honest, but it would break the existing three-zone request shape. This is documented in `rest_v1.yml` but a client cannot detect it programmatically.
- **`Color16` is 16-bit per channel and the firmware is 8.** Writes take the high byte of each channel; reads byte-replicate back (`0xAB` becomes `0xABAB`). Round-trips are lossless for anything that originated as 8-bit colour, which is every colour picker I know of, but a genuinely 16-bit source loses its low byte silently.
- **A missing or malformed `front`/`back` in a preview request previews black.** That is `Color16.fromJson`'s existing semantics, not a decision I made here, but it means `POST /ledStrip/preview` with `{}` turns both strips off. I documented it rather than special-casing it, because the alternative is to diverge from how `Color16` is parsed everywhere else in the API. If you would rather have a `400` there, say so and I will add it.

## Risks & Mitigations

- **Risk:** The connect-time hydration read fails (older firmware, a flaky transport), the cache stays all-off, and the user then clicks "clear preview" - which writes the cached awake palette, i.e. **black** - turning the strips off when they meant to restore them.
  - **Mitigation:** The failure path is logged at `warning` and the spec documents the exact sequence. It is bounded: any `PUT` or `POST /reset` repairs the cache. I chose a tolerant hydration (a failed read never fails the connect) over a strict one, because a Bengle that cannot connect because its LEDs would not answer is a much worse failure than a strip that goes dark.
- **Risk:** The register addresses are wrong or the byte order is wrong, and the strips do something surprising on a real machine.
  - **Mitigation:** The contract checker pins all six against `assets/api/bengle_hw_v1.yml`, so they cannot silently drift. It does not, however, prove they were right to begin with - see "what you did not verify" above. A hardware smoke test is the mitigation, and it has not been done.

### DIFF
--- a/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
+++ b/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
@@ -50,6 +50,8 @@ curl -s -X PUT http://localhost:8080/api/v1/scale/tare
 
 Expected: `200 OK`, empty body.
 
+On real hardware this generic endpoint reaches the machine's `ScaleTare` MMR write-trigger (`0x0080388C`) through `BengleVirtualScale.tare()` → `tareIntegratedScale()`; the firmware performs an immediate tare and subsequent `0xA013` weight arrives already net of the new zero. Confirm a tare by watching the weight drop toward 0 — never the `0xA013` Flags bit.
+
 Re-sample the snapshot and confirm weight is ~0 (MockBengle resets `_tareOffset` to the current accumulated weight, so the next emission reads near zero):
 
 ```bash
--- a/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
+++ b/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
@@ -21,7 +21,7 @@ curl -sf http://localhost:8080/api/v1/machine/capabilities | jq .
 Expected:
 
 ```json
-{ "capabilities": ["cupWarmer", "integratedScale", "ledStrip", "stopAtWeight"] }
+{ "capabilities": ["cupWarmer", "integratedScale", "ledStrip", "stopAtWeight", "scaleCalibration"] }
 ```
 
 Quick assertions:
--- a/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
+++ b/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
@@ -1,6 +1,6 @@
 # Scenario: Bengle integrated scale end-to-end
 
-Verifies that when a Bengle is the connected machine, the integrated scale is auto-attached as a virtual scale (no external scale connection needed), capability discovery advertises `integratedScale`, the `/api/v1/scale/*` REST surface and `/ws/v1/scale/snapshot` stream both flow through the integrated scale, and software stop-at-weight (SAW) ends an espresso shot when the integrated scale weight crosses the profile target.
+Verifies that when a Bengle is the connected machine, the integrated scale is auto-attached as a virtual scale (no external scale connection needed), capability discovery advertises `integratedScale` and `stopAtWeight`, the `/api/v1/scale/*` REST surface and `/ws/v1/scale/snapshot` stream both flow through the integrated scale, and **autonomous** stop-at-weight (SAW) ends an espresso shot: the app reflects the workflow's `targetYield` into the machine's SAW target and defers the final yield stop to the machine (`machineHasAutonomousSAW`), which `MockBengle` simulates firmware-side.
 
 ## Preconditions
 
@@ -12,7 +12,7 @@ No `--connect-scale` flag. On Bengle the integrated scale always wins — extern
 
 ## Steps
 
-### 1. Capability discovery lists both Bengle surfaces
+### 1. Capability discovery lists the Bengle surfaces
 
 ```bash
 curl -sf http://localhost:8080/api/v1/machine/capabilities | jq .
@@ -21,17 +21,17 @@ curl -sf http://localhost:8080/api/v1/machine/capabilities | jq .
 Expected:
 
 ```json
-{ "capabilities": ["cupWarmer", "integratedScale"] }
+{ "capabilities": ["cupWarmer", "integratedScale", "ledStrip", "stopAtWeight"] }
 ```
 
 Quick assertions:
 
 ```bash
 curl -sf http://localhost:8080/api/v1/machine/capabilities \
-  | jq -e '.capabilities | index("integratedScale") != null and index("cupWarmer") != null'
+  | jq -e '.capabilities | index("integratedScale") != null and index("stopAtWeight") != null'
 ```
 
-Exit 0 → both identifiers present.
+Exit 0 → both identifiers present. `stopAtWeight` advertises that the machine firmware runs its own stop-at-weight from the integrated scale — there is no dedicated SAW endpoint; the target rides `PUT /api/v1/workflow` (`context.targetYield`).
 
 ### 2. Scale snapshot stream is alive without an external scale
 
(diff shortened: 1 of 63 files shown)


tadelv:
Superseded by #612 (merged), which implements the real persisted LED palette. Deliberately dropped/redesigned: live-register preview endpoints (persisted snapshots clobbered by state changes, not a config surface), the fake independent frontSwitch control (derived, not a register), and commit/reset as persist/rollback (no commit latch; reset = truthful reload).

PR #465, opened 2026-07-16
feat(bengle): wire the cup warmer and the milk-probe steam stop to their real firmware registers

ChampionDesigns:
> **Stacked PR — B-7 of 10.** Builds on #464 (`feat/bengle-led-strip`). Until that merges this PR's diff includes its commits; please review/merge in order **B-1 → B-10**.

Ask a Bengle for a 45 degree cup warmer and the app wrote 450 to the machine. The register is a whole-degree integer and reaprime was encoding it as deci-degrees, so every setpoint went out ten times too hot and every read came back ten times too cold. It did not actually cook anyone's cups, and the reason is the second bug: setting the temperature does nothing at all unless you also set `CupWarmerMode`, a separate enable register the app had never heard of, so the mat never came on in the first place. The milk-probe steam stop was in a third state again - its MMR address was the literal placeholder `0x00000000`, so the auto-stop target was cached in the app and never written anywhere. This PR wires all three to the registers the firmware actually has, adds a live mat-temperature readout so the UI can show what the warmer is doing, and pins the connect-time feature-flag advertisement that the whole presence system depends on.

## Summary

- **Problem:** Three separate thermal features were mis-wired. (1) `MatSetPoint` (`0x00803874`) is `mult = 1` in the firmware - whole degrees C - but the app declared `writeScale: 10.0` and `readScale: 0.1`, so every write was 10x too hot and every read 10x too cold. (2) Writing a setpoint alone is inert: the warmer's real enable is `CupWarmerMode` (`0x008038AC`), which the app never wrote, so the mat never heated regardless. (3) `BengleSteamMmr.stopAtTemperatureTarget` was a `0x00000000` stub that cached the target locally and logged once; the real register, `TargetMilkTemp` (`0x008038A8`), was never written, so the firmware's autonomous milk stop could never be armed.
- **Why it matters:** The cup warmer is a headline Bengle feature and it did not work at all. The 10x scaling error was live in a shipped register write - it did no harm only because the enable register kept the mat off. Fixing the enable without fixing the scale would have made a latent bug into a hot one, which is why they land together.
- **What changed:** `MatSetPoint` is now unscaled with a `max` of 80 C. `setCupWarmerTemperature` writes the setpoint **and** `CupWarmerMode` (a positive target enables, 0 disables), and `Bengle.onConnect` re-asserts both, because `CupWarmerMode` is deliberately RAM-only in firmware. `TargetMilkTemp` is wired for real, clamped 0 to 85 C, and `SteamSequencer.useFwAutonomousStop` now returns true when a probe is attached with a positive target, so the app stops racing the firmware's own stop. `GET /api/v1/machine/cupWarmer` gains a nullable `currentTemperature` read from `MatCurrentTemp` (`0x008038CC`).
- **What did NOT change (scope boundary):** No new endpoint and no new WebSocket topic. `PUT /api/v1/machine/cupWarmer` keeps its shape and its 0 to 80 range. The scheduled pre-warm is **not** here - it is B-10, and it is blocked. Extra cup-warmer instrumentation that exists in firmware development was deliberately **dropped**: nothing in the app or any skin consumes it.

## Change Type (select all)

- [x] Bug fix
- [x] Feature
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [x] REST API / handlers
- [ ] WebSocket API
- [x] Machine state / shot logic (the steam sequencer defers to the firmware stop)
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

- Closes #

## Root Cause (if bug fix)

- **Root cause:** All three registers were declared from documentation that was either wrong or absent, and never reconciled against the firmware's own the firmware register table. `MatSetPoint` had been mis-encoded twice - first as a raw IEEE-754 float32 (an early firmware note, still in the code comment this PR deletes), then as deci-degrees. the firmware register table row 36 says `mult 1` and the firmware write path consumes the raw value as degrees C; the hardware wins. The enable register and the milk target were simply not known about when the capability was written.
- **Missing detection or guardrail:** Nothing checked the app's register declarations against the firmware's. B-1 introduced exactly that guardrail - `assets/api/bengle_hw_v1.yml` plus a contract checker in CI - and this PR registers all four registers with it, so none of these three encodings can silently regress again.
- **Contributing context:** The 10x error was invisible because the second bug masked it. A mat that never turns on cannot be observed to be 450 degrees.

  - [ ] Unit test
- **Target test or file:** `test/models/device/bengle_cup_warmer_test.dart` (+152 lines), `test/models/device/bengle_steam_stop_test.dart` (rewritten from stub-pinning to wire-byte assertions), `test/unit/models/device/impl/bengle/mmr_contract_test.dart`, and `test/models/device/unified_de1_presence_flags_test.dart` (new).
- **Scenario the test should lock in:** Byte-exact wire assertions, not cache assertions. Requesting 70 C must put a little-endian `70` on the wire at `0x00803874`, not `700`; a raw read of `50` must decode as `50.0`; an over-range request must clamp to 80 on the wire. A positive setpoint must produce a **second** write of `1` to `0x008038AC`, and `0.0` must produce a `0`. A reconnect must re-assert both - and, critically, an app instance that never enabled the warmer must write **nothing** on connect. A plain DE1 connect must never touch either register. For milk: 65 C must encode as `[0x8A, 0x02]` at `0x008038A8`, `0` must disable, and the clamp must hold at 0 and 85. The contract test fails CI if any address, scale or bound drifts from the contract file.

- [x] API spec updated: `assets/api/rest_v1.yml` - `CupWarmerState.currentTemperature` (nullable, read-only); the `cupWarmer` `PUT` and `GET` descriptions now state the enable semantics and the reboot re-assert; `SteamSettings.stopAtTemperature` is now `0..85` and described as a real register rather than "scaffolding only".
- [ ] N/A - no docs affected

The e2e scenario `docs/e2e/decent-app/scenarios/bengle-cup-warmer.md` is refreshed in the same commit (it described a `POST` returning `202`; the endpoint is a `PUT` returning `200`).

## Security Impact (required)

- New or changed REST endpoints? `No` - `GET /api/v1/machine/cupWarmer` gains a field; no new path.
- BLE/USB surface changed? `Yes` - four Bengle-only MMR registers: `MatSetPoint` (re-scaled), `CupWarmerMode` (new write), `MatCurrentTemp` (new read), `TargetMilkTemp` (new write). All four are registered with the contract checker.

Worth stating plainly, because it is a heater: the firmware does **not** clamp Bengle macro-path writes to `MatSetPoint`. The app's declared `max: 80` is the **only** guard on how hot the mat is told to get. That is the reason this PR narrows the declared range from `800` to `80` rather than leaving it wide, and the reason the clamp is asserted on the wire, in a test, at both ends.

## User-Visible Changes

The cup warmer works. Setting a temperature now heats the mat, at the temperature asked for rather than ten times it, and the setting survives a machine reboot (the app re-asserts it on connect). `GET /api/v1/machine/cupWarmer` gains `"currentTemperature"`, a number or `null`, so a UI can show the live mat temperature - clients must tolerate both `null` and the key being absent entirely on an older app. Milk steam stop-at-temperature can now be armed, up to 85 C rather than the 80 the API previously advertised, and when a probe is attached the **firmware** stops the steam rather than the app.

- [x] `flutter test` - **2188 tests pass** on this branch at `76fcd534` (B-6 was 2170, so this branch adds 18).

- Simulated devices? (`simulate=1`): `Yes` - `GET /api/v1/machine/cupWarmer` returns `{"temperature":0.0,"currentTemperature":null,...}`; `PUT {"temperature":45.0}` returns `200` and the subsequent `GET` echoes `45.0`.
- What you personally verified and how: the bytes on the wire, through `FakeBleTransport`. The register facts (addresses, `mult`, permissions, firmware maxima) come from the pinned contract firmware's the firmware register table and are locked by the contract checker.
- Edge cases checked: over-range setpoint clamps on the wire; `0.0` disables via `CupWarmerMode`; reconnect re-assert; **no** re-assert when the warmer was never enabled; a plain DE1 never sees either register; milk target clamps at 0 and 85; `MatCurrentTemp` raw `0` and a transport failure both map to `null`.
- What you did **not** verify: **nothing here has been run against a real Bengle.** In particular I have not confirmed that a real mat reaches the temperature asked for, nor that a real milk probe triggers the firmware's autonomous stop - the live probe reading is still serialised as `0` by the current firmware, so the `probeAttached` path cannot be exercised end to end today. **The cup warmer is a heater and the app is its only clamp; this wants a hardware smoke test before merge.**

### Evidence

- [x] Test output (2188 pass; the rewritten steam-stop and cup-warmer tests fail against the old encodings)
- [ ] Log snippets
- [x] curl output (the cup-warmer `GET`/`PUT` round-trip, against `simulate=1`)

## Compatibility & Migration

- Backward compatible? `Yes`. `GET /api/v1/machine/cupWarmer` adds a key; well-behaved clients ignore unknown keys. `SteamSettings.stopAtTemperature` widens from 0..80 to 0..85 - a widening, not a break. No path, method or status code changes.

A machine that a previous app build left with a stale `MatSetPoint` will be corrected on the next `PUT`. Nothing needs to be done by hand.

- **The connect-time re-assert is conditional, on purpose.** `CupWarmerMode` is `PERM_RW`, not `PERM_RWD` - the firmware deliberately does **not** persist it, so a machine can never boot with the mat silently heating, and it resets to 0 on every boot. That means the app has to re-push it on every connect or an enabled warmer silently dies at the next power cycle. But the re-assert fires only when this app instance has a positive cached target. An app that never touched the warmer writes nothing, so it cannot stomp state set by another client sharing the same machine. The cost of this choice: a fresh app install does not know the machine's remembered setpoint (`MatSetPoint` **is** persisted, but the app does not read it back at connect to seed the cache), so the warmer stays off until the user sets it once. I think that is the right trade for a heater. Tell me if you disagree.
- **A raw `MatCurrentTemp` of `0` is reported as `null`, not `0.0`.** Raw zero is how the firmware reports an open or shorted NTC, and it is indistinguishable from a genuine reading of 0 C. I chose to surface `null` ("no valid reading") so a UI renders a placeholder rather than a confident, wrong `0.0`. The cost is that a mat that really is at 0 C - a cold machine in a cold room - also reads `null`. I would rather a placeholder than fake data, but this is a judgement call and it is worth a second opinion.
- **The milk clamp is 85, while the rest of the thermal surface uses 80.** That is the firmware's own ceiling for `TargetMilkTemp` (max 850 deci-degrees). And note the asymmetric scaling that goes with it: this **target** is 10x on the wire, while the live `MilkTemp` reading that arrives on the `0xA013` shot-sample stream is divided by 100. Confusing the two puts the auto-stop off by an order of magnitude. The code comments say so loudly for exactly that reason.
- **The milk target is written even when no probe is attached.** The firmware stops autonomously the moment a probe is physically plugged in, so the target has to already be current when that happens. Gating the write on probe presence would mean the first steam after plugging a probe in used a stale target.

## Risks & Mitigations

- **Risk:** The app is the sole clamp on a heating element. If the 80 C ceiling is ever removed or a scale is reintroduced, the mat is told to go somewhere it should not.
  - **Mitigation:** The clamp is asserted on the wire at both ends in `bengle_cup_warmer_test.dart`, and the contract checker independently fails CI if the declared `min`/`max`/scale drifts from `assets/api/bengle_hw_v1.yml`. Two independent tests would have to be deleted for this to regress.
- **Risk:** `SteamSequencer.useFwAutonomousStop` now returns true where it previously always returned false, so the **app stops stopping the steam** and hands that to the firmware. If the firmware's autonomous stop misbehaves, there is no longer an app-side backstop for that path.
  - **Mitigation:** The predicate requires all of: a Bengle, a probe reported attached, and a positive target. The current firmware serialises the probe reading as `0`, so `probeAttached` is false today and the predicate cannot fire on real hardware yet - the behaviour change is latent until the firmware ships the probe reading. That is also, honestly, why it is untested against hardware. The reason for the change is real, though: double-stopping races the state machine.
- **Risk:** A client renders `currentTemperature: null` as `0.0`.
  - **Mitigation:** `rest_v1.yml` marks it nullable and read-only and the description says what `null` means. The e2e scenario checks the placeholder path explicitly.

### DIFF
--- a/.agents/skills/decent-app/scenarios/bengle-cup-warmer.md
+++ b/.agents/skills/decent-app/scenarios/bengle-cup-warmer.md
@@ -1,12 +1,14 @@
 # Scenario: Bengle cup-warmer + capability discovery
 
-Verifies the first Bengle peripheral surface end-to-end: `GET /api/v1/machine/capabilities` lists `cupWarmer` when a Bengle is connected, `GET /api/v1/machine/cupWarmer` returns the current setpoint, `POST /api/v1/machine/cupWarmer` writes a new setpoint, and out-of-range writes are rejected at 400 before reaching the device.
+Verifies the first Bengle peripheral surface end-to-end: `GET /api/v1/machine/capabilities` lists `cupWarmer` when a Bengle is connected, `GET /api/v1/machine/cupWarmer` returns the current setpoint plus the live mat temperature (`currentTemperature`, `null` when the firmware has no valid reading), `PUT /api/v1/machine/cupWarmer` writes a new setpoint, and out-of-range writes are rejected at 400 before reaching the device.
 
 ## Preconditions
 
 ```bash
-scripts/sb-dev.sh start --connect-machine MockBengle --connect-scale MockScale
+scripts/sb-dev.sh start --connect-machine MockBengle
 ```
+
+(No `--connect-scale`: on Bengle the integrated scale always wins and `preferredScaleId` is ignored.)
 
 ## Steps
 
@@ -18,24 +20,24 @@ curl -sf http://localhost:8080/api/v1/machine/capabilities | jq -e '.capabilitie
 
 Exit 0 → `cupWarmer` present in the capability list.
 
-### 2. Read initial setpoint (off)
+### 2. Read initial setpoint (off) + live mat temperature placeholder
 
 ```bash
-curl -sf http://localhost:8080/api/v1/machine/cupWarmer | jq -e '.temperature == 0'
+curl -sf http://localhost:8080/api/v1/machine/cupWarmer | jq -e '.temperature == 0 and .currentTemperature == null'
 ```
 
-MockBengle boots with `_cupWarmerTemp = 0.0`. Exit 0 → off.
+MockBengle boots with `_cupWarmerTemp = 0.0` and no simulated mat reading — `currentTemperature` is `null` (the "no valid reading" placeholder case; field firmware without the MatCurrentTemp register behaves the same). Exit 0 → off + placeholder.
 
 ### 3. Set a valid setpoint
 
 ```bash
-curl -sf -X POST http://localhost:8080/api/v1/machine/cupWarmer \
+curl -sf -X PUT http://localhost:8080/api/v1/machine/cupWarmer \
   -H 'Content-Type: application/json' \
   -d '{"temperature": 60}' \
   -o /dev/null -w '%{http_code}\n'
 ```
 
-Expected: `202`.
+Expected: `200`. (Also enables the warmer — a target `> 0` = on; the FW enable register is app-managed.)
 
 ### 4. Confirm the new setpoint
 
@@ -48,7 +50,7 @@ Exit 0 → MockBengle stored the value.
 ### 5. Reject out-of-range
 
 ```bash
-curl -s -X POST http://localhost:8080/api/v1/machine/cupWarmer \
+curl -s -X PUT http://localhost:8080/api/v1/machine/cupWarmer \
   -H 'Content-Type: application/json' \
   -d '{"temperature": 100}' \
   -o /dev/null -w '%{http_code}\n'
--- a/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
+++ b/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
@@ -50,6 +50,8 @@ curl -s -X PUT http://localhost:8080/api/v1/scale/tare
 
 Expected: `200 OK`, empty body.
 
+On real hardware this generic endpoint reaches the machine's `ScaleTare` MMR write-trigger (`0x0080388C`) through `BengleVirtualScale.tare()` → `tareIntegratedScale()`; the firmware performs an immediate tare and subsequent `0xA013` weight arrives already net of the new zero. Confirm a tare by watching the weight drop toward 0 — never the `0xA013` Flags bit.
+
 Re-sample the snapshot and confirm weight is ~0 (MockBengle resets `_tareOffset` to the current accumulated weight, so the next emission reads near zero):
 
 ```bash
--- a/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
+++ b/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
@@ -21,7 +21,7 @@ curl -sf http://localhost:8080/api/v1/machine/capabilities | jq .
 Expected:
 
 ```json
-{ "capabilities": ["cupWarmer", "integratedScale", "ledStrip", "stopAtWeight"] }
+{ "capabilities": ["cupWarmer", "integratedScale", "ledStrip", "stopAtWeight", "scaleCalibration"] }
 ```
 
 Quick assertions:
--- a/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
+++ b/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
@@ -1,6 +1,6 @@
 # Scenario: Bengle integrated scale end-to-end
 
-Verifies that when a Bengle is the connected machine, the integrated scale is auto-attached as a virtual scale (no external scale connection needed), capability discovery advertises `integratedScale`, the `/api/v1/scale/*` REST surface and `/ws/v1/scale/snapshot` stream both flow through the integrated scale, and software stop-at-weight (SAW) ends an espresso shot when the integrated scale weight crosses the profile target.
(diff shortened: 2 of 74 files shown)


tadelv:
Superseded by #612 (merged) plus #601 (TargetMilkTemp) and #605 (MatSetPoint). #612 adds cup-warmer manual mode, live mat temperature, and scheduled pre-warm. No auto-re-enable on reconnect (mode is RAM-only and boots off by design).

PR #466, opened 2026-07-16
feat(bengle): upload v2 shot profiles, not v1

ChampionDesigns:
> **Stacked PR — B-8 of 10.** Builds on #465 (`feat/bengle-thermal`). Until that merges this PR's diff includes its commits; please review/merge in order **B-1 → B-10**.

Today, no profile can be uploaded to a Bengle, and therefore no shot can be pulled on one. reaprime serialises shot descriptors in the DE1's v1 wire format, and Bengle firmware is v2-only: it checks the header version byte, and if it is not `2` it memsets the header and returns, silently dropping the entire profile. There is no error, no rejection, nothing on the wire to notice - the machine simply has no profile, and every shot request goes nowhere. Underneath that, the two formats disagree about how flow and pressure are encoded: v1 packs them as `byte / 16`, v2 as `byte * 0.1`. The frames are byte-width-identical, so nothing catches the mismatch structurally - a v1-encoded 6 ml/s flow decodes on a Bengle as 9.6 ml/s, a 60 percent over-command, and anything above 15.9 ml/s wraps around the byte. This PR branches the encoder on the machine it is talking to.

## Summary

- **Problem:** `UnifiedDe1Profile._writeHeader` hard-coded the header version byte to `1`, and the frame encoders hard-coded the v1 `value * 16` flow/pressure scale. Bengle firmware rejects any header whose version is not `2` and decodes flow and pressure as **U8D1** (`byte * 0.1`, range 0 to 25.5), not v1's **U8P4** (`byte / 16`, range 0 to 15.9375).
- **Why it matters:** This is the difference between a Bengle that can pull a shot and one that cannot. The profile never lands, so there is nothing to run. And had the header been accepted, the scale mismatch would have been worse than the drop: every flow and pressure target in the profile would have been silently over-commanded by 60 percent.
- **What changed:** The encoder is gated on the runtime `isBengle` flag (`v13Model >= 128`, resolved at connect in B-1). A Bengle gets header version `2` and U8D1 encoding for `SetVal`, `TriggerVal`, the header's minimum-pressure and maximum-flow bytes, and the extension frame's limiter value and range. The global max-flow ceiling rises from the DE1's 8 ml/s to 20 ml/s on a Bengle, mirroring de1plus, which raises `max_flowrate` to 20 when BLE v2 is negotiated; flow-priority targets are clamped to it before encoding. The hard-coded `12 * 16` header byte is replaced by a properly encoded value.
- **What did NOT change (scope boundary):** The DE1 path is **byte-for-byte identical**, including its existing mod-256 wrap above 15.9375 (there is a test that witnesses the wrap, deliberately). No REST endpoint and no WebSocket topic changed - `rest_v1.yml` and `websocket_v1.yml` are untouched. Profile storage, parsing and the profile model are all unchanged; this is purely the device-upload encoder.

## Change Type (select all)

- [x] Bug fix
- [x] Feature
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [ ] REST API / handlers
- [ ] WebSocket API
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

- Closes #

## Root Cause (if bug fix)

- **Root cause:** reaprime was written for the DE1 and its profile encoder had exactly one wire format hard-coded into it, with no concept of a protocol version. When Bengle support was added, the encoder was never revisited. The header version byte and both flow/pressure scales were literal constants inside `_writeHeader` and `_writeFrame`.
- **Missing detection or guardrail:** The two formats produce frames of **identical byte width**, so nothing structural can catch the mismatch - not a length check, not a parse. The only guard is the header version byte, and the firmware's response to a wrong one is to silently discard rather than to complain. A silent-drop failure with no error path is invisible to the app by construction.
- **Contributing context:** The DE1's own v1 encoder has a latent mod-256 wrap above 15.9375 (`(0.5 + value * 16).toInt()` on a byte). It never bites, because DE1 flow and pressure stay well inside range. It is preserved here, deliberately and with a test, so that this PR does not quietly change DE1 behaviour while fixing Bengle.

  - [ ] Unit test
- **Target test or file:** `test/models/device/unified_de1_bengle_profile_v2_test.dart` (new, 272 lines) and `test/unit/models/device/impl/bengle/bengle_firmware_prelude_test.dart` (+38 lines).
- **Scenario the test should lock in:** Drive a real `UnifiedDe1` over `FakeBleTransport` and assert the **captured header and frame bytes** for both machine classes. On a Bengle: header byte 0 is `2`; a 6 ml/s flow target encodes as `0x3C`; 20 ml/s encodes as `0xC8` **without wrapping**; a 30 ml/s target clamps to 20; a 9 bar pressure target encodes as `0x5A`. On a DE1: the header still says `1` and the same profile still produces the old v1 bytes, wrap included. Plus `convert_float_to_U8D1` boundary units (saturation at 0 and 255, round-half-up), and a 500 ms DFU-prelude spacing test with a wall-clock lower bound.

- [ ] API spec updated: `assets/api/rest_v1.yml` / `websocket_v1.yml` - **not applicable, and deliberately untouched.** The wire format between the app and the machine is not part of the app's public HTTP API; no endpoint, request shape or response shape changes.
- [x] Profile docs updated: `doc/Profiles.md` - new "device upload wire format" section covering the v1/v2 split, the U8P4 and U8D1 byte tables, and the 20 ml/s Bengle ceiling.
- [ ] N/A - no docs affected

the contract file header section 5.6 gains a pointer to the profile endpoint and the byte table, and states the matched-pair deployment requirement (below).

## Security Impact (required)

- BLE/USB surface changed? `Yes` - the bytes written to the profile header and frame endpoints change **for Bengle machines only**. No new endpoint or characteristic is used; the same writes carry different, correct content. The DE1 byte stream is unchanged.

The 20 ml/s ceiling is enforced **in the encoder**, because reaprime is headless - there is no UI slider bounding a flow input, and a profile can arrive over REST or as an imported file. The encoder is the last place before the wire, so it is the enforcement point. Values above 25.5 saturate rather than wrap, so no input can be crafted that wraps a flow byte around into a small number on the Bengle path.

## User-Visible Changes

Profiles upload to a Bengle, and shots run on a Bengle. Neither was possible before. Bengle profiles can specify flow up to 20 ml/s (a DE1 remains capped at 8); a flow-priority step asking for more is clamped to 20 rather than being sent as a wrapped byte. Nothing changes for DE1 users - the bytes on their wire are identical, byte for byte.

- [x] `flutter test` - **2213 tests pass** on this branch at `7f2f015c` (B-7 was 2188, so this branch adds 25). Note: the commit message says 2179, which was an intermediate figure taken before the branch was rebuilt onto its final base; the gate log is the number to trust.

- OS / platform tested: Linux (analyzer and tests).
- Simulated devices? (`simulate=1`): `No` - the simulated machine does not decode profile bytes, so a simulated run would prove nothing about this change.
- What you personally verified and how: the exact bytes, captured off `FakeBleTransport` and asserted against both wire formats. The firmware facts (header-version rejection, U8D1 decoding) come from reading the firmware source at the pinned commit, not from experiment.
- Edge cases checked: the U8D1 clamp at both ends; the 20 ml/s ceiling applied to a flow-priority target before encoding; a pressure target riding the 25.5 clamp rather than the flow ceiling; the DE1 path's mod-256 wrap explicitly witnessed so a future change to it cannot be accidental.
- What you did **not** verify: **no profile from this branch has been uploaded to a real Bengle.** This is the single most important thing in the stack to smoke-test on hardware, because it is the thing that decides whether the machine can make coffee. It has not been done, and the tests, being byte assertions against my own reading of the firmware, cannot substitute for it.

### Evidence

- [x] Test output (2213 pass; the new byte-exact upload test fails against the old encoder for the Bengle case and passes for the DE1 case both before and after)
- [ ] Log snippets

## Compatibility & Migration

- Backward compatible? `Yes` for the DE1 - byte-for-byte identical output. For the Bengle it is a **breaking change against old firmware**, by necessity: see the risk below.

- **The version branch lives inside the encoder, not in a subclass.** `_writeHeader` and the frame encoders are private static extension methods on `UnifiedDe1`; a `Bengle` subclass cannot override them. Rather than restructure the profile encoder into something overridable (a much larger diff on a path shared with every DE1 user), I branched on the existing runtime `isBengle` flag inside the three encode sites and pulled the shared decision into one helper, `_encodeFlowPressure`. If you would rather see a proper strategy object here, that is a reasonable ask and I will do it - it is just a bigger blast radius on DE1 code than this fix needs.
- **The Bengle path clamps; the DE1 path still wraps.** Only `convert_float_to_U8D1` saturates. The v1 `(0.5 + value * 16).toInt()` path is untouched and still wraps mod 256 above 15.9375. I could have clamped both, and arguably should - but that changes DE1 bytes, which is exactly what this PR promises not to do. There is a test that pins the wrap as a **witness**, not an endorsement, so that a future PR that fixes it does so on purpose. I would support that being a separate PR.
- **The 20 ml/s ceiling is a policy decision copied from de1plus.** de1plus raises `max_flowrate` to 20 when BLE v2 is negotiated. reaprime has no equivalent setting, so I hard-coded the constant with a comment. If Bengle firmware ever raises its own ceiling, this is the line to change.
- **A nit I did not want to fix silently:** `unified_de1.profile.dart` refers to `BENGLE_FIXES.md` in two comments, and that file does not exist anywhere in the repository. It is a workspace artefact that leaked into a shipped source comment. It should be scrubbed or replaced with a pointer to the contract file header before this merges. I am flagging it rather than fixing it here so you can see what the commit actually contains.

## Risks & Mitigations

- **Risk (the big one): the app and the firmware must deploy as a matched pair.** The encoder decides the wire version from `isBengle` - the machine **model**, not the machine's firmware version. There is no version negotiation and no capability bit to key off. So an app carrying this change, talking to an older Bengle whose firmware is v1-only, will send v2 profiles that the old firmware mis-decodes - and because the frames are byte-width-identical, it will mis-decode them **silently**, at a 60 percent scale error on every flow and pressure target rather than refusing them.
  - **Mitigation:** the contract file header section 5.6 states the matched-pair requirement explicitly. That is documentation, not a guardrail, and I want to be straight about that: **there is no runtime check.** If you know of a firmware-version field that is readable at connect and reliable across the fleet, this should key off it instead, and I will happily add it. 
- **Risk:** A profile with a flow target above 20 ml/s is silently clamped rather than rejected.
  - **Mitigation:** This mirrors de1plus and is documented in `doc/Profiles.md`. The clamp is asserted in the tests at 30 ml/s. Rejecting instead would mean a REST-imported profile could fail to upload with no UI to report it in, which is worse.

### DIFF
--- a/.agents/skills/decent-app/scenarios/bengle-cup-warmer.md
+++ b/.agents/skills/decent-app/scenarios/bengle-cup-warmer.md
@@ -1,12 +1,14 @@
 # Scenario: Bengle cup-warmer + capability discovery
 
-Verifies the first Bengle peripheral surface end-to-end: `GET /api/v1/machine/capabilities` lists `cupWarmer` when a Bengle is connected, `GET /api/v1/machine/cupWarmer` returns the current setpoint, `POST /api/v1/machine/cupWarmer` writes a new setpoint, and out-of-range writes are rejected at 400 before reaching the device.
+Verifies the first Bengle peripheral surface end-to-end: `GET /api/v1/machine/capabilities` lists `cupWarmer` when a Bengle is connected, `GET /api/v1/machine/cupWarmer` returns the current setpoint plus the live mat temperature (`currentTemperature`, `null` when the firmware has no valid reading), `PUT /api/v1/machine/cupWarmer` writes a new setpoint, and out-of-range writes are rejected at 400 before reaching the device.
 
 ## Preconditions
 
 ```bash
-scripts/sb-dev.sh start --connect-machine MockBengle --connect-scale MockScale
+scripts/sb-dev.sh start --connect-machine MockBengle
 ```
+
+(No `--connect-scale`: on Bengle the integrated scale always wins and `preferredScaleId` is ignored.)
 
 ## Steps
 
@@ -18,24 +20,24 @@ curl -sf http://localhost:8080/api/v1/machine/capabilities | jq -e '.capabilitie
 
 Exit 0 → `cupWarmer` present in the capability list.
 
-### 2. Read initial setpoint (off)
+### 2. Read initial setpoint (off) + live mat temperature placeholder
 
 ```bash
-curl -sf http://localhost:8080/api/v1/machine/cupWarmer | jq -e '.temperature == 0'
+curl -sf http://localhost:8080/api/v1/machine/cupWarmer | jq -e '.temperature == 0 and .currentTemperature == null'
 ```
 
-MockBengle boots with `_cupWarmerTemp = 0.0`. Exit 0 → off.
+MockBengle boots with `_cupWarmerTemp = 0.0` and no simulated mat reading — `currentTemperature` is `null` (the "no valid reading" placeholder case; field firmware without the MatCurrentTemp register behaves the same). Exit 0 → off + placeholder.
 
 ### 3. Set a valid setpoint
 
 ```bash
-curl -sf -X POST http://localhost:8080/api/v1/machine/cupWarmer \
+curl -sf -X PUT http://localhost:8080/api/v1/machine/cupWarmer \
   -H 'Content-Type: application/json' \
   -d '{"temperature": 60}' \
   -o /dev/null -w '%{http_code}\n'
 ```
 
-Expected: `202`.
+Expected: `200`. (Also enables the warmer — a target `> 0` = on; the FW enable register is app-managed.)
 
 ### 4. Confirm the new setpoint
 
@@ -48,7 +50,7 @@ Exit 0 → MockBengle stored the value.
 ### 5. Reject out-of-range
 
 ```bash
-curl -s -X POST http://localhost:8080/api/v1/machine/cupWarmer \
+curl -s -X PUT http://localhost:8080/api/v1/machine/cupWarmer \
   -H 'Content-Type: application/json' \
   -d '{"temperature": 100}' \
   -o /dev/null -w '%{http_code}\n'
--- a/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
+++ b/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
@@ -50,6 +50,8 @@ curl -s -X PUT http://localhost:8080/api/v1/scale/tare
 
 Expected: `200 OK`, empty body.
 
+On real hardware this generic endpoint reaches the machine's `ScaleTare` MMR write-trigger (`0x0080388C`) through `BengleVirtualScale.tare()` → `tareIntegratedScale()`; the firmware performs an immediate tare and subsequent `0xA013` weight arrives already net of the new zero. Confirm a tare by watching the weight drop toward 0 — never the `0xA013` Flags bit.
+
 Re-sample the snapshot and confirm weight is ~0 (MockBengle resets `_tareOffset` to the current accumulated weight, so the next emission reads near zero):
 
 ```bash
--- a/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
+++ b/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
@@ -21,7 +21,7 @@ curl -sf http://localhost:8080/api/v1/machine/capabilities | jq .
 Expected:
 
 ```json
-{ "capabilities": ["cupWarmer", "integratedScale", "ledStrip", "stopAtWeight"] }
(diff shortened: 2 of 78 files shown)


tadelv:
closing in favor of #486

PR #467, opened 2026-07-16
feat(bengle): sync the wake schedule and sleep timeout to the machine

ChampionDesigns:
> **Stacked PR — B-9 of 10.** Builds on #466 (`feat/bengle-profile-v2-upload`). Until that merges this PR's diff includes its commits; please review/merge in order **B-1 → B-10**.

A Bengle can run its own wake schedule with no tablet in the room - the firmware holds a weekly table of wake windows and a wall clock, and wakes the machine on its own. Nothing in the app had ever written to them. On a bench machine all three registers read zero: no clock, no entries, schedule disabled. So the "wake at 6:30 on weekdays" feature that the app appears to offer was, in fact, a 30-second timer inside the app, running off the tablet's clock, and it only worked if the tablet was awake, connected, and running the app. Leave the tablet at your desk and your machine is cold when you get to it. This PR pushes the wall clock, the wake windows and the sleep timeout into the machine's own registers, so the firmware runs the schedule itself. It also refuses, categorically, to write a sleep timeout of zero - because that register is the machine's thermal safety net, not a user preference.

## Summary

- **Problem:** The Bengle firmware owns four registers for autonomous scheduling - `InactivitySleepTimeout`, `SetLocalTimeOfWeek`, `ScheduleEntry` and `ScheduleControl` (register-table rows 54 to 57). The app wrote none of them, so the machine's own scheduler never ran. The app-side wake schedule was a `PresenceController` timer that required a live tablet.
- **Why it matters:** The whole point of a machine-side scheduler is that it works when the tablet is not there. This is also the only path by which a machine can heat up before you walk into the kitchen.
- **What changed:** A new `BengleScheduleSync` controller watches the settings and the connected machine and pushes the local time-of-week, the wake-window table and the sleep timeout into the machine. It re-pushes on every connect, because the clock and the table are **RAM-only in firmware and do not survive a power cycle** - the app is the only durable store for them. A new pure module, `wake_schedule_windows.dart`, derives concrete windows from the app's `WakeSchedule` list in the firmware's own semantics, and `sleep_timeout_safety.dart` decides what may be written into the sleep-timeout register.
- **What did NOT change (scope boundary):** No REST endpoint, no WebSocket topic, no settings-UI change. This PR reads settings that already exist and pushes them to a machine that already had the registers. It is **Bengle-gated**: a plain DE1 connecting causes not one byte to go on the wire, and `PresenceController`'s app-side wake and sleep behaviour is untouched for it.

## Change Type (select all)

- [ ] Bug fix
- [x] Feature
- [ ] Docs
- [x] Security hardening (the sleep-timeout floor; see below)
- [ ] Chore / infra

- [ ] REST API / handlers
- [ ] WebSocket API
- [x] Machine state / shot logic (machine sleep and wake)
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [ ] Docs / specs

## Linked Issues

- Closes #

## Root Cause (if bug fix)

N/A - this is a new capability, not a bug fix. The nearest thing to a root cause is that the firmware's scheduling registers were published and never consumed.

N/A as a regression plan - but the new code is covered, and here is what by:

- **`test/controllers/bengle_schedule_sync_test.dart`** (741 lines): the push on connect; the re-push after a power cycle (detected by the firmware clock reading back as `0`, the "rebooted, never synced" sentinel); the sleep-timeout floor, including the `0`, the negative, and the master-toggle-off cases; and a machine swapped **mid-write**, where the drain re-targets the new machine rather than finishing the write against the old one and pushing nothing to the machine that is actually there.
- **`test/models/device/bengle_wake_schedule_test.dart`** (176 lines): the register writes themselves.
- **`test/models/wake_schedule_windows_test.dart`** (41 cases): window derivation, midnight-crossing splits, the 32-entry cap and merge, and two DST spring-forward regression cases.
- **`test/settings/sleep_timeout_safety_test.dart`** (79 lines).
- **`test/unit/models/device/impl/bengle/mmr_contract_test.dart`**: rows 54 to 57 registered with the firmware contract checker.

## Merge-order note - please read this before merging

This branch ships two modules that are **also introduced by two of the independent PRs** going to `main`. Each side adds the file from nothing, because each has to stand alone:

| File (+ its test) | Also introduced by | Resolution when the second one merges |
|---|---|---|
| `lib/src/models/wake_schedule_windows.dart` | `fix/presence-keep-awake-window` (I-E) | **Take either - the copies are byte-identical.** |
| `lib/src/settings/sleep_timeout_safety.dart` | `fix/settings-bound-sleep-timeout` (I-F) | **Hand-written UNION. Do not pick a side.** |

**Git does not de-duplicate these for you.** Measured with `git merge-tree --write-tree` against both branches, all four paths come back as **`CONFLICT (add/add)`**. Whichever PR merges second will land on them.

### `wake_schedule_windows.dart` + its test - take either, they are byte-identical

This branch's copy of both the lib file and its test is **byte-identical** to I-E's (verified: `md5sum` matches on both files against `fix/presence-keep-awake-window`). So the add/add conflict is trivial - take either side. There is nothing to choose between them.

This was not always true, and the history is worth one sentence so it stays fixed. An earlier cut of this branch carried an older `windowEnd()` that computed a window's end by adding an absolute `Duration`, which on a spring-forward day lands an hour off the wall clock (a 05:30-07:00 window reports 08:00), and it was missing the two DST regression tests that pin the corrected form. On this branch alone that was dormant - nothing here calls `windowEnd()`, since `bengle_schedule_sync.dart` uses only `expandWindows`, `packWakeWindow` and `localSecondsOfWeek`. But it would **not** have been dormant in the merged tree: it would have landed straight in the `keepAwakeUntil` that I-E ships. This branch's copy has since been rebuilt from I-E's, so both the corrected `windowEnd()` and its two DST tests are now present here (the test count is 41, up from 39). If a future edit makes the two copies drift apart again, stop and re-sync them - they are meant to be identical.

**Verified, not asserted:** the two files hash-match across the branches, this branch is `flutter analyze`-clean, and the full suite passes at 2295 (the two DST cases are the +2 over the previous 2293).

### `sleep_timeout_safety.dart` + its test - union, and mind the duplicate

The two copies are complementary halves of one file, split by symbol. Taking either side alone drops symbols the other PR's code imports, and the build breaks.

| Symbol | Declared by |
|---|---|
| `kSafetySleepFloorMinutes`, `kMinMachineSleepTimeoutMinutes`, `machineSleepTimeoutMinutes` | **this branch** (what may be written into the *machine's* safety register) |
| `kMinSleepTimeoutSetting`, `kMaxSleepTimeoutSetting`, `isValidSleepTimeoutSetting`, `sanitizeSleepTimeoutSetting` | **I-F** (bounds on the *stored preference*) |
| `kMaxSleepTimeoutMinutes` | **both** - and that is the trap |

Both copies declare `const int kMaxSleepTimeoutMinutes = 240;` (same value, different doc comment), and both carry their own `library;` directive with its own library doc block. **A naive concatenation of the two sides will not compile** - duplicate top-level declaration, duplicate `library;`. Resolve it like this:

1. Take **this branch's copy in full** (its library doc block, `kSafetySleepFloorMinutes`, `kMinMachineSleepTimeoutMinutes`, `kMaxSleepTimeoutMinutes`, `machineSleepTimeoutMinutes`).
2. Append **only these four symbols** from I-F's copy: `kMinSleepTimeoutSetting`, `kMaxSleepTimeoutSetting`, `isValidSleepTimeoutSetting`, `sanitizeSleepTimeoutSetting`.
3. Drop I-F's `library;` directive, its library doc block and its `kMaxSleepTimeoutMinutes` declaration - all three are duplicates.

The result has exactly one `library;` and one declaration of each of the eight symbols, and it is the file the two halves were split out of.

**The test file resolves the same way, one level down.** Both copies are a `void main()` with non-overlapping `group()`s - this branch contributes `machineSleepTimeoutMinutes: never returns 0`; I-F contributes `sanitizeSleepTimeoutSetting: bounds untrusted REST/import input` and `SettingsController.setSleepTimeoutMinutes bounds what it stores`. Keep **one** `void main()` containing all three groups, and take **I-F's import block** - it is a superset of this branch's.

**Verified, not asserted:** this exact union was built and merged (this branch's lib copy in full + I-F's four preference symbols; one `void main()` with all three test groups). The resulting tree is `flutter analyze`-clean and the full suite passes (2300/2300), with both PRs' consumers compiling against it.

If the independent PRs are not merged, this branch is self-contained and none of the above applies.

- [ ] API spec updated: `assets/api/rest_v1.yml` / `websocket_v1.yml`
- [x] N/A - no docs affected

**This branch touches zero files under `doc/` and zero API specs**, and I want to be explicit about why rather than just ticking N/A. No REST endpoint, WebSocket topic, plugin event, skin behaviour or profile handling changes. The four registers already exist in `assets/api/bengle_hw_v1.yml` at `contract_version: 1`; this PR adds app-side **registration entries** for them, so the contract file is not edited and the version is not bumped.

That said, a reviewer could reasonably argue that "the machine now runs its own wake schedule, and the app clamps the machine's sleep timeout to a 60-minute floor" is a device-behaviour change that belongs in `doc/DeviceManagement.md`. I did not write it because I was not sure where you would want it. **If you want that doc, say so and I will add it.**

## Security Impact (required)

- BLE/USB surface changed? `Yes` - four Bengle-only MMR registers (rows 54 to 57), all registered with the contract checker. A non-Bengle DE1 gets nothing on the wire.

**The safety-relevant part.** `InactivitySleepTimeout` is the timer that turns a Bengle's heaters off when nobody is there - a dead tablet battery, a crashed app, a blackout. Two firmware facts make writing `0` into it unacceptable rather than merely unwise:

1. The firmware treats a value less than or equal to zero as "never sleep". The timer simply never runs.
2. **The write sticks.** The register is `PERM_RWD` - disk-backed in firmware NVM and restored at every boot. A `0` written once survives every power cycle until something writes a non-zero value back. And the machine boots hot.

So an app that wrote `0` would leave a machine **less safe than one that had never met the app at all**, since the firmware's own default is 60 minutes. That is a regression against doing nothing. `machineSleepTimeoutMinutes()` therefore collapses everything that would disable the net - the master toggle off, a "Disabled" (`0`) in the dropdown, a rogue negative arriving from REST or an imported settings blob - to a **60-minute floor, which is the firmware's own default**, and clamps everything else into the 1-to-240 range the firmware accepts. The function is **guaranteed to return a value in 1 to 240, never 0, whatever the inputs**.

This costs the user nothing they can perceive. The firmware ignores this timer entirely while a tablet is connected, so it only ever acts once the tablet is already gone - precisely the case it exists for.

## User-Visible Changes

A Bengle now honours its wake schedule with the tablet switched off, closed, or in another room. Previously the schedule only worked while the app was running and connected.

The one thing a user could notice as a change rather than an addition: **the "Disabled" option for the sleep timeout no longer disables the machine's own sleep.** It still disables the app's idle-sleep timer, which is what the setting is for, but the machine will still put itself to sleep after 60 minutes of nobody touching it once the tablet is gone. A user who deliberately wanted an always-hot machine with no tablet attached cannot get that from this app. That is intentional and I would defend it, but it is a real behaviour change and it should not be buried.

- [x] `flutter test` - **2295 tests pass** on this branch at `35f630eb` (B-8 was 2213, so this branch adds 82).

- OS / platform tested: Linux (analyzer and tests).
- Simulated devices? (`simulate=1`): `No`. `MockBengle` accepts the schedule writes but has no clock and no scheduler, so a simulated run would only re-confirm what the unit tests already assert.
- Real hardware? (DE1/Bengle/scale): `No` for this branch's code. The **premise** was established on hardware: all three RAM-only registers were read on a bench Bengle and all three came back zero, which is what "nothing was writing them" means in practice.
- What you personally verified and how: the writes and the sequencing, through the tests. The firmware semantics (the day field is an index and not a bitmask, `startMin` inclusive and `endMin` exclusive, `startMin >= endMin` silently dropped, the 32-entry table, the write-then-echo read behaviour, the boot-time zeroing) come from reading the firmware source at the pinned commit.
- Edge cases checked: a table with a midnight-crossing window (split app-side, because the firmware drops it otherwise); more than 32 windows (merged and capped app-side, because the firmware silently drops the 33rd); a machine that reboots mid-session (the 15-minute tick sees the clock read back `0` and re-pushes); a machine swapped mid-write; every path by which the sleep timeout could become `0`.
- What you did **not** verify: **no machine has actually woken itself up from a schedule pushed by this code.** That is the entire feature and it is unproven end to end. The write protocol, the packing and the register semantics are all inferred from the firmware source and asserted in unit tests, but a Bengle sitting on a bench overnight, waking at a scheduled time with the tablet powered off, is the test that matters and it has not been run. I also could not verify clock drift, because **no register exposes the running firmware clock** - the four registers are write-driven and a read echoes the last value written, not live device state. The 15-minute clock resync is a mitigation for a drift I cannot measure.

### Evidence

- [x] Test output (2295 pass)
- [ ] Log snippets

## Compatibility & Migration

- Backward compatible? `Yes` for every API surface - nothing public changed. See "User-Visible Changes" for the one behavioural change (the sleep-timeout floor).

A Bengle whose `InactivitySleepTimeout` was left at some other value by a previous app or by hand will be overwritten on the next connect, to the user's setting, or to 60 if their setting would disable it.

- **The 60-minute floor is a safety decision that overrides a user setting.** I want it looked at squarely rather than skimmed. The argument for it is in the Security Impact section above; the short version is that a thermal safety net must not be a UI preference, and 60 is not a number I chose - it is the firmware's own default, so the app can never make a machine less safe than a factory-fresh one. The argument against it is that a user asked for something and did not get it. I think the asymmetry settles it: the failure mode of the floor is a machine that goes to sleep when someone wanted it hot; the failure mode of honouring the `0` is a machine that stays hot forever, permanently, because the write persists.
- **The schedule table is re-pushed only when it actually changed, or when the machine demonstrably lost it.** A re-push clears the table first (`ScheduleControl` `0`, then the entries, then `ScheduleControl` `1`), and clearing it drops the firmware's "was in an awake window" edge - so a gratuitous re-push in the middle of a window would **re-wake a machine the user had just manually put to sleep**. Hence the diff, and hence the 15-minute clock tick deliberately never touches the table.
- **A wake-only schedule (no `keepAwakeFor`) becomes a 30-minute window.** The firmware cannot express "wake, then hand straight back to the idle timer" - it rejects a window whose start is not before its end, so a window must have real length. 30 minutes matches the app's default sleep timeout. A one-minute window is semantically purer, but it is a fragile target against any firmware-clock error, and a missed window means the machine never wakes at all. This is a guess at what the user meant, and it is the kind of guess worth a second opinion.
- **Two register bounds are declared narrower than the contract file allows, on purpose.** `SetLocalTimeOfWeek` is capped at 604799, not 604800, because the firmware rejects a value equal to the seconds-in-a-week and would silently leave the clock invalid. `ScheduleControl` is capped at `1`, not 255, because the firmware reads only bit 0 on a non-zero write - so a stray `2` would disable the schedule **without clearing it**. Declaring `max: 1` makes that structurally impossible, since `writeMmrInt` clamps to the declared range.
- **The controller is constructed in `main.dart` and never referenced again** (there is an `// ignore: unused_local_variable`). It self-wires by subscribing to the controller and the settings. That is the pattern the other bridges use, but the lint suppression is ugly and I would take a suggestion.

## Risks & Mitigations

- **Risk:** A future edit lets this branch's `wake_schedule_windows.dart` drift from I-E's again, reintroducing the DST bug through a bad conflict resolution.
  - **Mitigation:** Already handled at the source - this branch's copy of the lib file and its test have been rebuilt from I-E's and are now byte-identical (hash-verified), so the add/add conflict is trivial and the two DST regression tests are present on both sides. The remaining risk is only that someone later edits one copy and not the other; the merge-order note says to stop and re-sync if the two ever differ.
- **Risk:** The firmware clock drifts, or a DST transition passes, and a wake window fires an hour late.
  - **Mitigation:** The clock is re-written every 15 minutes while a Bengle is connected. This bounds DST error and crystal drift, and sits well inside the smallest useful window. It cannot help a machine that is disconnected from the tablet for a long time - which is exactly the case this feature exists for. **The size of that drift is unmeasured**, because no register exposes the running clock.
- **Risk:** A push fails mid-sequence and the machine is left with a cleared, disabled table.
  - **Mitigation:** Every write path is wrapped; a failure logs and retries with capped backoff (3s, 10s, 30s, then 30s repeating), and the desired state is only stamped as pushed **after** the write lands, so a failure during the connect flow cannot crash it and self-heals on the next trigger at the latest. The machine falls back to its `InactivitySleepTimeout`, which this PR guarantees is never `0`.

### DIFF
--- a/.agents/skills/decent-app/scenarios/bengle-cup-warmer.md
+++ b/.agents/skills/decent-app/scenarios/bengle-cup-warmer.md
@@ -1,12 +1,14 @@
 # Scenario: Bengle cup-warmer + capability discovery
 
-Verifies the first Bengle peripheral surface end-to-end: `GET /api/v1/machine/capabilities` lists `cupWarmer` when a Bengle is connected, `GET /api/v1/machine/cupWarmer` returns the current setpoint, `POST /api/v1/machine/cupWarmer` writes a new setpoint, and out-of-range writes are rejected at 400 before reaching the device.
+Verifies the first Bengle peripheral surface end-to-end: `GET /api/v1/machine/capabilities` lists `cupWarmer` when a Bengle is connected, `GET /api/v1/machine/cupWarmer` returns the current setpoint plus the live mat temperature (`currentTemperature`, `null` when the firmware has no valid reading), `PUT /api/v1/machine/cupWarmer` writes a new setpoint, and out-of-range writes are rejected at 400 before reaching the device.
 
 ## Preconditions
 
 ```bash
-scripts/sb-dev.sh start --connect-machine MockBengle --connect-scale MockScale
+scripts/sb-dev.sh start --connect-machine MockBengle
 ```
+
+(No `--connect-scale`: on Bengle the integrated scale always wins and `preferredScaleId` is ignored.)
 
 ## Steps
 
@@ -18,24 +20,24 @@ curl -sf http://localhost:8080/api/v1/machine/capabilities | jq -e '.capabilitie
 
 Exit 0 → `cupWarmer` present in the capability list.
 
-### 2. Read initial setpoint (off)
+### 2. Read initial setpoint (off) + live mat temperature placeholder
 
 ```bash
-curl -sf http://localhost:8080/api/v1/machine/cupWarmer | jq -e '.temperature == 0'
+curl -sf http://localhost:8080/api/v1/machine/cupWarmer | jq -e '.temperature == 0 and .currentTemperature == null'
 ```
 
-MockBengle boots with `_cupWarmerTemp = 0.0`. Exit 0 → off.
+MockBengle boots with `_cupWarmerTemp = 0.0` and no simulated mat reading — `currentTemperature` is `null` (the "no valid reading" placeholder case; field firmware without the MatCurrentTemp register behaves the same). Exit 0 → off + placeholder.
 
 ### 3. Set a valid setpoint
 
 ```bash
-curl -sf -X POST http://localhost:8080/api/v1/machine/cupWarmer \
+curl -sf -X PUT http://localhost:8080/api/v1/machine/cupWarmer \
   -H 'Content-Type: application/json' \
   -d '{"temperature": 60}' \
   -o /dev/null -w '%{http_code}\n'
 ```
 
-Expected: `202`.
+Expected: `200`. (Also enables the warmer — a target `> 0` = on; the FW enable register is app-managed.)
 
 ### 4. Confirm the new setpoint
 
@@ -48,7 +50,7 @@ Exit 0 → MockBengle stored the value.
 ### 5. Reject out-of-range
 
 ```bash
-curl -s -X POST http://localhost:8080/api/v1/machine/cupWarmer \
+curl -s -X PUT http://localhost:8080/api/v1/machine/cupWarmer \
   -H 'Content-Type: application/json' \
   -d '{"temperature": 100}' \
   -o /dev/null -w '%{http_code}\n'
--- a/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
+++ b/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
@@ -50,6 +50,8 @@ curl -s -X PUT http://localhost:8080/api/v1/scale/tare
 
 Expected: `200 OK`, empty body.
 
+On real hardware this generic endpoint reaches the machine's `ScaleTare` MMR write-trigger (`0x0080388C`) through `BengleVirtualScale.tare()` → `tareIntegratedScale()`; the firmware performs an immediate tare and subsequent `0xA013` weight arrives already net of the new zero. Confirm a tare by watching the weight drop toward 0 — never the `0xA013` Flags bit.
+
 Re-sample the snapshot and confirm weight is ~0 (MockBengle resets `_tareOffset` to the current accumulated weight, so the next emission reads near zero):
 
 ```bash
--- a/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
+++ b/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
@@ -21,7 +21,7 @@ curl -sf http://localhost:8080/api/v1/machine/capabilities | jq .
 Expected:
 
 ```json
-{ "capabilities": ["cupWarmer", "integratedScale", "ledStrip", "stopAtWeight"] }
+{ "capabilities": ["cupWarmer", "integratedScale", "ledStrip", "stopAtWeight", "scaleCalibration"] }
 ```
 
 Quick assertions:
--- a/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
+++ b/.agents/skills/decent-app/scenarios/bengle-integrated-scale.md
@@ -1,6 +1,6 @@
 # Scenario: Bengle integrated scale end-to-end
 
-Verifies that when a Bengle is the connected machine, the integrated scale is auto-attached as a virtual scale (no external scale connection needed), capability discovery advertises `integratedScale`, the `/api/v1/scale/*` REST surface and `/ws/v1/scale/snapshot` stream both flow through the integrated scale, and software stop-at-weight (SAW) ends an espresso shot when the integrated scale weight crosses the profile target.
+Verifies that when a Bengle is the connected machine, the integrated scale is auto-attached as a virtual scale (no external scale connection needed), capability discovery advertises `integratedScale` and `stopAtWeight`, the `/api/v1/scale/*` REST surface and `/ws/v1/scale/snapshot` stream both flow through the integrated scale, and **autonomous** stop-at-weight (SAW) ends an espresso shot: the app reflects the workflow's `targetYield` into the machine's SAW target and defers the final yield stop to the machine (`machineHasAutonomousSAW`), which `MockBengle` simulates firmware-side.
 
 ## Preconditions
 
@@ -12,7 +12,7 @@ No `--connect-scale` flag. On Bengle the integrated scale always wins — extern
 
 ## Steps
 
-### 1. Capability discovery lists both Bengle surfaces
+### 1. Capability discovery lists the Bengle surfaces
 
 ```bash
 curl -sf http://localhost:8080/api/v1/machine/capabilities | jq .
@@ -21,17 +21,17 @@ curl -sf http://localhost:8080/api/v1/machine/capabilities | jq .
 Expected:
 
 ```json
-{ "capabilities": ["cupWarmer", "integratedScale"] }
+{ "capabilities": ["cupWarmer", "integratedScale", "ledStrip", "stopAtWeight"] }
 ```
 
 Quick assertions:
 
 ```bash
 curl -sf http://localhost:8080/api/v1/machine/capabilities \
-  | jq -e '.capabilities | index("integratedScale") != null and index("cupWarmer") != null'
+  | jq -e '.capabilities | index("integratedScale") != null and index("stopAtWeight") != null'
 ```
 
-Exit 0 → both identifiers present.
+Exit 0 → both identifiers present. `stopAtWeight` advertises that the machine firmware runs its own stop-at-weight from the integrated scale — there is no dedicated SAW endpoint; the target rides `PUT /api/v1/workflow` (`context.targetYield`).
 
 ### 2. Scale snapshot stream is alive without an external scale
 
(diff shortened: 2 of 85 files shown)


tadelv:
Superseded by #612 (merged), which syncs the firmware wake schedule and inactivity timeout. Removed from this PR: the ScheduleStatus invention (no such MMR), the forced 60-minute floor (firmware 0 = disabled), and the arbitrary 30-minute wake window (240-min firmware-max fallback per product decision).

PR #468, opened 2026-07-16
feat(bengle): wire the firmware-owned scheduled cup-warmer pre-warm (contract v2)

ChampionDesigns:
> **Stacked PR — B-10 of 10.** Builds on #467 (`feat/bengle-wake-schedule`). Until that merges this PR's diff includes its commits; please review/merge in order **B-1 → B-10**.

The Bengle firmware can pre-warm the cup-warmer mat ahead of a scheduled wake: set `MatPreheatEnable`, give it a lead time, and it runs the mat from that many minutes before a wake window opens and holds it until the window closes. It does this with no tablet connected, and without waking the machine - the boilers stay cold, so you get warm cups without an hour of the group heating up for nothing. Nothing in the app ever wrote those registers, so the feature existed in the firmware and was unreachable from anywhere. This PR is pure register plumbing: the firmware owns all the timing, and the app writes two settings and reads one status flag. It builds directly on B-9, because a pre-warm is meaningless without a wake schedule to be early for.

## Summary

- **Problem:** Firmware register-table rows 59 to 61 - `MatPreheatEnable` (`0x008038D0`), `MatPreheatLeadMin` (`0x008038D4`) and the read-only `MatPreheatActive` (`0x008038D8`) - had no app-side consumer. The scheduled pre-warm could not be turned on.
- **Why it matters:** It is the feature that makes the wake schedule worth having on an espresso machine with a cup warmer, and it is free in energy terms compared to waking the machine early, since the boilers stay off.
- **What changed:** Three registers, one nullable REST field triple, and a status flag. `GET /api/v1/machine/cupWarmer` gains `prewarmEnabled`, `prewarmLeadMinutes` and `prewarmActive`; `PUT` accepts the pre-warm pair alongside or instead of the setpoint. `MatPreheatActive` answers the question "why did the cup warmer come on by itself?" - it is true when the **schedule** is driving the mat, as opposed to the user's manual toggle. The app never writes it. The contract file goes to `contract_version: 2`.
- **What did NOT change (scope boundary):** **The app implements no pre-warm scheduling of its own.** There is no timer, no lead-time arithmetic, no wake-window lookup in this PR - the firmware does all of it. The app writes two settings and reads a flag. No new endpoint path; no WebSocket topic.

## Change Type (select all)

- [ ] Bug fix
- [x] Feature
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [x] REST API / handlers
- [ ] WebSocket API
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

- Closes #
- Related # (Bengle stack B-1 through B-10; depends on B-9 for the wake schedule this pre-warms against)

## Root Cause (if bug fix)

N/A - new capability.

N/A as a regression plan. Coverage of the new code:

- **`test/models/device/bengle_cup_warmer_test.dart`** (+287 lines): byte-exact writes to all three registers; the lead-minutes clamp at both ends; a negative case asserting the app **never** writes `MatPreheatActive`; the absent-register degradation path; the no-retry-storm latch, **including the concurrent-readers case**; and the re-probe on reconnect.
- **`test/services/webserver/de1handler_cup_warmer_test.dart`** (+181 lines): the REST shape, a `400` on an empty body, and `prewarmActive` being ignored in a request body.
- **`test/unit/models/device/impl/bengle/mmr_contract_test.dart`**: rows 59 to 61 plus the `contract_version` bump.
- **`test/models/device/mock_bengle_test.dart`** (+55 lines).

- [x] API spec updated: `assets/api/rest_v1.yml` - the three new `CupWarmerState` fields, the `PUT` semantics, the `400`, and the explicit statement that a pre-warm additionally needs a mat setpoint above zero **and** at least one wake window to do anything.
- [ ] N/A - no docs affected

`assets/api/bengle_hw_v1.yml` goes to `contract_version: 2` and the additive pin is documented in the contract file header (below).

## Security Impact (required)

- New or changed REST endpoints? `No` new path - `GET`/`PUT /api/v1/machine/cupWarmer` gain fields.
- BLE/USB surface changed? `Yes` - three Bengle-only MMR registers. Two writes, one read-only.

The registers drive a heating element, so: lead minutes are clamped to 0 to 120 **app-side, before the write**. The firmware clamps too, but a firmware-rejected MMR write is a **silent no-op**, not an error, so the app must not depend on it - a rejected write would leave the previous value in place while the app believed it had changed it. The REST layer additionally rejects an out-of-range `prewarmLeadMinutes` with a `400` rather than clamping it, so a client learns it asked for something impossible.

## User-Visible Changes

A Bengle can be told to warm its cup mat ahead of a scheduled wake, and the app can show whether the mat is currently being driven by the schedule rather than by hand. `GET /api/v1/machine/cupWarmer` gains three keys, all of which can be `null`. `PUT` now accepts a body carrying only the pre-warm pair (previously `temperature` was mandatory), and rejects a body carrying none of the three writable fields with a `400`.

- [x] `flutter analyze` - clean (`No issues found!`, 10.5s)
- [x] `flutter test` - **2331 tests pass** on this branch at `274fce9a` (B-9 was 2295, so this branch adds 36).

The number is recorded: the tip gate run shows `07:02 +2331: All tests passed!` on this branch at `274fce9a`, including the integration suite (hot-water stop-at-weight, steam-record persistence). An earlier gate run iterated only B-2 through B-9 and left this tip ungated; that gap has since been closed by a dedicated run on the tip.

- Simulated devices? (`simulate=1`): `Yes`. `GET /api/v1/machine/cupWarmer` returns `{"temperature":0.0,"currentTemperature":null,"prewarmEnabled":false,"prewarmLeadMinutes":30,"prewarmActive":false}`; `PUT {"prewarmEnabled":true,"prewarmLeadMinutes":45}` returns `200` with `{"status":"accepted","prewarmEnabled":true,"prewarmLeadMinutes":45}` and the subsequent `GET` echoes it.
- What you personally verified and how: the bytes on the wire, through `FakeBleTransport`, plus the REST shape against `MockBengle`.
- Edge cases checked: firmware **without** rows 59 to 61 (older firmware is exactly this) - reads report `null`, writes land in unmapped space and are inert; concurrent `GET`s while the first read is still walking the MMR timeout ladder; a reconnect to firmware that **does** have the rows.
- What you did **not** verify: **the entire feature, on hardware.** No mat has been pre-warmed. Worse than the usual "untested on hardware" caveat: **the firmware that has these registers has never been flashed to the bench machine.** The flashed build (90) does not carry rows 59 to 61, which is why the degradation path is so carefully built and so well tested - it is the only path that has actually been exercised. The happy path is asserted against a `FakeBleTransport` that answers because I told it to.

### Evidence

- [x] Test output (2331 pass)
- [ ] Log snippets
- [x] curl output (the cup-warmer `GET`/`PUT` round-trip against `simulate=1`)

## Compatibility & Migration

- Backward compatible? `Yes` on the API - `GET` adds keys and `PUT` **relaxes** its requirement (a body with only pre-warm fields is now valid where previously `temperature` was mandatory). No path or status-code change on any previously-valid request.
- **Contract migration: yes, and it is the first one.** `contract_version` goes 1 to 2. The contract file and the checker's `_expectedContractVersion` constant move **in the same commit**, and the check is strict equality, so they can never be split.

- **The contract file gets an `additive pin`, and the base pin does not move.** This is the unusual part and it is worth understanding before you approve it. Rows 0 to 58 stay pinned to the hardware-validated firmware build. Rows 59 to 61 are pinned **separately**, to the newer firmware revision that introduces them. The reason the whole file is not re-pinned to that revision is that its register table is not a superset. It has **no rows 39 to 49** (the scale-calibration registers, `ScaleTare`, the six LED rows, `TargetMilkTemp`) and **no row 58** (`MatCurrentTemp`) - every one of which the app consumes today, and the validated build provides. Regenerating from it would delete those rows and break the contract checker. So the file is pinned in two places. It is documented in the contract file header, and I verified it against both firmware revisions, but a two-source contract file is a smell and you should decide whether you can live with it.
- **The "firmware does not have these registers" case is latched per connection, and concurrent readers share one in-flight read.** Both are needed, and the second is the non-obvious one. On firmware without the rows, an MMR read does not fail fast - the firmware **accepts** the read request and then simply never answers, so it fails only when the read-timeout ladder gives up, seconds later. The REST surface is polled far faster than that (a cup-warmer page every few seconds, a header every minute), so a second caller arrives **while the first read is still open**, at which point the latch is necessarily still false and the second caller opens its own timeout ladder. That is precisely the retry storm the latch exists to prevent. Sharing the in-flight future bounds it to one ladder per connection, full stop; the latch then absorbs every later poll. The latch is dropped on reconnect so a machine reflashed with pre-warm firmware is picked up in the same app session.
- **Neither setting is re-asserted on connect, and that is deliberate** - it is the opposite of what B-7 does with `CupWarmerMode`. `MatPreheatEnable` and `MatPreheatLeadMin` are `PERM_RWD`, i.e. persisted in firmware; `CupWarmerMode` is RAM-only and has to be re-pushed. Getting this backwards in either direction would be a bug, so it is worth a second pair of eyes.
- **A partial `PUT` reads the other half from the machine first.** The two pre-warm registers are written as a pair, so a request carrying only `prewarmEnabled` needs a lead time from somewhere. It comes from the machine's current persisted state, falling back to the firmware default of 30. That means a `PUT` can perform a read, which is slightly surprising for a write endpoint.

## Risks & Mitigations

- **Risk: the firmware pin (see the note at the top of this page).** The contract pins the pre-warm registers to a firmware dev commit rather than a released build, and CI does not check the pin.
  - **Mitigation:** Not a code issue. The pinned commit exists in the firmware lineage, and the registers exist in the current firmware build. It is a review-time coordination point for the whole Bengle contract, and the reference is intentionally provisional while the firmware is still in active development - `firmware.commit` is finalized to a released or tagged firmware reference once the firmware is cut as a release, not tied to an interim dev build before then - not a defect specific to this PR.
- **Risk:** A user enables the pre-warm with no wake schedule, or with the mat setpoint at zero, and nothing happens - silently. The firmware needs all three: an enabled pre-warm, a setpoint above zero, and at least one wake window. And the wake table and the wall clock are RAM-only, so a machine that has been power-cycled since the app last connected has no schedule to be early for either (B-9 re-pushes on connect, which is what makes this work at all).
  - **Mitigation:** `rest_v1.yml` says so explicitly. That is documentation, not a guardrail: **the API will happily accept a pre-warm that cannot fire.** A UI should not offer the toggle without a schedule, and that is a Streamline concern rather than one this PR can solve.
- **Risk:** On firmware without the registers, a `PUT` writes into unmapped space and the write is silently inert - so the app could report a success that did not happen.
  - **Mitigation:** It does not. After a pre-warm write the handler **re-reads** and echoes what the machine actually reports, so on unsupported firmware the response carries `null` rather than the values the caller sent. The app never fabricates a value here - and in particular it never reports `prewarmActive: false` when the truth is that it cannot tell.


tadelv:
Superseded by #612 (merged), which wires the firmware-owned scheduled pre-warm. Removed from this PR: contract v2, the additive pin, the two-firmware-source contract model, and assets/api/bengle_hw_v1.yml.

PR #478, opened 2026-07-19
fix(ble): liveness probe recovers a silently dead scale-watch scan

giladger:
## Summary

- Problem: field reports of the preferred scale sometimes never auto-connecting until a manual tap, with the log showing `Background device watch started` and then silence. The watch's native scan can die without anything reaching Dart.
- Why it matters: the background scale watch (#447) is the only auto-reconnect mechanism on Android; when its scan dies silently, scale reacquisition is dead for up to 25 minutes (the opportunistic-downgrade refresh) or until a manual connect.
- What changed: while the watch believes its scan is running, it probes `UniversalBle.isScanning()` every 90s (host-side bookkeeping check — zero radio cost). A dead scan is torn down and restarted through the existing `_restartWatchOrReportFailure` path, so a restart that also fails feeds the legacy-backoff fallback instead of dying silently. Probe errors count as alive — an unprovable probe must not churn the session. `ScaleWatch`'s sighting log moves to INFO so field logs distinguish armed-but-no-sighting from sighted-but-connect-failed.
- What did NOT change (scope boundary): burst scans, watch arming policy, ConnectionManager, and the fork itself are untouched. No file overlap with #476 — complementary (#476 reduces the scan churn that triggers one of the death mechanisms; this detects and recovers when it happens anyway).

## Root cause (in the fork, observed from this repo)

- `SafeScanner`'s scan-frequency throttle (>5 starts/30s) swallows a start entirely: the Dart call returns success while the real start is deferred — and a later `stopScan` cancels the deferral (`handler.removeCallbacksAndMessages(null)`). Busy periods strand `_watchScanActive = true` with no native scan and none coming.
- `onScanFailed` is logged Kotlin-side and never surfaced to Dart.

Coverage note (honest limits): `isScanning()` reflects SafeScanner's host-side bookkeeping, so the probe reliably catches the throttle-swallow case (bookkeeping never went true). A late `onScanFailed` does not clear that bookkeeping today, so that flavor stays bounded by the 25-min refresh until the fork surfaces scan failures — candidate fork follow-up: update SafeScanner state in `onScanFailed` and/or emit a scan-state event.

## Change Type (select all)

- [x] Bug fix
- [ ] Feature
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [ ] REST API / handlers
- [ ] WebSocket API
- [x] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [ ] Docs / specs

## Linked Issues

- Related #447 (background scale watch), #476 (scan-session hardening — complementary, no file overlap)

## Root Cause (if bug fix)

- Root cause: see fork mechanisms above — the watch had no way to observe a native scan that stopped existing after a successful-looking start.
- Missing detection or guardrail: no liveness verification between the 25-minute refreshes.
- Contributing context: scan start/stop churn (machine reconnects, bursts, watch pause/resume) is what trips the SafeScanner throttle; #476 reduces that churn.

  - [x] Unit test
- Target test or file: `test/services/universal_ble_discovery_service_test.dart` — `resilience` group: "the liveness probe restarts a silently dead native scan", "a live scan passes the liveness probe untouched". The fake platform now mirrors native scanner state (`nativeScanning`), so tests simulate silent death by flipping it without a stop call.
- Scenario the test should lock in: watch active + native scan gone → fresh filtered start within one probe interval; healthy scan → zero session churn across many probe intervals.
- If no new test added, why not: N/A

- [ ] API spec updated
- [ ] API docs updated
- [ ] Plugin docs updated
- [ ] Skin docs updated
- [ ] Profile docs updated
- [x] Device docs updated
- [ ] N/A — internal watch-lifecycle hardening; DeviceManagement.md's watch section already describes self-healing at this level of detail

## Security Impact (required)

- BLE/USB surface changed? Yes
- If any `Yes`, explain risk and mitigation: adds a 90s host-side `isScanning()` poll while the watch runs (queue command, no radio activity) and a scan restart when it reports dead. Worst case under a persistently failing stack is one restart attempt per 90s, which SafeScanner's own throttle bounds.

## User-Visible Changes

- Android: a preferred scale powered on after the watch's scan silently died now connects within ~90s + discovery time instead of up to 25 minutes / manual tap.
- One new INFO log line per watch-driven reconnect (`Preferred scale … sighted`).

## Compatibility & Migration

- Backward compatible? Yes

- [x] `flutter analyze` — clean (re-run after rebase onto merged #476)
- [x] `flutter test` — all pass (2402 passing, 0 failures on the rebased branch; discovery/watch + connection/wake/Bengle suites 185/185)

- OS / platform tested: unit-tested against the fake platform; tablet soak pending (needs a naturally-occurring silent death or an adb-forced one — the triggering condition is timing-dependent by nature)
- Simulated devices? No (watch is BLE/Android-scoped)
- Real hardware? Field report drove the diagnosis; the fix's on-device validation is the next scale-off/on cycle on the reporter's tablet
- What you did **not** verify: a real SafeScanner throttle event end-to-end on hardware

### Evidence

- [x] Test output — probe test fails RED without the probe (scan stays dead), passes with it
- [x] Log snippets — field log shows `Background device watch started (prefix: null)` with no subsequent sighting despite the scale advertising (manual tap connected it immediately)

## Risks & Mitigations

- Risk: probe restarts churn the scan session on false "dead" readings.
  - Mitigation: probe errors count as alive; restart only on an explicit `isScanning() == false`; state re-checked after the await so bursts/stops that took over are respected.
- Risk: restart loop under a persistently broken stack.
  - Mitigation: every restart routes through `_restartWatchOrReportFailure` — a failing restart reports watch-unavailable and activates the legacy backoff fallback.

### DIFF
--- a/lib/src/controllers/connection/scale_watch.dart
+++ b/lib/src/controllers/connection/scale_watch.dart
@@ -157,9 +157,6 @@ class ScaleWatch {
   Future<void> _onSighting(Scale scale, int gen) async {
     _connecting = true;
     try {
-      // INFO deliberately: one line per reconnect, and the key field
-      // breadcrumb for diagnosing "scale did not auto-connect" reports
-      // (armed-but-no-sighting vs sighted-but-connect-failed).
       _log.info(
         'Preferred scale ${scale.deviceId} sighted; '
         'stopping watch and connecting',
--- a/lib/src/controllers/connection/scale_watch.dart
+++ b/lib/src/controllers/connection/scale_watch.dart
@@ -157,7 +157,10 @@ class ScaleWatch {
   Future<void> _onSighting(Scale scale, int gen) async {
     _connecting = true;
     try {
-      _log.fine(
+      // INFO deliberately: one line per reconnect, and the key field
+      // breadcrumb for diagnosing "scale did not auto-connect" reports
+      // (armed-but-no-sighting vs sighted-but-connect-failed).
+      _log.info(
         'Preferred scale ${scale.deviceId} sighted; '
         'stopping watch and connecting',
       );
--- a/lib/src/services/universal_ble_discovery_service.dart
+++ b/lib/src/services/universal_ble_discovery_service.dart
@@ -95,13 +95,9 @@ class UniversalBleDiscoveryService extends BleDiscoveryService
   /// the watch scan before that kicks in.
   static const _watchRefreshInterval = Duration(minutes: 25);
 
-  /// Liveness probe cadence. The native scan can die without anything
-  /// reaching Dart: the fork drops `onScanFailed` (logs only), and its
-  /// SafeScanner throttle can swallow a start entirely (returns success,
-  /// defers the real start, and a later stopScan cancels the deferral).
-  /// `UniversalBle.isScanning()` is a cheap host-side check — no radio
-  /// use — so probe often enough that a dead watch recovers in ~1 probe
-  /// interval instead of waiting for the 25-min refresh.
+  /// Dead-scan probe cadence. The native scan can die without any event
+  /// reaching Dart; see doc/AI_BLE_NOTES.md ("Watch Scan Silent Death")
+  /// for the fork mechanics and the probe's coverage limits.
   static const _watchLivenessInterval = Duration(seconds: 90);
   Timer? _watchLivenessTimer;
 
@@ -301,11 +297,11 @@ class UniversalBleDiscoveryService extends BleDiscoveryService
       try {
         alive = await UniversalBle.isScanning();
       } catch (e, st) {
-        // A failed probe proves nothing — don't churn the scan over it.
+        // Fail open: an unprovable probe must not churn the session.
         log.fine('Watch liveness probe failed', e, st);
         alive = true;
       }
-      // Re-check: a burst/stop may have taken over during the await.
+      // A burst/stop may have taken ownership during the await.
       if (!_watchScanActive || _isScanning) return;
       if (alive) {
         _armWatchLiveness();
--- a/lib/src/services/universal_ble_discovery_service.dart
+++ b/lib/src/services/universal_ble_discovery_service.dart
@@ -95,6 +95,16 @@ class UniversalBleDiscoveryService extends BleDiscoveryService
   /// the watch scan before that kicks in.
   static const _watchRefreshInterval = Duration(minutes: 25);
 
+  /// Liveness probe cadence. The native scan can die without anything
+  /// reaching Dart: the fork drops `onScanFailed` (logs only), and its
+  /// SafeScanner throttle can swallow a start entirely (returns success,
+  /// defers the real start, and a later stopScan cancels the deferral).
+  /// `UniversalBle.isScanning()` is a cheap host-side check — no radio
+  /// use — so probe often enough that a dead watch recovers in ~1 probe
+  /// interval instead of waiting for the 25-min refresh.
+  static const _watchLivenessInterval = Duration(seconds: 90);
+  Timer? _watchLivenessTimer;
+
   final StreamController<void> _watchFailureController =
       StreamController.broadcast();
 
@@ -157,6 +167,8 @@ class UniversalBleDiscoveryService extends BleDiscoveryService
   }) async {
     _watchRefreshTimer?.cancel();
     _watchRefreshTimer = null;
+    _watchLivenessTimer?.cancel();
+    _watchLivenessTimer = null;
     _cancelWatchScanSub();
     _watchScanActive = false;
     if (stopOsScan) {
@@ -276,7 +288,35 @@ class UniversalBleDiscoveryService extends BleDiscoveryService
     }
     _watchScanActive = true;
     _armWatchRefresh();
+    _armWatchLiveness();
     log.info('Background device watch started (prefix: $namePrefix)');
+  }
+
+  void _armWatchLiveness() {
+    _watchLivenessTimer?.cancel();
+    _watchLivenessTimer = Timer(_watchLivenessInterval, () async {
+      _watchLivenessTimer = null;
+      if (!_watchScanActive || _isScanning) return;
+      bool alive;
+      try {
+        alive = await UniversalBle.isScanning();
+      } catch (e, st) {
+        // A failed probe proves nothing — don't churn the scan over it.
+        log.fine('Watch liveness probe failed', e, st);
+        alive = true;
+      }
+      // Re-check: a burst/stop may have taken over during the await.
+      if (!_watchScanActive || _isScanning) return;
+      if (alive) {
+        _armWatchLiveness();
+        return;
+      }
+      log.warning(
+        'Watch scan died silently (isScanning=false); restarting',
+      );
+      await _deactivateWatchScan(stopOsScan: false, context: 'liveness');
+      await _restartWatchOrReportFailure('liveness restart');
+    });
   }
 
   void _armWatchRefresh() {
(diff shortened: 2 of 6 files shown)


lib/src/services/universal_ble_discovery_service.dart
```diff
@@ -244,9 +256,37 @@ class UniversalBleDiscoveryService extends BleDiscoveryService
...
   }
 
+  void _armWatchLiveness() {
+    _watchLivenessTimer?.cancel();
+    _watchLivenessTimer = Timer(_watchLivenessInterval, () async {
+      _watchLivenessTimer = null;
+      if (!_watchScanActive || _isScanning) return;
+      bool alive;
+      try {
+        alive = await UniversalBle.isScanning();
```
tadelv, line 298: This adds another direct `universal_ble` operation and more `SafeScanner`-specific knowledge outside `lib/src/services/ble/`, despite the hard boundary in `AGENTS.md` and `doc/AI_BLE_NOTES.md`. Please move this concrete scanner behavior behind the BLE boundary, add a BLE-layer abstraction for scan liveness, or explicitly document an architectural exception. Keeping the liveness policy here is preferable to moving it into `ConnectionManager`; the issue is the package boundary.

lib/src/controllers/connection/scale_watch.dart
```diff
@@ -157,7 +157,10 @@ class ScaleWatch {
   Future<void> _onSighting(Scale scale, int gen) async {
     _connecting = true;
     try {
-      _log.fine(
+      // INFO deliberately: one line per reconnect, and the key field
```
tadelv: Please move this diagnostic rationale into `doc/AI_BLE_NOTES.md` and keep the production comment minimal or remove it. `AGENTS.md` says not to add explanatory comments to new code and directs debugging history/hardware constraints to the matching `doc/AI_*_NOTES.md` file.

test/services/universal_ble_discovery_service_test.dart
```diff
@@ -445,6 +452,57 @@ void main() {
...
+        );
+        zoned.initialize();
+        async.flushMicrotasks();
+        zoned.startDeviceWatch(_watchFilter);
+        async.flushMicrotasks();
+
+        async.elapse(const Duration(minutes: 10));
+        async.flushMicrotasks();
+
+        expect(platform.startScanCalls, hasLength(1),
```
tadelv, line 611: Please also add suite-level teardown for the regular `service`, not only cleanup inside these two new tests. The new 90-second timer can retain a service from an earlier test and later call through the globally replaced `UniversalBle` platform fake, creating order-dependent/full-suite flakes.

tadelv (dismissed):
Thanks for the focused fix. The liveness mechanism itself looks reasonable: it reuses the existing teardown/restart/fallback path, treats probe failures conservatively, and re-checks ownership after the await.

I’m requesting changes for the following repository-guideline and structural issues:

1. **Resolve the BLE package boundary.** `AGENTS.md` and `doc/AI_BLE_NOTES.md` require third-party BLE libraries to stay under `lib/src/services/ble/`. This PR adds another `UniversalBle`-specific operation and native-fork implementation knowledge to `lib/src/services/universal_ble_discovery_service.dart`. Please either move the concrete scanner implementation behind the BLE boundary, introduce a BLE-layer scan-session abstraction, or explicitly document an architectural exception. The liveness policy should remain in the scanner lifecycle rather than moving into `ConnectionManager`.

2. **Update the required documentation and trim production-code rationale.** `doc/DeviceManagement.md` currently documents the 25-minute refresh and adapter recovery, but not the new 90-second dead-scan probe, fail-open probe behavior, failed-restart fallback, or the remaining `onScanFailed` limitation. Please update that document and add the fork/debugging lesson to `doc/AI_BLE_NOTES.md` (and preferably the archived background-watch design). `AGENTS.md` directs hardware constraints/debugging history to `doc/AI_*_NOTES.md` and asks new code to avoid explanatory comments.

3. **Make timer cleanup deterministic in tests.** The new 90-second timer can keep old service instances alive across tests and later touch the globally replaced `UniversalBle` fake. Please add teardown that stops active watches/timers for the regular test service and ensure every `fakeAsync` service is cleaned up before its test exits.

4. **Rebase onto current `main` and rerun verification.** This branch is behind the now-merged #476 changes. After rebasing, please rerun `flutter analyze`, the discovery/watch suites, relevant connection tests, and full `flutter test`, and update the evidence in the PR description.

Real Android validation of the fork throttle path remains desirable, but I would not block the code solely on reproducing that timing-dependent native failure once the structural, documentation, cleanup, and rebase items are addressed.

tadelv (changes requested):
Corrected review after re-checking the BLE seam: `lib/src/services/universal_ble_discovery_service.dart` is the permitted boundary between `universal_ble` and the application, so the direct `UniversalBle.isScanning()` usage is structurally appropriate. I have dismissed the earlier review and resolved that inline thread.

The remaining requested changes are:

1. **Update documentation and move explanatory rationale out of production code.** Document the 90-second dead-scan probe, fail-open behavior when `isScanning()` throws, restart/fallback behavior, and the remaining native `onScanFailed` limitation in `doc/DeviceManagement.md` and `doc/AI_BLE_NOTES.md`. Remove or substantially trim the explanatory comments added in production code in accordance with `AGENTS.md`.

2. **Make timer cleanup deterministic in tests.** Add suite-level teardown for the regular discovery service so active watch/liveness timers cannot survive into later tests and interact with a replaced global `UniversalBle` fake. Keep explicit cleanup for the `fakeAsync` service instances as well.

3. **Rebase onto current `main` and rerun verification.** After rebasing over the merged #476 changes, rerun `flutter analyze`, the discovery/watch and relevant connection suites, and full `flutter test`, then update the PR’s verification evidence.

The placement of the liveness logic in the discovery-service seam is correct, and the implementation’s reuse of the existing teardown/restart/fallback path still looks sound.

giladger:
All three requested changes are in. The branch is rebased onto current main (over the merged #476 — clean, no conflicts; #476 preserved the watch seams, and the rebase picked up the `DeviceWatchFilter` slimming) with the review fixes as a separate commit (1166bf7) on top of the rebased probe commit (ffff863):

1. **Documentation + comment trim** — `doc/DeviceManagement.md`'s watch section now documents the 90s dead-scan probe, fail-open probe errors, the restart-failure → legacy-backoff fallback, and the `onScanFailed` coverage limit. `doc/AI_BLE_NOTES.md` gains footgun #4 ("Watch Scan Silent Death") with the fork mechanics, the `isScanning`-is-host-bookkeeping caveat, the candidate fork follow-up, and field-triage log breadcrumbs; the archived background-watch design doc records the deviation. Production comments are trimmed to constraints, with the interval constant pointing at the footgun entry.

2. **Deterministic test cleanup** — suite-level `tearDown` stops the shared service's watch so refresh/liveness timers cannot survive into later tests against a replaced `UniversalBle` fake, and every `fakeAsync` service instance (including the failed-refresh one that previously skipped it) stops its watch before its test exits.

3. **Rebase + verification rerun** — on the rebased branch: `flutter analyze` clean; discovery/watch + connection/wake/Bengle suites 185/185; full `flutter test` **2402 passing, 0 failures**. PR description's verification section updated to match.

Agreed on the real-Android throttle validation being desirable — the reporter's tablet is running this build next, and footgun #4's triage breadcrumbs (`Background device watch started` / `Preferred scale … sighted` at INFO) are there to make the field evidence unambiguous either way.

tadelv (approved):

tadelv:
thanks!

Issue #385, opened 2026-06-25
[Feature Request] Migrate missing default profiles from prior-TCL based app


dhcabinian:
Specifically, the Weber Spring Clean profile is no longer available in assets/defaultProfiles (https://raw.githubusercontent.com/decentespresso/de1app/refs/heads/main/de1plus/profiles/weber_spring_clean.tcl):

```
[...]

PR #501, opened 2026-07-25
fix(profiles): derive legacy de1app frames instead of copying stale ones

skialpine:
I have spent time this weekend to make sure that the de1app, Decanza and reaprime all make the same coffee from the same named profile. I have found and fixed a handful of issues in Decenza, and so thought I would bring that knowledge to reaprime.  I believe with this change all 3 apps profiles make the same coffee from the same name profile. In addition this weekend I have changed Decenza to be able to injest reaprime profiles. While we were both using json they uses different string and number encodings. Now they should be the same and profiles can be copied back and forth.

## Summary

- **Problem:** `tools/ingest_profiles.py` copies de1app's stored `advanced_shot` array for `settings_2a`/`2b` profiles. de1app's `save_profile` writes that array out of the *global* `::settings`, so a simple profile ships whatever frames were loaded when it was last saved — and de1app never reads them back for those types, it regenerates from the scalars at load time. The existing guard fired only when the array was **empty**, which is the one case where copying is harmless.
- **Why it matters:** `Default1.json` runs its last four frames at **75 °C and 54 °C** against a declared `espresso_temperature` of 90.0. `Classic Italian espresso` stops at **60 g** where de1app stops at 36 g. `Preinfuse then 45ml of water` and `Gentle and sweet` had no stop target at all. Four A-Flow profiles were frozen at 6 frames against the plugin's 9, and could never self-correct.
- **What changed:** Ported de1app's `pressure_to_advanced_list` / `flow_to_advanced_list`; frames for legacy types are now derived and the stored array is ignored outright. Stop targets resolve by profile type the way de1app's own stop switches do. A-Flow is harvested from the plugin directory. Beverage types `tea` / `tea_portafilter` / `filter` are carried through instead of being flattened onto `pourover`. 33 corpus files rewritten (11 content, 15 beverage-type, 6 float-noise normalisations, plus the manifest), one added, provenance recorded per profile.
- **What did NOT change (scope boundary):** The runtime import path is untouched — `de1app_importer.dart` scans `profiles_v2` JSON and `tcl_parser.dart` never reads `advanced_shot`, so this defect never reached it. No behaviour is keyed on the new beverage types. The rebuild is a deliberate allowlist, not a full re-ingest: re-ingesting all 63 attributable sources reproduces 49 byte-identically and rewrites 14, and those 14 would *lose* information — five carry author attributions applied during the earlier curation pass (`Damian's *`, `Cremina`) that de1app does not have, and ten carry `lang`, `reference_file`, `legacy_profile_type` and `changes_since_last_espresso`, none of which the tool emits.

### Result against Decenza

Decenza derives these frames through de1app's own builders, and its corpus is 89/89 in sync with de1app. Comparing the two bundled corpora by machine equivalence (absent/`""`/`0` unified, the axis a frame's pump does not drive ignored, a zero-value limiter treated as absent):

| | machine-equivalent | differ | reaprime-only |
|---|---|---|---|
| before | 33 | 29 | 8 |
| after | **63** | **0** | 8 |

**Every profile both apps ship is now machine-equivalent.**

The last four took one extra step. After the frame work they still differed by float-representation noise carried in de1app's own `.tcl` files — `9.999999999999993` against `10.0`. That was never a behaviour difference (the DE1 takes a frame's driven axis as one byte, `(0.5 + value * scale).toInt()` at `unified_de1.profile.dart:57`, scale 16 on DE1 and 10 on Bengle — every differing pair encoded identically at both), but it was permanent diff noise between two corpora that ought to be comparable. `convert_step` now snaps to six decimals on serialise so a later harvest cannot reintroduce it, and six corpus files carrying old noise were normalised in place.

This is deliberately **not** Decenza's own format — it writes every number to two decimals (`"36.00"`), and adopting that would change the serialised form of all 71 profiles and therefore every profile id, for no behavioural gain.

Verified across every file this PR touches: **no DE1 byte moves except in the eleven profiles whose content is deliberately corrected.** The 15 beverage-type changes and the 6 float normalisations move nothing.

The 8 reaprime-only profiles are deliberate additions with no de1app counterpart: the four `Baseline` variants, `D-Flow / default`, `I Can't Believe It's Not Filter`, `PSPH` and `Soup 58`.

## Change record (OpenSpec)

This change was planned and executed through OpenSpec, and the record ships with it.
The capability is now a spec rather than only a commit message:

- **`openspec/specs/default-profile-ingest/spec.md`** — six requirements covering how
  legacy frames are derived, how stop targets resolve by profile type, that derived
  frames must agree with their own scalars, that beverage type is preserved rather than
  flattened, that a profile in two source directories is an error, and that the corpus
  records where each profile came from. This is the repo's **first** entry under
  `openspec/specs/`, so the PR also establishes that directory.
- **`openspec/changes/archive/2026-07-25-fix-legacy-profile-ingest/`** — the planning
  trail: `proposal.md`, `design.md` (the six decisions D1–D6), `tasks.md` (35/35), plus
  two artifacts produced while working:
  - `baseline.md` — the controlled experiments the proposal rests on, the de1app and
    A_Flow revisions used, and the three independent validations of the port.
  - `migration-impact.md` — what the profile-hash change does and does not invalidate,
    and why design open question 1 needed no decision in the end.

Reviewers who want the *why* should read `design.md`; the evidence for every factual
claim in this PR is in `baseline.md`.

## Change Type (select all)

- [x] Bug fix
- [ ] Feature
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [x] REST API / handlers
- [ ] WebSocket API
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

- Closes #
- Related #385 — adds `A-Flow / default-light`, a de1app default the corpus never carried. (Does not close it; `Cleaning/Weber Spring Clean` and other missing profiles are still absent.)
- Upstream: decentespresso/de1app#350 — the A-Flow shadowing. This PR is correct however that resolves; see the provenance guard below.

## Root Cause (if bug fix)

- **Root cause:** de1app's `save_profile` serialises `advanced_shot` from the global `::settings` rather than from the profile being saved, so a `settings_2a`/`2b` `.tcl` on disk routinely carries frames belonging to a different profile. de1app tolerates this because `sync_from_legacy` regenerates from the scalars for those types and never reads the array back. The ingest tool had no equivalent regeneration, so it took the array at face value. The same mistake was made a second time on the stop targets: the tool read `final_desired_shot_weight_advanced` unconditionally, where de1app reads the plain spelling for anything that is not `settings_2c`.
- **Missing detection or guardrail:** The guard at `ingest_profiles.py:113` was conditioned on `not steps and settings_type in (...)`. The `not steps` clause is the defect — it makes the check fire only when the array is empty, and never when it is populated with the wrong content. Nothing downstream compared a bundled profile's frames against its own scalars.
- **Contributing context:** `doc/plans/archive/default-profiles-curation/audit.md` had the mechanism in its own words ("de1app's `advanced_shot` is garbage") but scoped it to "contained to the 3 levers" on the strength of a scan for one known step *pattern*. That scan cannot see a profile whose stale frames are merely the wrong temperature. Five more profiles were in the same class. That note is corrected in this PR.

  - [x] Unit test
- **Target test or file:** `tools/ingest_profiles_test.py` (new, 19 tests) and `test/profiles/default_profiles_bundled_test.dart`.
- **Scenario the test should lock in:**
  - `LegacyStepSynthesisTest` feeds a `settings_2a` fixture whose stored `advanced_shot` holds two pour-over frames at 75 °C / 54 °C against a declared 90 °C, and asserts none of them reaches the output — the exact shape that produced `Default1.json`.
  - `SourceCollisionTest` asserts two disagreeing source directories fail naming both paths, that identical copies are *not* an error, and that a collision writes no output file.
  - `default_profiles_bundled_test.dart` adds three assertions over the 12 `pressure`/`flow` typed bundled profiles: frame names must come from the generator's vocabulary, no two frames may differ by more than 10 °C, and each frame's pump must match the profile type. The shipped `Default1.json` failed all three.

- [x] API spec updated: `assets/api/rest_v1.yml` — both `beverage_type` / `beverageType` enums widened
- [x] API docs updated: `doc/Api.md` — wire spellings noted on the Profiles section
- [x] Profile docs updated: `doc/Profiles.md` — manifest `provenance`, beverage-type spellings
- [ ] N/A — no docs affected

Also: `doc/AI_STORAGE_NOTES.md` gains the ingest-tool rationale (why frames are derived, which de1app quirks are reproduced deliberately, why A-Flow comes from the plugin directory), and the stale scoping claim in `doc/plans/archive/default-profiles-curation/audit.md` is marked superseded.

## Security Impact (required)

- New or changed REST endpoints? **No** — existing endpoints, wider enum value set only.
- New or changed network calls? **No**
- File system access changed? **No** — the ingest tool is a developer tool, not shipped in the app.
- Plugin sandbox boundary changed? **No**

## User-Visible Changes

- **Eleven bundled profiles now brew differently, because they were wrong.** `Default` no longer drops to 75 °C / 54 °C. `Classic Italian espresso` stops at 36 g instead of 60 g. `7g basket` stops at 35 g instead of 36 g. `Preinfuse then 45ml of water` and `Gentle and sweet` gain the stop targets they were missing. Four A-Flow profiles go from 6 frames to 9. `Cleaning/Forward Flush x5` regains its lost `Pressure rise 1 start` frame.
- **`Advanced spring lever` is now de1app's profile** (Decent, 5 frames at 88 °C) rather than the John Weiss variant that had been occupying that name. The Weiss profile is already shipped correctly and separately as `Weiss advanced spring lever`, unchanged by this PR. If the displaced variant is wanted it can be re-added under a distinguishing title — additive, and no reason to hold this.
- **`A-Flow / default-light` is new** — the corpus never carried it.
- **Fifteen profiles report their real beverage type**: thirteen tea profiles and two filter profiles were labelled `pourover`. Display only; no shot behaviour is keyed on it.
- **Existing installs:** corrected profiles arrive through the mechanism already in place — the new content is stored under its new content hash and the superseded record is *hidden*, never deleted. A user's own derivative of a bundled profile is untouched, because a bundled default cannot be modified in place (`ProfileController.update` rejects it), so any variant is already a separate `isDefault: false` record. **One side effect:** a user who *hid* one of these 27 profiles will see the corrected version reappear, since the replacement is created with `Visibility.visible`. Not new behaviour, but reaching 27 profiles at once for the first time. Noted rather than fixed — carrying visibility forward changes seeding semantics for all bundled profiles and belongs in its own change.

- [x] `flutter analyze` — clean (Flutter 3.44.8)
- [x] `flutter test` — 2496 / 2498 pass
- [ ] `(cd packages/dye2-plugin && npm run build)` — not applicable, plugin untouched

`flutter analyze` reports one warning, `The asset directory 'assets/plugins/dye2.reaplugin/' doesn't exist`. It is **pre-existing** — I checked out `main` and reproduced it there, unchanged.

`flutter test` has two failures, both in `test/webui_support/webui_token_injection_test.dart` (`accepts another local interface address`, `accepts the cached WiFi address when enumeration fails`). Also **pre-existing** — the same two fail identically on `main`. They enumerate local network interfaces, so they are environment-dependent and untouched by this PR.

The changed test files pass in full: `flutter test test/profile_test.dart test/profiles/` → **63 / 63**.

- **OS / platform tested:** macOS (tooling and corpus only)
- **Simulated devices?** No
- **Real hardware?** No
- **What I personally verified and how:**
  - `flutter analyze` and `flutter test` under Flutter 3.44.8, with both pre-existing failures confirmed against `main`.
  - `cd tools && python3 -m unittest ingest_profiles_test` — 37 / 37 pass.
  - Reproduced the original finding before changing anything: the old tool reproduces `7g_basket.json`, `Classic_Italian_espresso.json` and `Preinfuse_then_45ml_of_water.json` **byte-identically** from de1app `fe5cf40`, and refuses `default.tcl` / `Gentle and sweet.tcl`, both of which have carried `advanced_shot {}` in every de1app commit back to 2020.
  - Cross-checked the port against Decenza's real `profile_sync` binary over all 89 de1app profiles. Disagreements: the four A-Flow source-path cases, four float-literal precision differences in de1app's files, and one profile absent from this corpus. **Zero attributable to the port.**
  - Third check, free: seven bundled profiles were already derived by de1app's own generator (they arrived via the Visualizer route). The port reproduces all seven **value-for-value** — same frames, names, pumps, transitions, temperatures, durations, targets, editor type.
  - The new Dart corpus assertions were run against the actual 71 corpus files in Python before being written as Dart: 12 derived profiles, zero failures.
- **Edge cases checked:** empty `advanced_shot` on a legacy type is now ingestible rather than refused; an underivable profile raises naming the missing field rather than falling back; identical copies in two source directories emit once without error; a collision writes no output file and exits 1.
- **What I did NOT verify:** no shot was pulled on real or simulated hardware; the corpus and encoder claims are static analysis plus the Decenza cross-check, not a brewed shot.

### Evidence

- [x] Test output

```
$ cd tools && python3 -m unittest ingest_profiles_test
Ran 37 tests in 0.109s
OK

$ flutter test test/profile_test.dart test/profiles/
00:00 +63: All tests passed!

$ flutter analyze
1 issue found.   # pre-existing missing-asset-dir warning, also on main

$ flutter test
2498 tests, 2 failures — both pre-existing on main (webui_token_injection_test)
```

Corpus against Decenza's materialised corpus:

```
before:  machine-equivalent 33   differ 29   reaprime-only 8
after:   machine-equivalent 63   differ  0   reaprime-only 8
```

Port against Decenza's `profile_sync` binary over all 89 de1app profiles: 84 agree, 4 disagree (the A-Flow source-path cases, fixed by harvesting from the plugin directory), 1 refuses to convert (`Test/profile_editor_demo` carries no `final_desired_shot_weight*` field at all; the tool now fails rather than emitting a profile with no weight stop — not a bundled profile here).

`Default1.json`, before and after:

```
before:  90.0  88.0  75.0  75.0  54.0  54.0
after:   90.0  88.0  88.0  88.0  88.0  88.0     (de1app declares espresso_temperature 90.0)
```

## Review findings addressed

A multi-agent review ran against this branch (code, tests, error handling, comments/doc accuracy, type design). One reviewer executed de1app's *actual* Tcl generators under `tclsh` against all 26 `settings_2a`/`2b` profiles and confirmed the port is faithful frame-for-frame. The findings that were real are fixed in `d19b7d2`..HEAD:

**Faithfulness to de1app**
- `espresso_temperature_steps_enabled` was compared as a string; de1app compares numerically, so `1.0` or `01` would have silently selected the single-temperature layout.
- Preinfusion scalars were read eagerly, rejecting a profile de1app derives fine when `preinfusion_time` is 0.

**Paths that could still emit a silently wrong profile**
- A v2 JSON export of a `settings_2a`/`2b` profile bypassed derivation entirely, carrying both the stale frames and the `_advanced` stop target — this PR's own bug, one file extension away. Now refused.
- `settings_2c` read its `_advanced` stop targets with a `0` default, so a profile missing them shipped with no stop. Now required; an explicit `0` still means "no stop".
- An unhandled `exit_type` dropped a frame's stop condition; an `advanced_shot` element that failed to parse dropped that frame; a zero-frame profile was written. All three now raise.
- `re.match` with `re.MULTILINE` only ever matched `advanced_shot` on line 1.
- A collision where one side failed to convert wrote the survivor — precedence by parseability, which D4 rejects.
- A failing run left the corpus half-rebuilt. Writes are now all-or-nothing.
- `source_provenance` recorded HEAD for untracked/modified files and an absolute local path outside a git checkout; `update_manifest` discarded hand-written `verified`/`note` annotations.
- One shipped provenance entry named `cleaning_forward_flush_x5.tcl`; the real file is `Cleaning_forward_flush_x5.tcl`, which resolved on APFS but not on Linux.

**Beverage type**
- `descale` mapped to `cleaning` in the tool but not in Dart, so a descale profile arriving over REST became `espresso` — and `cleaning` is the one value that changes machine conduct. `BeverageType.tryParse` now carries the alias and trims, and gives callers a way to distinguish "unknown" from "espresso".

**Docs that had drifted** — the re-ingest reproduction figures (49 reproduce byte-identically, 14 differ, and re-ingesting those 14 would *lose* author attributions and four keys the tool does not emit), the id-churn count (26 of 70, not 27 — `default-light` is an addition), the DE1 scale factor, the `""` quirk mechanism, and that a `de1plus`-root run cannot emit A-Flow by design.

**Test gaps the review found, now closed.** The corpus test asserted frame shape but never the stop targets, so the "60 g instead of 36 g" half of the bug could have returned with a green suite; it also keyed the derived set off each profile's own `type`, so a profile misclassified as `advanced` — the exact failure being guarded against — silently exempted itself from every check. Both fixed, plus: the 12 derived profiles are pinned by name; `empty` is no longer whitelisted in the frame vocabulary; an absolute 75–100 °C bound sits alongside the spread bound; every bundled profile's beverage type is asserted to round-trip. `BeverageType` invariants now iterate `values` rather than a hardcoded list. The Python suite went 19 → 37 tests, adding golden frame values (dropping the `hold_time -= 3` decrement now fails; it previously passed), the untested duration branches, the limiter that six shipped profiles carry, de1app's hold-time-gated flow decline, and the end-to-end write path.

## Compatibility & Migration

- **Backward compatible?** For the API, yes — `beverage_type` is additive, and unrecognised values still fall back to `espresso`. For brewing, deliberately **no**: eleven bundled profiles change what they send to the machine, which is the point of the PR.
- **Config / env changes needed?** No
- **Database migration needed?** No. 27 of 71 bundled profiles change content hash, but nothing outside the profile store references a profile id — no foreign key exists, and `Workflow` embeds the whole `Profile` by value, so historical shots keep the profile they were pulled with. The existing `_retireStaleDefaults` path handles the id change on next launch.

## Risks & Mitigations

- **Risk:** The synthesis port is a reimplementation and could be subtly wrong in the same direction for every profile.
  - **Mitigation:** Validated against two independent oracles — Decenza's real C++ generator across all 89 de1app profiles, and seven corpus profiles that de1app's own generator produced. Both agree.
- **Risk:** Someone dialled in on the current `Default` gets a different shot.
  - **Mitigation:** The honest framing is that 54 °C brew water is a defect, not a profile anyone chose. Users who built on it have their own `isDefault: false` copy, which is untouched.
- **Risk:** de1app #350 resolves by making `de1plus/profiles/` canonical, turning the harvest path stale.
  - **Mitigation:** This is exactly what the collision guard is for — the run fails and names every path rather than quietly harvesting the wrong one. The A_Flow author has proposed resolving #350 that way, which is why precedence ("plugin wins") was rejected.
- **Risk:** Float normalisation changes six more profile ids than the content fixes alone.
  - **Mitigation:** Verified inert — no DE1 byte moves in any of the six. The cost is that those six retire-and-readd on upgrade like the other 26.


tadelv (changes requested):
Thank you for the investigation and for identifying the incorrect bundled profiles. The underlying problem is valid, and the corrected profile data is valuable.

This PR cannot be merged in its current form. It introduces a permanent de1app ingestion framework, a provenance system, a parallel OpenSpec hierarchy, broad corpus policies, unrelated behavioural changes and a large test suite for functionality Reaprime does not intend to maintain.

Please reduce the PR to a one-time correction of the known-bad bundled profiles.

## Required scope

Keep the verified corrections to these five legacy profiles:

- `7g_basket.json`
- `Classic_Italian_espresso.json`
- `Default1.json`
- `Gentle_and_sweet1.json`
- `Preinfuse_then_45ml_of_water.json`

Keep the refresh of the four existing A-Flow profiles from the current canonical `Jan3kJ/A_Flow` source:

- `A-Flow____default-dark.json`
- `A-Flow____default-like-dflow.json`
- `A-Flow____default-medium.json`
- `A-Flow____default-very-dark.json`

Keep the addition of `A-Flow____default-light.json` as the fifth profile from the same authoritative source and revision.

Update `manifest.json` only as necessary to include the new A-Flow file.

## Do not maintain a de1app frame generator

Remove the Python ports of `pressure_to_advanced_list` and `flow_to_advanced_list`, together with all supporting frame-generation logic.

Reaprime is not adopting a permanent implementation of de1app’s legacy profile generators. We do not expect to ingest these profiles frequently enough to justify maintaining and testing a second implementation of de1app’s behaviour.

Replace this with a small safety guard in `tools/ingest_profiles.py`:

- reject every `settings_2a` and `settings_2b` TCL profile;
- reject them whether `advanced_shot` is empty or populated;
- explain in the error that the stored frames are not authoritative;
- direct the maintainer to generate final advanced JSON externally before ingestion.

Ordinary advanced profiles must remain ingestible.

Remove all machinery introduced for the permanent ingestion workflow, including:

- recursive discovery of de1app profile, plugin and editor directories;
- source collision detection;
- source selection policy;
- repository revision detection;
- subprocess-based Git inspection;
- generator-specific defaults and compatibility branches.

## Remove provenance

Remove provenance completely from:

- `assets/defaultProfiles/manifest.json`;
- `tools/ingest_profiles.py`;
- `doc/Profiles.md`;
- `doc/AI_STORAGE_NOTES.md`;
- all tests.

Provenance is not consumed by Reaprime and is not being adopted as a repository requirement.

For every profile retained in this PR, record the canonical upstream repository, source path and revision in the PR description. Any exceptional source-selection rationale may also be recorded concisely in the existing archived profile-curation documentation.

Do not add mandatory metadata or bookkeeping for every future bundled profile.

## Remove OpenSpec

Delete the entire new `openspec/` directory.

Reaprime already defines its planning and documentation structure in `AGENTS.md`. Durable rationale belongs in `doc/plans/archive/` and the relevant `doc/AI_*_NOTES.md` file. Do not introduce a parallel proposal, specification and change-archive system.

Condense only the necessary findings into:

- a short update to `doc/AI_STORAGE_NOTES.md`; and
- the existing `doc/plans/archive/default-profiles-curation/audit.md`.

Do not move the full OpenSpec material into another directory.

## Remove unrelated changes

Remove the changes to:

- `Cleaning__Forward_Flush_x5_2.json`
- `advanced_spring_lever.json`

These have different root causes and change machine behaviour. Submit them separately with their own evidence and review scope.

Remove all beverage-type work from this PR, including:

- changes to tea and filter profile JSON;
- new `BeverageType` values;
- `wireName` parsing and serialisation;
- profile hash changes;
- `ProfileController` changes;
- REST API schema changes;
- API and profile documentation for the new values;
- beverage parsing, alias, round-trip and hashing tests.

That is a separate change and must be reviewed separately.

Remove all float-only normalisation changes. They do not change machine behaviour and create unnecessary profile-ID churn and review noise.

## Reduce the tests

Delete the generator test suite and all new profile-specific regression tests.

The bundled profile JSON files are curated data, not logic that Reaprime needs to reimplement or independently derive. Do not add tests that pin exact profile frames, temperatures, names, counts, stop targets, temperature ranges or other profile contents. Those tests duplicate the data and create unnecessary maintenance whenever a profile is intentionally updated.

The correctness of the changed profile files must instead be demonstrated in the PR description with:

- the canonical upstream source file and revision;
- a concise before-and-after comparison;
- the command or process used to produce the corrected JSON;
- evidence that the resulting machine-relevant values match the authoritative source.

Keep only focused tests for behaviour that remains in `tools/ingest_profiles.py`:

- `settings_2a` and `settings_2b` TCL profiles are rejected even when they contain a populated `advanced_shot`;
- ordinary advanced profiles remain ingestible.

Do not add tests for provenance, source scanning, source collisions, beverage-type expansion, float normalisation or the exact contents of bundled profiles.

Existing general tests that verify bundled profiles can be loaded and parsed should remain unchanged.

## Update the PR description

Rewrite the PR description to match the reduced scope.

It must clearly state:

- which bundled profiles are corrected;
- the canonical upstream repository, source file and revision for each retained correction;
- what machine-relevant behaviour changes;
- how the corrected files were produced and verified;
- that Reaprime now explicitly rejects legacy simple-profile TCL ingestion instead of reimplementing de1app’s generators;
- that no runtime import, shot execution, BLE or USB behaviour is changed.

Remove discussion of:

- OpenSpec;
- provenance;
- broad corpus synchronisation with Decenza;
- beverage-type expansion;
- float normalisation;
- source collision handling;
- the removed cleaning and spring-lever corrections.

After narrowing the PR, run:

- the focused importer tests;
- `flutter analyze`;
- the full `flutter test` suite.

Include the exact results and identify any pre-existing failures separately.

This scope reduction is required before the PR can be reviewed for merge.

skialpine:
I think I will leave it to you to do with this information what you want. I just did not want the information lost and an issue felt like too little. Go ahead and rework as you see fit.

tadelv:
Thank you, Jeff, for the detailed investigation, source work, and profile corrections in this PR. The approved subset has now merged via #505 (merge commit `80f64cbdcbfc1020adfc9df7549ddd4062a6a060`): the five corrected legacy profiles, the canonical five-profile A-Flow set, and a focused guard against unsafe `settings_2a`/`settings_2b` TCL ingestion. The broader generator, provenance, normalization, OpenSpec, and unrelated profile changes were intentionally left out following review. Closing this PR as superseded by #505, with attribution preserved there.

PR #543, opened 2026-08-02
fix(scale): accept protocol-correct Acaia weight frames

ODevStudio:
## Summary

- Problem: protocol-correct Acaia Lunar/Pyxis frames were rejected because declared lengths and record tails were validated too narrowly.
- Why it matters: initialization could miss valid weight/settings notifications and repeatedly disconnect.
- What changed: accept real settings frames, short heartbeat buttons, bit-field weight flags, and complete weights with unknown/incomplete trailing records while preserving frame isolation.
- What did NOT change: readiness still requires a decoded valid weight; strict checksum rejection was not added; physical hardware validation remains outstanding.

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Docs
- [ ] Security hardening
- [ ] Chore / infra

## Scope

- [ ] REST API / handlers
- [ ] WebSocket API
- [x] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [ ] Docs / specs

## Independent Protocol Alignment

- The timer notification registration argument changed from `0x01` to `0x05`. It controls the requested number of heartbeats between timer notifications.
- Both maintained/reference implementations use `0x05`: [`encode_notification_request()` in aioacaia](https://github.com/zweckj/aioacaia/blob/main/aioacaia/encoder.py) and [`encodeNotificationRequest()` in pyacaia](https://github.com/lucapinello/pyacaia/blob/master/pyacaia/__init__.py).
- Focused unit coverage asserts the registration payload. Physical Lunar/Pyxis beta validation remains outstanding.
- This alignment is independent of and not required for the declared-length parser correction.

## Linked Issues

- Related #509
- Regression from #528

## Root Cause

- Root cause: Acaia byte 3 declares the span beginning at the length byte, including event and semantic records but excluding checksum bytes. The parser compared it to inner body sizes and required every trailing record to be recognized.
- Missing detection or guardrail: settings required exactly 3; all heartbeat events required 8; weight flags were treated as a scalar; partial-frame recovery could consume the next frame header as a flag byte.
- Contributing context (if known): real aioacaia captures use declared-length-9 settings frames, declared-length-6 button heartbeats, combined flag bits, and optional/unknown trailing records.

## Regression Test Plan

  - [x] Unit test
  - [ ] End-to-end test (simulate=1 + curl/websocat)
- Target test or file: `test/unit/models/acaia_scale_test.dart`
- Scenario the test should lock in: real settings battery 93, short heartbeat button liveness, flags 0x03 negative weight, direct and heartbeat weights with unknown/incomplete tails, and recovery after truncated frames.
- If no new test added, why not: New tests were added.

## Documentation Obligations

- [x] N/A - no docs affected

## Security Impact

- BLE/USB surface changed? Yes - notification parsing is more compatible, but remains bounded by declared length and known body validation.
- If any Yes, explain risk and mitigation: The parser processes only declared frame payloads; boundary-adjacent headers are used only for bounded resynchronization, incomplete weight bodies are rejected, and strict structural validation remains for non-weight heartbeat records.

## User-Visible Changes

Acaia Lunar/Pyxis scales can remain initialized when they send protocol-correct settings, button, or weight frame variants that were previously rejected.

### Local gates

- [x] Direct Dart formatter on both changed files - no remaining formatting changes
- [x] `flutter analyze --no-pub` - no issues found
- [x] `flutter test test/unit/models/acaia_scale_test.dart` - all 34 tests passed
- [ ] `flutter test --no-pub` - 2,581 passed; 14 unrelated environment/asset failures remain: QuickJS DLL loading, onboarding cascade, and missing bundled-skins manifest
- [ ] `(cd packages/dye2-plugin && npm run build)` - not run; no plugin files changed

### Manual verification

- OS / platform tested: Windows
- Simulated devices? (simulate=1): No
- What you personally verified and how: Protocol fixtures and mock BLE transport tests.
- Edge cases checked: split and concatenated frames, truncated weight recovery, real-record truncation at cuts 11-16 preserving the following frame, embedded headers, unknown records, settings length 9, button length 6, and combined flags.
- What you did not verify: physical LUNAR-E01ED7 or Pyxis hardware.

### Evidence

- [x] Test output (focused tests and analyzer)
- [x] Log snippets

## Compatibility & Migration

- Backward compatible? Yes
- If any No or Yes, explain exact steps: None.

## Risks & Mitigations

- Risk: Acaia firmware variants may still use unobserved record layouts.
  - Mitigation: frame-bounded parsing, strict non-weight record validation, and physical validation before merge.

### DIFF
--- a/lib/src/models/device/impl/acaia/acaia_scale.dart
+++ b/lib/src/models/device/impl/acaia/acaia_scale.dart
@@ -14,6 +14,8 @@ import 'package:rxdart/subjects.dart';
 
 enum AcaiaProtocol { ips, pyxis }
 
+enum _AcaiaFrameResult { accepted, ignored, malformed }
+
 class AcaiaScale implements Scale {
   static final _ipsService = BleServiceIdentifier.short('1820');
   static final _ipsCharacteristic = BleServiceIdentifier.short('2a80');
@@ -37,7 +39,10 @@ class AcaiaScale implements Scale {
   static const _header1 = 0xEF;
   static const _header2 = 0xDD;
   static const _metadataLength = 5;
+  static const _checksumLength = 2;
+  static const _weightBodyLength = 6;
   static const _maxPayloadLength = 64;
+  static const _recordBodyLengths = <int, int>{5: 6, 6: 1, 7: 3, 8: 1, 11: 2};
   static const _identPayload = [
     0x30,
     0x31,
@@ -62,7 +67,7 @@ class AcaiaScale implements Scale {
     0x01,
     0x02,
     0x02,
-    0x01,
+    0x05,
     0x03,
     0x04,
   ];
@@ -363,15 +368,9 @@ class AcaiaScale implements Scale {
   Future<void> resetTimer() async {}
 
   void _parseNotification(List<int> data) {
-    _buffer.addAll(data);
+    _buffer = [..._buffer, ...data];
     while (true) {
-      var headerIndex = -1;
-      for (var i = 0; i + 1 < _buffer.length; i++) {
-        if (_buffer[i] == _header1 && _buffer[i + 1] == _header2) {
-          headerIndex = i;
-          break;
-        }
-      }
+      final headerIndex = _findHeader(_buffer);
 
       if (headerIndex < 0) {
         _buffer = _buffer.isNotEmpty && _buffer.last == _header1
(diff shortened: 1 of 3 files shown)


lib/src/models/device/impl/acaia/acaia_scale.dart
```diff
@@ -406,17 +432,114 @@ class AcaiaScale implements Scale {
...
+      5 => _metadataLength,
+      11
+          when frame.length > _metadataLength + 2 &&
+              frame[_metadataLength + 2] == 5 =>
+        _metadataLength + 3,
+      _ => -1,
     };
+    if (weightOffset < 0) return false;
+    final nextHeader = _findHeader(frame, start: 2);
+    return nextHeader == weightOffset + _weightBodyLength - 1;
```
tadelv: Non-blocking: this recovery check only recognizes the next header when it lands in the missing final weight-flag slot. If a declared multi-record frame is cut after all six weight bytes—so only its trailing record/checksum is missing—the following frame begins at `weightOffset + _weightBodyLength`. The parser can then fill the first frame with bytes from the next one, accept the complete weight, and lose that next frame. Could we add a regression test using a truncated `_realWeightFrame` followed by a frame with a different weight, and preserve the second frame during resync?

lib/src/models/device/impl/acaia/acaia_scale.dart
```diff
@@ -62,7 +67,7 @@ class AcaiaScale implements Scale {
     0x01,
     0x02,
     0x02,
-    0x01,
+    0x05,
```
tadelv, line 70: Non-blocking scope note: changing this byte from `0x01` to `0x05` also changes the requested notification interval, independently of the declared-length parser fix. Since real hardware verification will need to come from beta users, could the PR summary explicitly call this out and cite the reference capture/implementation supporting `0x05`? Splitting it would also be reasonable, but I do not think this needs to block the parser fix.

tadelv (commented):
Two non-blocking observations from a light review. Neither is intended as a request for changes.

tadelv (approved):

PR #565, opened 2026-08-05
Add shot-upload plugin + enable plugin Decent-proxy writes

johnbuckman:
## Summary

- **Problem:** Decaid has no way to upload finished shots to the user's decentespresso.com account for cloud shot history and charts.
- **Why it matters:** lets users see their shot history and visualizations on the Decent website, gated to machines they own.
- **What changed:** a new bundled JS plugin `shot-upload.reaplugin` that, on shot completion, injects the connected machine's serial number (not carried in the shot JSON) and POSTs the shot to `support/api/shot_upload` through the **authenticated Decent proxy**, reusing the user's existing Decent login. To enable that, the plugin Decent-proxy bridge now allows `POST`/`PUT` with a body (it was `GET`-only); credentials are still never exposed to plugin JS and the `support/api/` allowlist is unchanged. Adds two `--dart-define` dev overrides (`DECENT_BASE_URL`, `MOCK_DE1_SERIAL`) to make local-server testing possible.
- **What did NOT change (scope boundary):** no Drift/schema changes; account credentials remain hidden from plugin JS; the Decent-proxy path allowlist (`support/api/`) is unchanged; existing GET proxy behavior is unchanged.

## Change Type (select all)

- [x] Feature
- [x] Docs

- [x] Plugins / JS runtime
- [x] Docs / specs

## Linked Issues

- Closes # N/A
- Related # N/A

## Root Cause (if bug fix)

- N/A (feature)

- N/A (feature). New unit coverage added — see below.

- [x] Plugin docs updated: `doc/Plugins.md` (documents that `host.decentProxy` now supports POST/PUT with a body)
- [x] N/A — no REST/WebSocket/Skins/Profiles/Device spec changes (the change is to the plugin-JS proxy bridge, not a new REST route)

## Security Impact (required)

- New or changed REST endpoints? **No**
- New or changed network calls? **Yes** — a permitted plugin can now `POST`/`PUT` (previously `GET` only) through the Decent account proxy.
- File system access changed? **No**
- Plugin sandbox boundary changed? **Yes** — the plugin Decent-proxy bridge now permits writes.

**Risk & mitigation:** writes remain gated by the `proxy.decent_api` permission, restricted to the existing `support/api/` allowlist, and the account credentials are never handed to plugin JS (the Dart proxy attaches Basic auth and relays only status/headers/body). Server-side authentication + machine-ownership checks still apply to every upload.

## User-Visible Changes

- New opt-in bundled plugin **"Decent shot upload"**. When enabled (and the user is logged into their Decent account), finished shots are uploaded automatically to the user's account. Settings: `AutoUpload` (default on) and `LengthThreshold` (skip flushes; default 5s). No credentials to enter — it reuses the existing Decent login.

- [x] `dart format lib test` — no remaining changes
- [x] `flutter test` — all pass
- [ ] `(cd packages/dye2-plugin && npm run build)` — N/A (dye2 not touched)

### Manual verification

- OS / platform tested: macOS (debug build)
- Simulated devices? (`simulate=1`): **Yes**
- Real hardware? (DE1/Bengle/scale): **No**
- What was verified and how: full live end-to-end in the running app — logged into a Decent account, pulled a shot on the simulated DE1, and confirmed the `shot-upload` plugin fetched the shot, injected the machine serial, and uploaded it through `host.decentProxy` to a local Decent server (`--dart-define=DECENT_BASE_URL=http://localhost:8000`, `--dart-define=MOCK_DE1_SERIAL=6262`). The server authenticated the account, verified machine ownership, stored the shot, and the shot rehydrated correctly on read.
- What was NOT verified: real hardware; production server (tested against a local Decent server).

### Evidence

- New unit test `test/plugins/plugin_decent_proxy_bridge_test.dart`: "declared permission forwards POST with body through DecentProxyService" asserts a permitted plugin's POST is forwarded upstream as a POST carrying the exact body + `application/json` + the account Basic auth; plus "unsupported method is rejected". Existing GET/permission/credential-isolation tests still pass (8/8 in the proxy suites).
- Live: decaid shot `de6ae026-…` uploaded → stored server-side with `profile_ref` and rehydrated with the full profile (Damian's LRv3, 119 measurements, serial 6262).

## Compatibility & Migration

- Backward compatible? **Yes**
- Config / env changes needed? **No** (the new `--dart-define`s default to production / the previous mock serial)
- Database migration needed? **No**

## Risks & Mitigations

- Risk: broadening the plugin proxy from GET-only to GET/POST/PUT enlarges what a permitted plugin can do with the account.
  - Mitigation: unchanged `proxy.decent_api` permission gate + `support/api/` allowlist + credentials never exposed to JS + server-side auth/ownership enforcement.

### DIFF
--- a/assets/plugins/shot-upload.reaplugin/plugin.js
+++ b/assets/plugins/shot-upload.reaplugin/plugin.js
@@ -1,16 +1,19 @@
 /* shot-upload.reaplugin
  *
  * Uploads finished espresso shots to the user's Decent account at
- * decentespresso.com (POST support/api/shot_upload). Authentication reuses the
- * Decent account the user is already logged into: the upload goes through
- * host.decentProxy, which attaches the account credentials in Dart and never
- * exposes them to plugin JS. The server verifies the account and that the
- * connected machine's serial belongs to it, then stores the shot.
+ * decentespresso.com (POST support/api/shot_upload) through the authenticated
+ * Decent proxy, reusing the account the user is already logged into. The proxy
+ * attaches the account credentials in Dart and never exposes them to plugin JS;
+ * the server verifies the account and that the connected machine's serial
+ * belongs to it, then stores the shot.
  *
- * There is no "shot stored" plugin event, so (like the Visualizer plugin) we
- * watch machine state for espresso -> non-espresso, then pull the just-stored
- * shot from the local REST API, inject the machine serial (which the shot JSON
- * does not carry) from /api/v1/machine/info, and upload it.
+ * Opt-in: AutoUpload defaults to FALSE, so nothing is uploaded until the user
+ * turns it on. (Beta stance. Post-beta, logging into the Decent account will
+ * itself serve as opt-in consent and AutoUpload will default to true.)
+ *
+ * The upload binds to the exact persisted shot via the `shotStored` event (fired
+ * with the shot id after persistence), so there is no timer/`/shots/latest`
+ * race, and machine identity is captured at completion time.
  *
  * Contract: must define createPlugin(host) returning {id, version, onLoad,
  * onUnload, onEvent}.
@@ -20,20 +23,18 @@ function createPlugin(host) {
   "use strict";
 
   const NS = "shot-upload.reaplugin";
+  const VERSION = "0.1.0";
   const LOCAL_API_URL = "http://localhost:8080/api/v1";
-  const UPLOAD_PATH = "support/api/shot_upload"; // relative to the Decent proxy base
-  const SHOT_FETCH_DELAY_MS = 5000; // let the app finish persisting the shot
+  const UPLOAD_PATH = "support/api/shot_upload"; // exact allowlisted proxy write path
   const RETRIES = 3;
   const RETRY_DELAY_MS = 2000;
 
-  let shotTimeoutId = null;
   let isUploading = false;
+  let decaidVersion = null;
 
   const state = {
-    autoUpload: true,
+    autoUpload: false, // opt-in; see header
     lengthThreshold: 5,
-    lastMachineState: null,
-    lastCheckedShotId: null,
     lastUploadedShot: null,
     lastResult: null,
   };
@@ -46,6 +47,14 @@ function createPlugin(host) {
     return await res.json();
   }
 
+  // Decaid app version (for provenance), cached.
+  async function getDecaidVersion() {
+    if (decaidVersion) return decaidVersion;
+    const info = await fetchLocal("/info");
+    decaidVersion = (info && (info.version || info.fullVersion)) || "unknown";
+    return decaidVersion;
+  }
+
   // seconds between first and last measurement
   function shotDuration(shot) {
     const m = shot && shot.measurements;
@@ -55,16 +64,16 @@ function createPlugin(host) {
     return isNaN(t0) || isNaN(t1) ? 0 : (t1 - t0) / 1000;
   }
 
-  // Inject machine identity (serial not carried in the shot JSON) + provenance.
+  // Inject machine identity (serial not carried in the shot JSON) + provenance,
+  // populated from the actual /machine/info fields (firmware = `version`).
   async function withMachine(shot) {
     const info = await fetchLocal("/machine/info");
-    const serial = info && (info.serialNumber || info.serial);
+    const serial = info && info.serialNumber;
     if (!serial) return null;
     shot.machine = { serialNumber: String(serial) };
-    if (info.id) shot.machine.bleId = String(info.id);
-    if (info.firmwareVersion) shot.machine.firmwareVersion = String(info.firmwareVersion);
+    if (info.version) shot.machine.firmwareVersion = String(info.version);
     if (info.model) shot.machine.model = String(info.model);
-    shot.app = { name: "decaid", version: "0.1.0", sourceFormat: "decaid" };
+    shot.app = { name: "decaid", version: await getDecaidVersion(), sourceFormat: "decaid" };
     shot.schemaVersion = 1;
     return shot;
   }
@@ -99,34 +108,41 @@ function createPlugin(host) {
     throw lastErr || new Error("upload failed");
   }
 
-  async function uploadShotById(shotId) {
+  // Upload one stored shot by id. Throws on failure (so callers can report
+  // status); marks e.skipped=true for a too-short shot. Returns the server result.
+  async function uploadShot(shotId) {
+    const full = await fetchLocal(`/shots/${shotId}`);
+    if (!full || !full.id) throw new Error(`shot ${shotId} not found`);
+
+    const dur = shotDuration(full);
+    if (dur < state.lengthThreshold) {
+      const e = new Error(`shot too short (${dur.toFixed(1)}s < ${state.lengthThreshold}s)`);
+      e.skipped = true;
+      throw e;
+    }
+
+    const payload = await withMachine(full);
+    if (!payload) throw new Error("no machine serial available");
+
+    const result = await postShot(payload);
+    state.lastUploadedShot = full.id;
+    state.lastResult = result;
+    host.storage({ type: "write", key: "lastUploadedShot", data: full.id });
+    host.emit("shotUploaded", { shotId: full.id, result: result, timestamp: Date.now() });
+    return result;
+  }
+
+  // Auto path: fire-and-forget with dedup + error handling (never throws).
+  async function autoUpload(shotId) {
     if (isUploading) return;
+    if (shotId && shotId === state.lastUploadedShot) { log(`shot ${shotId} already uploaded`); return; }
     isUploading = true;
     try {
-      let meta = shotId ? { id: shotId } : await fetchLocal("/shots/latest");
-      if (!meta || !meta.id) { log("no shot available"); return; }
-
-      if (!shotId && meta.id === state.lastCheckedShotId) { log(`shot ${meta.id} already handled`); return; }
-      state.lastCheckedShotId = meta.id;
-
-      const full = await fetchLocal(`/shots/${meta.id}`);
-      if (!full) { log(`could not fetch shot ${meta.id}`); return; }
-
-      const dur = shotDuration(full);
-      if (dur < state.lengthThreshold) { log(`shot ${meta.id} too short (${dur.toFixed(1)}s); skipping`); return; }
-
-      const payload = await withMachine(full);
-      if (!payload) { log("no machine serial available; skipping"); return; }
-
-      const result = await postShot(payload);
-      state.lastUploadedShot = full.id;
-      state.lastResult = result;
-      host.storage({ type: "write", key: "lastUploadedShot", namespace: NS, data: full.id });
-      log(`uploaded ${full.id} -> ${result && result.profile_ref ? result.profile_ref : "ok"}`);
-      host.emit("shotUploaded", { shotId: full.id, result: result, timestamp: Date.now() });
+      const r = await uploadShot(shotId);
+      log(`uploaded ${shotId} -> ${r && r.profile_ref ? r.profile_ref : "ok"}`);
     } catch (e) {
-      log(`error: ${e.message}`);
-      host.emit("uploadError", { error: e.message, timestamp: Date.now() });
+      if (e.skipped) { log(`skipped ${shotId}: ${e.message}`); }
+      else { log(`error uploading ${shotId}: ${e.message}`); host.emit("uploadError", { shotId: shotId, error: e.message, timestamp: Date.now() }); }
     } finally {
       isUploading = false;
     }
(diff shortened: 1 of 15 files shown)


lib/src/plugins/plugin_loader_service.dart
```diff
@@ -470,6 +470,7 @@ class PluginLoaderService {
       'assets/plugins/settings.reaplugin',
       'assets/plugins/dye2.reaplugin',
       'assets/plugins/decent-profile.reaplugin',
+      'assets/plugins/shot-upload.reaplugin',
```
tadelv, line 473: **P1 — this is not opt-in in practice.** `_ensureBundledPluginsAutoLoadEnabled()` sets a missing autoload key to `true` for every path in this list. On the first launch after upgrading, this new plugin is therefore loaded automatically; with no saved settings, `applySettings({})` also defaults `AutoUpload` to `true`. A user who is already logged into a Decent account will have the next shot uploaded without taking any affirmative action, contrary to the PR description. Please keep this uploader disabled by default (or default `AutoUpload` to false) until explicit consent, and cover the upgrade/fresh-install behavior with a test.

lib/src/plugins/plugin_decent_proxy_bridge.dart
```diff
@@ -33,15 +35,36 @@ class PluginDecentProxyBridge {
...
+    final callerId = 'plugin:$pluginId';
+    final DecentProxyResponse response;
+    switch (normalizedMethod) {
+      case 'GET':
+        response = await decentProxyService!.proxyGet(
+          callerId: callerId,
+          path: path,
+          query: query,
+        );
+      case 'POST':
```
tadelv, line 55: **P1 security — do not silently turn the existing read permission into a general write credential.** `proxy.decent_api` previously granted GET-only access, but this switch now gives any already-installed plugin declaring it POST and PUT access across the entire `support/api/` namespace. The unchanged permission name gives users no indication that its authority has expanded, and server-side authentication only confines writes to their own account; it does not prevent an authorized plugin from making unwanted writes. This feature only needs `POST support/api/shot_upload`: please introduce a distinct write/shot-upload capability and enforce method + path, rather than enabling unused PUT and arbitrary `support/api/` writes.

assets/plugins/shot-upload.reaplugin/plugin.js
```diff
@@ -0,0 +1,196 @@
...
+      if (i < RETRIES - 1) await new Promise(r => setTimeout(r, RETRY_DELAY_MS));
+    }
+    throw lastErr || new Error("upload failed");
+  }
+
+  async function uploadShotById(shotId) {
+    if (isUploading) return;
+    isUploading = true;
+    try {
+      let meta = shotId ? { id: shotId } : await fetchLocal("/shots/latest");
```
tadelv: **P1 correctness — bind the upload to the exact persisted shot instead of looking up `/shots/latest` after a fixed delay.** A routine sequence can lose the real shot: finish an espresso, quickly run a short flush within this five-second window, and the flush becomes `latest`; the threshold then skips the flush while the actual espresso is never uploaded. Slow/failed persistence or concurrent import/sync can similarly select an old or unrelated record. Please emit a `shotStored` plugin event only after persistence succeeds, include the exact shot ID, and upload that ID. That removes the timer race and provides a reliable point for retry/accounting.

assets/plugins/shot-upload.reaplugin/plugin.js
```diff
@@ -0,0 +1,196 @@
...
+        }
+        case "settingsUpdated":
+          applySettings(event.payload);
+          break;
+      }
+    },
+
+    // Optional control endpoints (POST upload {shotId}, GET status).
+    async __httpRequestHandler(request) {
+      const id = (request && request.id) || "";
```
tadelv: **P2 correctness — both advertised HTTP endpoints use the wrong host contract.** `PluginsHandler` supplies `{requestId, endpoint, method, headers, body, query}`, but this code reads `request.id` and `request.data`, so both `status` and `upload` always fall through to `unknown endpoint` (reproduced with the real payload shape). The returned `body` should also be JSON text with a content-type header, matching the existing plugins, and `uploadShotById()` currently catches failures internally so this endpoint would report `{ok:true}` even when no upload occurred. Please use `request.endpoint`/`request.body`, serialize the response, propagate failure status, and add endpoint tests.

assets/plugins/shot-upload.reaplugin/plugin.js
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  return {
+    id: NS,
+    version: "0.1.0",
+
+    onLoad(settings) {
+      applySettings(settings);
+      try {
+        const saved = host.storage({ type: "read", key: "lastUploadedShot", namespace: NS });
```
tadelv: **P2 correctness — plugin storage reads are event-based.** `host.storage()` sends a command and returns `undefined`; the value arrives later through a `storageRead` event. Consequently `lastUploadedShot` is never restored here, and there is no `storageRead` case in `onEvent`. Handle that event and use the restored ID for durable idempotency/status (or remove the unused persistence), with a reload test.

assets/plugins/shot-upload.reaplugin/plugin.js
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  // Inject machine identity (serial not carried in the shot JSON) + provenance.
+  async function withMachine(shot) {
+    const info = await fetchLocal("/machine/info");
+    const serial = info && (info.serialNumber || info.serial);
+    if (!serial) return null;
+    shot.machine = { serialNumber: String(serial) };
+    if (info.id) shot.machine.bleId = String(info.id);
+    if (info.firmwareVersion) shot.machine.firmwareVersion = String(info.firmwareVersion);
```
tadelv: **P2 correctness — the uploaded provenance is inaccurate and can identify the wrong machine.** `/machine/info` exposes firmware as `version`, not `firmwareVersion`, and has no `id`, so those branches never populate the claimed fields. `shot.app.version` below is also hard-coded to the plugin version `0.1.0`, not the running Decaid version available from `/api/v1/info`. In addition, fetching the serial at upload time means a manual historical `shotId` is attributed to whichever machine is currently connected, not necessarily the machine that produced it. Capture machine identity alongside the exact completed-shot event (or restrict manual upload accordingly) and populate provenance from the actual API fields.

tadelv (changes requested):
Requesting changes. The underlying proxy still keeps credentials out of plugin JS and the bridge tests pass, but the feature currently uploads personal shot data without the stated opt-in, silently expands an existing read permission into a broad authenticated write capability, and can miss the intended shot because it discovers the record through a delayed `/shots/latest` lookup. The advertised HTTP endpoints are also non-functional. Inline comments cover those blockers plus the storage and provenance correctness issues. Please add behavior-level coverage for the plugin lifecycle and HTTP endpoints; the current tests exercise only the Dart forwarding layer.

johnbuckman:
Thanks for the thorough review — all six addressed in `a395dd9d`, plus behavior-level tests. Summary:

**P1 — opt-in.** `AutoUpload` now defaults to **false** (`applySettings` treats it as `settings.AutoUpload === true`; manifest default `false`), so a logged-in user uploads nothing until they turn it on. New test `does not upload when AutoUpload is off (opt-in default)` locks this in. (Note for later: post-beta, logging into the Decent account is intended to serve as opt-in consent, at which point the default flips to true.)

**P1 — write capability.** `proxy.decent_api` stays **GET-only**. Added a distinct `proxy.decent_api.write` permission; the bridge now enforces **method + path**: `POST` only, and only to an explicit allowlist (currently just `support/api/shot_upload`). `PUT` and arbitrary `support/api/` writes are gone. Tests cover: POST forwards with the write perm; POST without it rejected; POST to a non-allowlisted path rejected; PUT unsupported.

**P1 — bind to the exact shot.** Added a core `shotStored` plugin event fired with the shot id right after persistence (`PersistenceController.onShotStored`, wired in `main.dart` to `broadcastEvent('shotStored', {id})`). The plugin uploads **that exact id** on the event — no timer, no `/shots/latest` race, no flush-steals-the-slot. Machine identity is captured at that point too.

**P2 — HTTP endpoints.** Now read `request.endpoint`/`request.body`, return `{status, headers:{'Content-Type':'application/json'}, body: JSON.stringify(...)}`, and propagate real failure (upload errors surface as 502/skip, not `{ok:true}`). `upload` uploads the **latest** shot (from the currently-connected machine, avoiding the historical-id misattribution). New test drives the `upload` endpoint end-to-end.

**P2 — storage.** `lastUploadedShot` is restored via the event-based `storageRead` (handled in `onEvent`), used for dedup. Test: a storageRead-restored id is not re-uploaded.

**P2 — provenance.** Firmware now comes from `/machine/info` `version`; the never-populated `firmwareVersion`/`id` branches are gone (no bogus `bleId`); app version comes from `/api/v1/info` (not the hard-coded plugin version). The auto path captures identity at the `shotStored` event; manual upload targets the latest shot only. The provenance test asserts `firmwareVersion` from `version` and `app.version` from `/api/v1/info`.

**Tests.** Added `test/plugins/shot_upload_plugin_test.dart` (lifecycle: shotStored→upload via the real proxy round-trip, opt-in-off, dedup, and the HTTP `upload` endpoint) beyond the Dart forwarding layer. Full suite green (2736).

tadelv:
The two failing tests appear to have the same test-harness problem: they wait for the `shotUploaded` event before advancing QuickJS’s pending Promise jobs.

The mocked `fetch` implementation is asynchronous, and the plugin performs several awaited operations. In both tests, execution reaches an `await fetch(...)` in JavaScript, but the Dart test then waits on:

```dart
await uploaded.timeout(const Duration(seconds: 5));
```

without calling `manager.js.executePendingJob()` first. The Promise continuation that emits `shotUploaded` therefore never runs, so the test times out.

The endpoint test has the same circular ordering: pending jobs are pumped only after waiting for `shotUploaded`, although producing `shotUploaded` requires those jobs to be pumped.

Please add a helper that repeatedly:

1. Drains QuickJS pending jobs.
2. Yields to the Dart event loop so `MockClient` and bridge callbacks can complete.
3. Stops once the expected event or HTTP response is available.

For example, the core loop could look like:

```dart
while (!completed && DateTime.now().isBefore(deadline)) {
  while (manager.js.executePendingJob() > 0) {}
  await Future<void>.delayed(Duration.zero);
}
```

Use that loop while waiting for `shotUploaded`, and in the HTTP endpoint test continue pumping until `getPendingHttpResponse('req-1')` returns a response.

It is also worth checking the production bridge calls that currently invoke `executePendingJob()` only once after completing asynchronous native work. Draining all currently pending jobs may be safer, but the immediate CI failures are caused by the tests waiting before pumping the QuickJS Promise queue.

lib/src/controllers/persistence_controller.dart
```diff
@@ -11,6 +11,12 @@ class PersistenceController {
 
   PersistenceController({required this.storageService});
 
+  /// Optional hook fired with the shot id immediately after a shot is
```
tadelv: no reason to add 4 lines of comments for 1 line of code. The code is well self-documented

assets/plugins/shot-upload.reaplugin/plugin.js
```diff
@@ -0,0 +1,220 @@
...
+          lastUploaded: state.lastUploadedShot,
+          lastResult: state.lastResult,
+        });
+      }
+      if (endpoint === "upload") {
+        try {
+          const latest = await fetchLocal("/shots/latest");
+          if (!latest || !latest.id) return jsonResponse(404, { ok: false, error: "no shot available" });
+          const result = await uploadShot(latest.id);
+          return jsonResponse(200, { ok: true, id: latest.id, result: result });
```
tadelv, line 211: if `result` contains backend id , plugin should backfill that to the shot record and save

tadelv (changes requested):

tadelv (approved):
I'll approve and merge for v1

assets/plugins/shot-upload.reaplugin/plugin.js
johnbuckman: On the id itself: decaid ids aren't globally unique — live shots use `Uuid().v4()`, but de1app-imported shots use `de1app-<clock>` (a start-timestamp), which can collide across machines. The stable globally-unique key is `<machine_sn>/<decaid_id>`: the SN is hardware-unique, and the decaid id is already unique *within* a machine (it's the primary key of that machine's per-SN store on the server). That `<sn>/<id>` is the canonical id the server uses.

On backfilling it onto the shot record: we'd rather not — the ref is deterministically derivable inside decaid from the shot's own id plus the machine SN, so there's nothing to persist or keep in sync (and no new backend id is minted; the server keys by that pair). Happy to have the app carry it explicitly if you'd prefer, but it'd just duplicate data already on hand. Does derive-on-the-fly work for you, or is there a case where the record needs the ref materialized?

johnbuckman:
Pushed `34dc0535` addressing the follow-up review:

- **Failing tests** — spot on. The two upload tests awaited `shotUploaded` / the HTTP response *before* pumping QuickJS's pending jobs, so the JS-stubbed `fetch` promises never resolved and they timed out. Added a `_pumpUntil(manager, ready)` helper that drains jobs (`while (js.executePendingJob() > 0) {}`) and yields to the Dart loop until the condition holds, and used it in both tests. All 4 plugin tests pass locally.
- **Production single-pump** — good catch, this was a real bug. `_completeDecentProxyRequest` and both `__fetchResponse__` paths pumped a single job, but resolving a proxy/fetch promise enqueues a whole then-chain, so the plugin's continuation could stall mid-upload. All three are now full drain loops.
- **`onShotStored` comment** — removed.

The backfill / shot-id point is answered inline on `plugin.js:211`.

assets/plugins/shot-upload.reaplugin/plugin.js
tadelv: imagine the scenario where a user would pull a shot on machine A, shot-upload would upload it automatically. 
Now the same user would take Decaid and connect to machine B and call plugin's `upload` method manually. 
Now we have the same shot record, made originally by machine A, stored under both machine A and machine B.
I think we could locally associate the serial number with the shot record as well - users might want to know locally which machine made a shot. Up to you, I just wanted to share my thoughts

PR #569, opened 2026-08-06
Stream backup export, import, and sync data instead of fully buffering it (issue #555)

### LINKED ISSUE #555: Stream backup export, import, and sync data instead of fully buffering it
## Problem

The backup pipeline keeps several full representations of the same data in memory at once.

On the server, [`DataExportHandler`](https://github.com/decentespresso/decaid/blob/f8cd95535c10215630fd60787a1633e4524a427c/lib/src/services/webserver/data_export_handler.dart) materializes section models, JSON strings, an in-memory `Archive`, and the final ZIP byte list. Import first buffers the entire request with `toList()`, then decodes the full ZIP and section JSON.

The native data-management caller also accumulates an export in a `BytesBuilder` before saving it and reads an entire import file before uploading it in [`data_management_page.dart`](https://github.com/decentespresso/decaid/blob/f8cd95535c10215630fd60787a1633e4524a427c/lib/src/settings/data_management_page.dart). [`DataSyncHandler`](https://github.com/decentespresso/decaid/blob/f8cd95535c10215630fd60787a1633e4524a427c/lib/src/services/webserver/data_sync_handler.dart) builds on the same byte-oriented export/import APIs.

As shot history grows, especially with measurement arrays, these overlapping copies can create substantial memory pressure and plausible out-of-memory termination on iOS and other constrained devices.

## Expected behavior

Backup export, import, and sync should use streaming and/or temporary-file-backed processing so peak memory does not contain multiple complete copies of a backup. The ZIP format and public API semantics must remain compatible.

## Acceptance criteria

- Add a design/implementation plan under `doc/plans/` before changing this multi-subsystem flow.
- Cover all of these paths:
  - server-side export response;
  - server-side import request;
  - native full-backup save and upload;
  - pull, push, and two-way sync.
- Use streaming archive/file APIs or bounded temporary-file staging, with reliable cleanup on success, error, cancellation, and timeout.
- Avoid converting a full ZIP or full HTTP body into `List<int>`, `Uint8List`, `BytesBuilder`, or equivalent unless a platform API makes that unavoidable; document any unavoidable platform-specific exception.
- Preserve the existing backup format, conflict handling, selected-section behavior, and transfer-result semantics.
- Add practical request/archive size and malformed-input protections where needed to keep resource use bounded.
- Add tests that exercise a large synthetic backup and demonstrate bounded processing behavior without depending on physical hardware.
- Update API/docs if transport details or limits become externally visible.

tadelv:
## Summary

- **Problem:** The backup pipeline kept several full representations of the same data in memory at once — section models, JSON strings, an in-memory `Archive`, the final ZIP byte list, plus full `BytesBuilder`/`readAsBytes` buffers in the native page and byte-oriented sync APIs. As shot history grows (especially with measurement arrays) these overlapping copies can terminate constrained iOS devices. `archive` 4.0.9's `ZipEncoder.add()` deflates each entry into an in-memory `OutputMemoryStream` (a section-sized compressed buffer) and `ZipDecoder.decodeStream()` materializes every entry, so "use `InputFileStream`/`ZipFileEncoder`" would not have fixed it.
- **Why it matters:** Peak memory now scales with one page of records and one JSON record, never with backup size.
- **What changed:** Export/import/sync/native-transfer all stream through uniquely owned temporary files; sections stream records via keyset-cursor paging; JSON is parsed with a real incremental parser; imports run a bounded validation pass before importing; ZIP I/O uses a small dependency-free streaming writer + file-backed reader with hard limits (ZIP-bomb, CRC, truncation, Zip64, duplicate-name, encrypted/unsupported-entry protection).
- **What did NOT change (scope boundary):** ZIP entry names, JSON shapes, `metadata.json` semantics, conflict strategies, selected-section behavior, platform warnings, atomic export, non-transactional section isolation, and `200`/`207`/`400` + sync phase semantics. Legacy "Export Shots", log export, and the DE1-app folder importer keep their current in-memory paths (tracked follow-up). Issue #559 is untouched.

## Change Type

- [x] Feature
- [x] Refactor required for the fix
- [x] Docs
- [x] Security hardening

## Scope

- [x] REST API / handlers
- [x] Storage / Drift database
- [x] Docs / specs

## Linked Issues

- Fixes #555

## Root Cause

- Root cause: byte-oriented ZIP + JSON + HTTP handling made memory proportional to backup size; the `archive` package's entry deflate path hides a section-sized in-memory buffer.
- Missing detection or guardrail: no per-entry/per-record/request size limits existed.
- Contributing context: shot records carry full measurement arrays; overlapping copies were plausible OOM on iOS.

## Regression Test Plan

  - [x] Unit test
  - [x] End-to-end test (`simulate=1` + curl/websocat)
- Target test or file: `test/data_export/*` (130 tests), `test/database/*` keyset paging tests, simulate-mode curl round-trip (below).
- Scenario the test should lock in: bounded paging (instrumented page seams), old-in-memory backup imports, new export decodes with `ZipDecoder`, malformed JSON never imports a prefix, limits at/over boundaries, temp-artifact cleanup on every failure path, ZIP bombs/duplicates/CRC/truncation fail safely.

## Documentation Obligations

- [x] API spec updated: `assets/api/rest_v1.yml`
- [ ] N/A — plugin/skin/profile/device docs unaffected

## Security Impact

- New or changed REST endpoints? **No** (same endpoints, streamed transport)
- New or changed network calls? **Yes** — sync pull/push now stream; timeouts split into header (10 s), idle (30 s), overall (10 min); response bodies capped (8 MiB).
- File system access changed? **Yes** — temp ZIP staging in `Directory.systemTemp` with per-operation unique dirs, deleted on success/error/cancel/timeout; native export defers cleanup (grace timer) because the OS share sheet reads asynchronously.
- Plugin sandbox boundary changed? **No**
- Risk and mitigation: hostile archives (ZIP bombs, Zip64, encrypted/unsupported entries, duplicate names, CRC failures, truncation, oversized records, excessive entry counts) all fail safely with `400`; limits centralized in `data_transfer_limits.dart` and documented in the spec.

## User-Visible Changes

Full-backup export on iOS/Android now uses the OS share/save sheet with a temporary file instead of `FilePicker.saveFile(bytes:)`; desktop still uses a save dialog. No other user-visible changes.

### Local gates

- [x] `dart format lib test` — clean
- [x] `flutter analyze` — clean (0 warnings/errors)
- [x] `flutter test` — **2747 tests pass**
- [x] `(cd packages/dye2-plugin && npm run build)` — builds

### Manual verification

- OS / platform tested: macOS (simulate mode), full test suite
- Simulated devices: `MockDe1`
- What you personally verified and how:
  - `GET /api/v1/data/export` → 3.3 MB baseline zip, 9 entries, 569 shots, `formatVersion: 1` (decoded with an independent zipfile decoder)
  - Streamed a 28 MB compressed / ~430 MB uncompressed synthetic backup (4000 shots with 120-frame measurements, built from a real baseline shot shape) via `curl --data-binary` → **HTTP 200, imported 4000, 0 errors, 42 s**
  - Re-exported → 30.6 MB streamed download, 4570 shots, all sections present
  - A 70 MB compressed / 1.2 GB uncompressed archive was **correctly rejected with HTTP 400** ("ZIP entry exceeds the uncompressed size limit") — the 1 GiB per-entry cap works
- Edge cases checked: truncated section JSON → 207 with section errors and no prefix import; individually invalid records → partial results; ZIP bombs/duplicates/CRC/truncation/Zip64/encrypted entries → safe 400; limits at/above boundaries via injected tiny limits; temp dirs removed after success, section failure, malformed import, oversized body, concurrent imports.
- What you did **not** verify: real iOS/Android share-sheet behavior (no physical device); native picker/share flows are covered by `backup_transfer_service_test.dart` against a loopback HTTP server.

### Evidence

- [x] Test output: `flutter test` → 2747 passed; `flutter analyze` clean
- [x] curl output: HTTP 200 import of 4000-shot backup, HTTP 200 streamed re-export, HTTP 400 rejection of 1.2 GB entry

## Compatibility & Migration

- Backward compatible? **Yes** — existing backups import through the new reader (verified: the old in-memory `ZipEncoder` output path is the test fixture for handler/sync tests), and newly exported backups decode with the previous `ZipDecoder` (verified in tests and via Python zipfile).
- Config / env changes needed? **No**
- Database migration needed? **No**

## Risks & Mitigations

- Risk: new streaming ZIP writer/reader diverge from `archive` on some edge case. Mitigation: byte-compatible verification against `ZipDecoder` in tests, plus CRC + size verification on read; writer emits standard data-descriptor ZIPs.
- Risk: per-entry inflate is CPU-bound in the handler isolate (same as before — the old encoder was also synchronous). Mitigation: unchanged behavior; pages stay small.
- Risk: deferred native-share cleanup could leak a temp file if the app is killed before the grace timer. Mitigation: files live in the OS temp directory which the system reclaims; timer deletes them in the normal case.

## Before/After memory model

| Stage | Before | After |
|-------|--------|-------|
| Server export | whole section model + whole JSON string + in-memory `Archive` + final ZIP byte list (multiples of backup size) | 1 page of records (≤200) + 1 encoded record (≤64 MiB) + deflate windows (~64 KiB) |
| Server import | full request `toList()` + full ZIP decode + full section JSON | streamed body (2 GiB cap) + per-entry bounded inflate + 1 record at a time |
| Native export | `BytesBuilder` accumulating the whole backup | streamed response to temp file; OS share sheet |
| Native import | `readAsBytes()` + `request.add(bytes)` | picker path/stream → `addStream` with known content length |
| Sync | `response.bodyBytes` / `exportToBytes` full buffers | pull → temp ZIP → file import; push → temp ZIP → streamed request with Content-Length |

Temporary files: one unique `Directory.systemTemp` subdir per operation, deleted in `finally` or on stream done/cancel; native export defers deletion (5 min grace) for the share sheet.

Chosen limits (documented in `assets/api/rest_v1.yml` and `doc/Api.md`): request body 2 GiB, entry count 4096, per-entry uncompressed 1 GiB, total uncompressed 2 GiB, metadata 64 KiB, per-record 64 MiB, ZIP header fields 256 B/64 KiB/64 KiB, sync request 1 MiB, target response 8 MiB; header 10 s / idle 30 s / overall 10 min timeouts. All below ZIP64 thresholds so Zip64 is rejected outright.

ZIP compatibility evidence: new exports decode with `archive`'s `ZipDecoder` (handler/zip tests), with Python's independent `zipfile` (smoke test), and old `ZipEncoder`-produced backups import through the new reader (handler + sync tests). Remaining bounded platform-specific exception: the OS share sheet reads the temp file asynchronously, so native-export cleanup is deferred.


tadelv:
Addressed all three P1 blockers plus the P2 items and the P3 leak, in commit `374b24da`. CI green (2747 → 2755 tests).

**P1-1 — Sync timeouts reject legitimate large syncs.** The 10 s "header timeout" wrapped `send()`, which for pull/push completes only after the target has generated/imported the whole archive (the 42 s smoke import would have failed). Now:
- The default sync client sets `HttpClient.connectionTimeout = 10 s` — TCP establishment only, never counted against server-side processing (`webserver_service.dart`).
- Phases are bounded by the 10 min overall timeout, applied **inside** `_pull`/`_push` (send, response stream, and response body each), so a timeout can't strand the phase.
- Timeouts cancel explicitly: pull uses a `listen`-based consumer whose subscription is cancelled in `finally`; a timed-out pull **never runs `importFromZipFile`** — no DB mutation after the caller saw the timeout. Push cancels via sink/response handling and `bodyFuture.ignore()` prevents unhandled async errors.
- Tests: slow-target pull and push (100 ms target delay, 1 ms connect timeout → success), and overall-timeout pull (502, reason `timeout`, section `calls == 0`).

**P1-2 — Container values bypass `maxRecordBytes`.** `_completeValue` only checked scalar/string tokens; object/array events went through `_closeContainer` → `_emitEvent(span)` unchecked. The event span is now bounded **incrementally while constructed** (`_spanWriteCharCode` tracks UTF-8 byte length per code unit), and all token/key limits measure UTF-8 bytes instead of UTF-16 code units (surrogate pairs count conservatively). Tests: object with many small members trips the cap with no oversized token; whole-document (depth 0) cap for singletons; multibyte string boundary test.

**P1-3 — Export streaming ignores backpressure/cancellation.** Replaced the `StreamController` with an async* generator (`yield* file.openRead()` + `finally` cleanup): pause/resume/cancel propagate to the source file stream, so a slow HTTP consumer can't re-accumulate the ZIP, and cleanup runs on done/error/cancel. 30 lines deleted.

**P2-1 — Total export limit not enforced.** `StreamingZipWriter` now accumulates `_totalUncompressed` and rejects writes past `maxTotalUncompressedBytes` (test: 98-byte entry passes, second write trips a 100-byte cap).

**P2-2 — ZIP integrity failures became 207.** `_importSection`'s validation pass and the section loop now convert `ZipReadException` (CRC mismatch, truncated compressed data, unsupported/encrypted entries) to `InvalidBackupException` → HTTP 400 for the whole import. Only JSON-level failures remain section-local 207s. Test: CD CRC byte flipped → 400.

**P2-3 — Inflate limit checked after accumulation.** The per-entry cap is enforced inside `_ChunkAccumulator.add` (the decoder's output sink): a forged entry is rejected as soon as the inflater exceeds the declared size; pending chunks are cleared on overflow.

**P3 — Native export temp leak on unmount.** `_exportFullBackup` now disposes `tempDir` before the `!mounted` early return.

Docs: `rest_v1.yml`, `doc/Api.md`, `AI_API_NOTES.md` updated for the new timeout semantics.

tadelv:
All three issues addressed in `7ef96b8`. CI green, 2758 tests pass (+3).

**P2 — Push timeout doesn't cancel the upload / not a real phase timeout.**
- `_push` now feeds the request body through a cancellable `listen` subscription (replacing the fire-and-forget `addStream`). On timeout or any phase failure, `abortUpload()` cancels the file read and closes the request body, so the remote receives a truncated archive and **cannot complete its import** after the caller saw the timeout.
- Both phases now run on a **single deadline** (`syncOverallTimeout`): local export, waiting for headers, streaming the body, reading the response, and the pull-side local import all consume from the same `_remaining(deadline)`, so a 10-minute phase cannot run substantially longer than 10 minutes. Pull's local `importFromZipFile` is included in the deadline.
- Test: push with a 50 ms deadline against a target that responds at 300 ms → 502, reason `timeout`, and the phase completes in <250 ms (at the deadline, not after the target's delay).

**P2 — Export record cap measured in UTF-16.** `_EntryJsonSink.writeRaw` now encodes the fragment once, checks `encoded.length` (UTF-8 bytes) against `maxRecordBytes`, and hands the `Uint8List` straight to the ZIP writer — no second copy. Test: 40 `é` chars (40 UTF-16 units, 80 UTF-8 bytes) with a 64-byte cap → export fails.

**P2 — Export not bounded by the import request limit.** `StreamingZipWriter.close()` now rejects a completed archive whose size (local data + central directory + EOCD) would exceed `maxImportRequestBytes`, so Decaid can never generate a backup its own import/sync endpoints reject. Test: 600 incompressible bytes with a 500-byte request limit → `ZipWriteException`, temp file deleted.

Docs (`rest_v1.yml`, `doc/Api.md`, `AI_API_NOTES.md`) updated: one overall deadline per phase covering every network stage plus pull-side import; push timeouts abort the upload; archives bounded by the 2 GiB import request limit; per-record caps in UTF-8 bytes.

tadelv:
All three issues addressed in `7b48b57`. CI green, 2759 tests pass.

**P1 — Timed-out push can still complete remotely.** The guarantee is now honest:
- The upload uses `http.AbortableStreamedRequest` with an abort trigger, so a deadline expiry while the body is still being sent aborts the transport (socket close for IOClient) — a push that is still uploading cannot complete remotely.
- Once the archive was fully uploaded before the deadline, the remote outcome is unknowable — the target may have already begun importing, and a completed upload cannot be recalled. The phase now reports reason **`timeout_unknown`** with the message "the target may have already imported the pushed archive", instead of claiming the push did not happen.
- The "deadline" claim was also narrowed in docs and code: it bounds the **network stages** (upload/download + response); local export and pull-side import are not cut off.
- Test updated: push timeout after the mock consumed the body → 502 with `reason: timeout_unknown`, still failing at the deadline (<250 ms, not after the target's 300 ms delay).

**P1 — Pull import continues mutating after timeout.** Removed `Future.timeout()` from `importFromZipFile`. Abandoning it mid-write left shots/records inserting in the background after the caller saw 502, and the `finally` deleted the temp dir under the still-running reader. The import now runs to completion and reports its actual result; the deadline covers the download only, and a timed-out download cancels the stream so the import never starts. New test: a pull whose local import outlasts the deadline still returns 200 with the real import result and one import call.

**P2 — Upload-body errors can become unhandled.** `bodyDone.future.ignore()` is attached immediately after the body feed starts, so an upload failure while `send()` is still pending can never surface as an unhandled async error; the awaited path still reports the failure as a phase error.

Docs (`rest_v1.yml`, `doc/Api.md`, `AI_API_NOTES.md`) updated: deadline = network stages only; local export/import run to completion; `timeout_unknown` documented.

tadelv:
All three findings addressed in `eee37e0`. CI green, 2761 tests pass (incl. two new real-transport regression tests).

**P1 — Push no longer applies backpressure.** The body is now fed with `request.sink.addStream(bodyStream())` instead of a manual `listen`+`add`. `addStream` pauses the source subscription while the sink's listener is paused, so on a slow transport the `zipFile.openRead()` inside the generator pauses — the unsent archive can no longer accumulate in the controller buffer. The sink is closed only after the file is fully read (`addStream` does not forward the source's done event) and never after an abort, when the controller is already cancelled.
Regression test (would hang/pass wrongly with the old feeder): a real `IOClient` pushes a 6 MB archive to a `ServerSocket` that never reads. The kernel buffer fills, the upload stalls mid-body, and the phase must report `timeout` (upload still in flight) — never `timeout_unknown`. With the old unbounded feeder the file drained into the controller and the phase reported `timeout_unknown`, so the test fails against the old code.

**P2 — Pull now aborts the pre-header request.** The GET uses `http.AbortableRequest` with a `Timer` that completes the abort trigger at the deadline. `Future.timeout()` still only stops waiting, but now the transport itself is aborted (`ioRequest.abort` in IOClient), so a timeout while the target is still generating its export closes the connection instead of leaving it running. `RequestAbortedException` (thrown by the abort) maps to reason `timeout` so the streaming-stage race resolves deterministically. Test: a real `ServerSocket` that never responds must observe the connection being closed right after the deadline — without the transport abort that future never completes.

**P2 — bodySent is now honest.** The flag is set only when the file read finishes (the `addStream` future also completes successfully on cancellation, so `isCompleted`-style checks are unusable). A body read error or an aborted upload therefore always reports as an interrupted upload (`timeout`), never `timeout_unknown`.

Docs updated: timeouts abort at the transport level on both phases; the upload is backpressured so the archive file is never read ahead of the network.

ODevStudio (approved):
Reviewed your commits and added:

- b326ac01: made imports two-phase and atomic for ZIP/structural validation, replaced the custom decompressor with archive's file-backed extraction path, and tightened store.json validation.
- 7f47c8d4: unified rune-aware UTF-8 key limits across KV export/import, made push cancellation wait for body shutdown, removed repeated profile/KV scans, and restored newest-first shot keyset pagination.
- 02267f39: preserved partial settings import counts, bounded per-section semantic errors, and rejected oversized nested KV map keys during export.
- 92826895: validated non-KV root container shapes during prevalidation so a later wrong-shaped section cannot follow earlier mutations.

The requested P1 and P2 correctness issues are covered by focused regression tests. Approved.

PR #583, opened 2026-08-07
feat(settings): add keepAwake toggle to persist wake lock (gh-572)

### LINKED ISSUE #572: Add toggle to persist wakelock while the app is running
Samsung tablets no longer include the 'Never' option for screen lock timeout. 
This locks the tablets screen when in `sleep` state as skins usually release the wake lock at that point.

tadelv:
## Summary

- Problem: Samsung tablets no longer offer a "Never" screen-lock timeout. Skins release the wake lock when the machine enters `sleep`, so the tablet screen locks — the app appears dead until the tablet is woken manually.
- Why it matters: The Decent app runs unattended on tablet-based setups; a locked screen on a sleep-scheduled machine (e.g. overnight) hides the UI and can block scheduled wake/charging flows.
- What changed: Added a `keepAwake` setting (default **true**) that forces the wake lock on while the app runs, as a full override of the auto-managed wake-lock logic. Toggle lives in Settings > Power ("Keep screen awake") and in the bundled `settings.reaplugin` (bumped to 0.1.6) via `POST /api/v1/settings`. `DisplayController` re-evaluates the wake lock immediately when the setting changes. The setting is also included in backup export/import.
- What did NOT change (scope boundary): Auto-managed wake-lock behavior (connected + not sleeping) remains the fallback when `keepAwake` is off; skin wake-lock override semantics unchanged; brightness logic untouched; no new endpoints (existing settings GET/POST extended).

## Change Type (select all)

- [x] Feature
- [x] Docs
- [x] Plugin (DYE2 or bundled skin) — `settings.reaplugin` bundled plugin

- [x] REST API / handlers
- [x] Plugins / JS runtime
- [x] UI / Flutter widgets
- [x] Docs / specs

## Linked Issues

- Closes #572
- Related #

## Root Cause (if bug fix)

For bugs: explain why this happened, not just what changed. Otherwise write `N/A`.

- Root cause: Hardware/OS change on Samsung tablets (removal of "Never" screen timeout) combined with the app's auto-managed wake-lock policy, which releases the lock when the machine sleeps. Nothing prevents the OS screen timeout from firing after the release.
- Missing detection or guardrail: No user-facing way to persist the wake lock independent of machine state.
- Contributing context (if known): Issue #572.

For bugs and refactors: name the smallest reliable test that should catch this. Otherwise write `N/A`.

  - [x] Unit test
- Target test or file: `test/controllers/display_controller_test.dart` — `keepAwake setting` group; `test/data_export/settings_export_section_test.dart` round-trip.
- Scenario the test should lock in: with `keepAwake=true`, wake lock stays enabled through machine sleep and disconnect; toggling it off while sleeping releases the lock; disabling the auto-manage tests explicitly (they now need `keepAwake=false` since the default flipped to true).
- If no new test added, why not: N/A — tests added.

These are **not optional**. Stale specs mislead clients and agents.

- [ ] Plugin docs updated: `doc/Plugins.md` (if events/API changed) — N/A: no new plugin events/API, settings page UI only
- [ ] Skin docs updated: `doc/Skins.md` (if skin behavior changed) — N/A: no skin behavior change
- [ ] Profile docs updated: `doc/Profiles.md` (if profile handling changed) — N/A
- [ ] Device docs updated: `doc/DeviceManagement.md` (if device flows changed) — N/A
- [ ] N/A — no docs affected

## Security Impact (required)

This app runs a local web server (port 8080) and connects to BLE/USB hardware.

- New or changed REST endpoints? (`Yes/No`) — No; existing `GET|POST /api/v1/settings` extended with one key
- New or changed WebSocket topics? (`Yes/No`) — No
- New or changed network calls? (`Yes/No`) — No
- BLE/USB surface changed? (`Yes/No`) — No
- File system access changed? (`Yes/No`) — No
- Plugin sandbox boundary changed? (`Yes/No`) — No
- If any `Yes`, explain risk and mitigation: N/A

## User-Visible Changes

- New "Keep screen awake" switch in Settings > Power (default on): prevents the tablet screen from turning off while the app is running.
- Same toggle in the bundled Settings plugin page (`settings.reaplugin` 0.1.6).
- `keepAwake` (and `lowBatteryBrightnessLimit`) now round-trip through backup export/import.
- Note: with the default `true`, the wake lock is now held while the app runs even when the machine sleeps or disconnects — battery impact on tablets, which is the intent of the setting.

- [x] `dart format lib test` — no remaining changes
- [x] `flutter test` — all pass (2818 tests)
- [ ] `./scripts/fetch_dye2_plugin.sh` — DYE2 plugin installed into `assets/plugins/dye2.reaplugin/` — N/A: no DYE2 change

- OS / platform tested: None (unit-tested only)
- What you personally verified and how: N/A
- Edge cases checked: N/A
- What you did **not** verify: N/A

### Evidence

Attach at least one:

- [x] Test output (failing before + passing after) — full `flutter test` run: 2818 tests passed; new `keepAwake setting` group in `display_controller_test.dart` and round-trip coverage in `settings_export_section_test.dart`

## Compatibility & Migration

- Backward compatible? (`Yes/No`) — Yes: new optional settings key; absent in requests/backups falls back to defaults
- Config / env changes needed? (`Yes/No`) — No
- Database migration needed? (`Yes/No`) — No

## Risks & Mitigations

List only real risks for this PR. If none, write `None`.

- Risk: Default `keepAwake=true` changes wake-lock behavior for existing users (screen stays on during machine sleep/disconnect, battery drain on tablets).
  - Mitigation: This is the requested default per #572; users can disable it in Settings > Power; skin wake-lock override still works on top.


ODevStudio (commented):
Overall this looks coherent and CI is green. I found two minor issues worth addressing:

1. **`keepAwake=true` is not enforced by `DisplayController.initialize()` itself.** `initialize()` only installs listeners. If settings have already been loaded before the display controller is initialized, the DE1 stream's initial `null` does not trigger `_evaluateWakeLock()` because `_de1` is already `null`, so the controller starts with the wake lock disabled despite `keepAwake == true`. The current startup ordering masks this because `loadSettings()` happens after `displayController.initialize()` and its `notifyListeners()` causes reevaluation. I suggest calling `unawaited(_evaluateWakeLock())` from `initialize()` and adding a test for initializing while disconnected with an already-loaded/default `keepAwake=true` setting.

2. **The OpenAPI request schema allows `keepAwake: null`, but the handler rejects it.** `ReaSettingsRequest.keepAwake` is declared `nullable: true`, while `SettingsHandler` accepts only a `bool` and returns 400 for `null`. Since omission already means "leave unchanged", I suggest removing `nullable: true` from this field.

Non-blocking product note: this defaults `keepAwake` to `true` globally, so existing desktop installs also switch from machine-state-managed wake-lock behavior to keeping the display awake while the app runs. The PR description calls this out, but it may be worth confirming that the global default-on behavior is intentional given that the motivating issue is specifically about Samsung tablets.

ODevStudio (approved):
Approved. The two review findings I left are minor and non-blocking; they can be addressed in this PR or as follow-up cleanup.

PR #584, opened 2026-08-07
fix(plugins): forward recipe and shot tags to Visualizer upload

bsnelson:
## Summary

- Problem: recipe, shot-review, and Visualizer-only tags could be lost during upload races, retries, and concurrent forward or back synchronization.
- Why it matters: lost tags damage a user's tasting history, while a timeout after a successful upload can prompt a retry that creates a duplicate Visualizer shot.
- What changed: uploads now return their Visualizer ID after persisting the mapping and pending sync, then refresh the local shot in the background. Tag ownership is stored with the pending PATCH before the request starts. Forward writes no longer advance the global back-sync cursor, stale successful requests discard fields they already applied, and legacy `metadata` edits keep their documented raw patch while Visualizer reads tags from the canonical shot.
- Scope boundary: no BLE/USB behavior, database schema, WebSocket contract, or core REST request/response schema changed.

## Change Type (select all)

- [x] Bug fix
- [ ] Feature
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [x] REST API / handlers
- [x] Plugins / JS runtime
- [x] Docs / specs

## Linked Issues

- Closes: N/A
- Related: N/A

## Root Cause (if bug fix)

- Visualizer's upload importer does not persist `app.data.settings.tags`, so tags require a supported post-upload PATCH.
- The upload handler waited for follow-up network requests under the same 30-second outer deadline, so it could hide a completed upload ID from the caller.
- Tag ownership was recorded only after a successful response, which misclassified a committed PATCH when its response was lost.
- Forward PATCHes advanced the global back-sync cursor past unrelated remote changes.
- Queued revisions retained fields that an older successful request had already applied.
- Visualizer only recognized canonical annotation patches, while the documented `shotUpdated.patch` value remains the raw PUT body.

- [x] Unit tests in `test/plugins/visualizer_plugin_test.dart` and `test/plugins/visualizer_plugin_sync_cases.dart`
- [x] Handler integration coverage in `test/webserver/shots_handler_test.dart`
- A stalled follow-up request cannot delay the `202` response containing `visualizer_id`.
- A committed PATCH with a lost response retains ownership through retry and later removal.
- A forward update at timestamp 120 leaves an unprocessed back-sync change at timestamp 110 visible from cursor 100.
- A stale successful notes PATCH does not carry notes into a newer tag-only PATCH.
- Legacy `metadata.tags` and `metadata: null` updates retain the raw event patch while Visualizer forwards canonical tags.
- Existing upload-race, remote-only tag, rapid-edit, retry, deduplication, and ancestor-clear cases remain covered.

- [ ] API spec updated: no REST or WebSocket request/response schema changed.
- [ ] API docs updated: no user-facing core endpoint contract changed.
- [x] Plugin docs updated: `doc/Plugins.md` documents immediate `202` upload responses and background follow-up sync.
- [ ] Skin docs updated: no skin behavior changed.
- [ ] Profile docs updated: no profile handling changed.
- [ ] Device docs updated: no device flow changed.

## Security Impact (required)

- New or changed network calls? Yes
- BLE/USB surface changed? No
- Risk and mitigation: the bundled plugin performs authenticated GET and PATCH requests against the existing Visualizer shot endpoints. Requests remain scoped to configured credentials and mapped shots.

## User-Visible Changes

Completed uploads return their Visualizer ID before tag sync finishes. Tags entered during upload remain queued, Visualizer-only tags survive local edits, local removals still propagate, and back-sync does not skip unrelated remote changes.

### Local gates

- [x] `dart format lib test`: 688 files, 0 changed
- [x] `flutter analyze --no-pub`: no issues
- [x] `flutter test test/plugins/visualizer_plugin_test.dart test/webserver/shots_handler_test.dart --no-pub`: 26 passed
- [x] `node --check assets/plugins/visualizer.reaplugin/plugin.js`
- [x] `git diff --check`

### Full-suite note

The Windows full-suite run passed 2,822 tests and hit the same eight pre-existing `origin/main` PluginLoaderService path-separator failures. Their independent fix is green in #592. GitHub's PR test job runs on Ubuntu.

### Manual verification

- OS / platform: Windows unit-test harness
- Simulated devices: No
- Real hardware: No
- Live Visualizer account: Not exercised

### Evidence

- [x] Tests failed on all five new regressions before the fix and passed afterward.

## Compatibility & Migration

- Backward compatible? Yes
- Bundled plugin version: 1.5.5

## Risks & Mitigations

- Risk: Visualizer state may change between the fresh GET and PATCH because the API does not expose a conditional tag update.
- Mitigation: the plugin reads immediately before PATCH, serializes pending local revisions, preserves remote-only tags, and persists retry state.

### DIFF
--- a/assets/plugins/visualizer.reaplugin/plugin.js
+++ b/assets/plugins/visualizer.reaplugin/plugin.js
@@ -589,6 +589,9 @@ function createPlugin(host) {
     });
     if (!res.ok) {
       const text = await res.text();
+      if ((res.status === 400 || res.status === 403) && hasOwn(body?.shot, "tag_list")) {
+        throw new Error(`Visualizer tag sync requires Visualizer Premium. HTTP ${res.status}: ${text || res.statusText}`);
+      }
       throw new Error(`HTTP ${res.status}: ${text || res.statusText}`);
     }
     return await res.json();
@@ -757,7 +760,10 @@ function createPlugin(host) {
             [visualizerId]: merged.managedTags
           };
         }
-        const detail = await visualizerPatch(`/shots/${visualizerId}`, { shot: merged.update });
+        const visualizerUpdate = hasOwn(merged.update, "tags")
+          ? { ...withoutKey(merged.update, "tags"), tag_list: merged.update.tags }
+          : merged.update;
+        const detail = await visualizerPatch(`/shots/${visualizerId}`, { shot: visualizerUpdate });
         const updatedAt = Number(detail?.updated_at) || Number(detail?.meta?.visualizer?.updated_at) || 0;
         if (updatedAt > 0) {
           state.backSyncState = {
--- a/assets/plugins/visualizer.reaplugin/plugin.js
+++ b/assets/plugins/visualizer.reaplugin/plugin.js
@@ -178,6 +178,20 @@ function createPlugin(host) {
     return Array.from(new Set(tags.map((tag) => String(tag).trim()).filter((tag) => tag.length > 0)));
   }
 
+  function canonicalizeVisualizerTag(tag) {
+    return String(tag).replace(/\s+/g, " ").trim().replace(/[^\w\s-]/g, "").toLowerCase();
+  }
+
+  function canonicalizeVisualizerTags(tags) {
+    if (!Array.isArray(tags)) return [];
+    return Array.from(new Set(
+      tags
+        .flatMap((tag) => String(tag).split(","))
+        .map(canonicalizeVisualizerTag)
+        .filter((tag) => tag.length > 0)
+    ));
+  }
+
   function extractTags(shot) {
     const annotations = shot?.annotations || {};
     const context = shot?.workflow?.context || {};
@@ -740,12 +754,14 @@ function createPlugin(host) {
     }
     const detail = await visualizerGet(`/shots/${visualizerId}?essentials=1`);
     const localTags = normalizeTags(pending.update.tags);
-    const managedTags = new Set(normalizeTags(state.managedLocalTags[visualizerId]));
-    const remoteOnlyTags = normalizeTags(detail?.tags).filter((tag) => !managedTags.has(tag));
-    const remoteOnlySet = new Set(remoteOnlyTags);
+    const managedTags = new Set(canonicalizeVisualizerTags(state.managedLocalTags[visualizerId]));
+    const remoteOnlyTags = normalizeTags(detail?.tags).filter(
+      (tag) => !managedTags.has(canonicalizeVisualizerTag(tag))
+    );
+    const remoteOnlySet = new Set(canonicalizeVisualizerTags(remoteOnlyTags));
     return {
       update: { ...pending.update, tags: normalizeTags([...localTags, ...remoteOnlyTags]) },
-      managedTags: localTags.filter((tag) => !remoteOnlySet.has(tag))
+      managedTags: canonicalizeVisualizerTags(localTags).filter((tag) => !remoteOnlySet.has(tag))
     };
   }
 
@@ -1138,7 +1154,7 @@ function createPlugin(host) {
   // Return the plugin object
   return {
     id: "visualizer.reaplugin",
-    version: "1.5.4",
+    version: "1.5.5",
 
     onLoad(settings) {
       state.username = settings.Username;
--- a/assets/plugins/visualizer.reaplugin/plugin.js
+++ b/assets/plugins/visualizer.reaplugin/plugin.js
@@ -45,6 +45,10 @@ function createPlugin(host) {
     localSyncStatus: { lastCheck: null, lastResult: null, lastError: null, lastShotId: null, lastVisualizerId: null },
     backSyncStatus: { lastCheck: null, lastResult: null, lastError: null, lastApplied: 0 },
   };
+  let managedLocalTagsReadyResolve;
+  const managedLocalTagsReady = new Promise((resolve) => {
+    managedLocalTagsReadyResolve = resolve;
+  });
 
   function log(msg) {
     host.log(`[visualizer] ${msg}`);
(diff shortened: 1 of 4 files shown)


ODevStudio (changes requested):
Thanks for working on this — the direction makes sense, and the added test is useful. I found a few issues that I think need addressing before this is ready to merge:

1. **Visualizer currently won’t persist `app.data.settings.tags`.** The Decent JSON importer only captures a fixed set of fields, and `tags` isn’t one of them. So the new test confirms that Decaid sends the tags, but not that they actually make it into the Visualizer shot. This likely needs either a corresponding Visualizer-side change or a post-upload update using the supported tag field/API.

2. **Tags added during shot review after the initial upload still won’t sync.** The existing `shotUpdated` → `pushLocalShotUpdate()` path doesn’t include tags in `localShotToVisualizerUpdate()`. Since the upload happens shortly after the shot ends, tags added later in review would still be missed. It would be good to extend that update path and add a regression test for the post-upload case.

3. **The Visualizer plugin version should be bumped.** The bundled plugin updater only replaces an installed plugin when the bundled manifest version is newer. Since this PR changes `plugin.js` but leaves the version at `1.5.2`, existing installs would keep the old implementation.

4. **CI is currently failing formatting.** `dart format` reports changes needed in `test/plugins/visualizer_plugin_test.dart`.

One smaller test suggestion: the deduplication behavior would be better covered by putting the same tag in both tag sources and asserting it only appears once.

Once the Visualizer persistence/update path and plugin versioning are handled, this should be in much better shape.

bsnelson:
Thanks for the thorough review — all four points addressed in bde21b1c:

1. **Tags now PATCH, not import.** Moved tag delivery out of `app.data.settings` (which the JSON importer ignores) into a follow-up `PATCH /shots/<visualizerId>` using the flat `tags` field, mirroring the shape Visualizer already returns on back-sync reads. New `pushTagsToVisualizer()` helper, called after both the auto-upload path and the manual `upload` HTTP endpoint.
2. **Post-upload tag edits now sync.** `localShotToVisualizerUpdate()` forwards `tags` whenever the incoming patch touches `annotations.extras.tags` or `workflow.context.extras.tags` (or on a forced full sync), so tags added later in shot review reach Visualizer via the existing `shotUpdated` → `pushLocalShotUpdate()` path.
3. **Version bumped** 1.5.2 → 1.5.3 in `manifest.json` so the bundled-plugin updater overwrites existing installs.
4. **Formatting fixed** — `dart format` was failing on the test file; ran it and confirmed `dart format --set-exit-if-changed` is now clean.

Also took the dedupe suggestion: the upload test now puts an overlapping tag ("bright") in both the recipe and shot-review sources and asserts it appears only once in the PATCH payload. Added a second test covering the post-upload forward-sync case specifically.

`flutter analyze` clean, `flutter test` all 2501 passing, `dart format --set-exit-if-changed` clean.

assets/plugins/visualizer.reaplugin/plugin.js
```diff
@@ -659,20 +731,124 @@ function createPlugin(host) {
...
+      const pending = state.localSyncPending[visualizerId];
+      try {
+        const merged = await mergeCurrentVisualizerTags(visualizerId, pending);
+        if (merged.managedTags !== null) {
+          state.managedLocalTags = {
+            ...state.managedLocalTags,
+            [visualizerId]: merged.managedTags
+          };
+        }
+        const detail = await visualizerPatch(`/shots/${visualizerId}`, { shot: merged.update });
```
tadelv: [P1] Translate the internal tag update to Visualizer's write-side `tag_list` field before sending this PATCH. Current Visualizer returns `tags` on reads, but `Shots::Editing#update_shot_params` permits `tag_list` for writes (and its API tests update tags with `{ shot: { tag_list: [...] } }`). As written, this sends `{ shot: { tags: [...] } }`, so the live endpoint will not apply the tag change even though our mocks report success. Keeping `tags` as the internal queue key is fine; convert it at this boundary and make the tests assert the real request shape. Please also handle/surface the upstream premium-only restriction on tag writes rather than assuming all authenticated accounts can update them.

tadelv (changes requested):
Blocking on one Visualizer API contract mismatch. The new sync queue uses the read-side `tags` key as a write parameter, but current Visualizer accepts `tag_list` for tag updates. The tests currently stub the unsupported request shape, so they can pass while the live service does not apply the tags. Please translate to the write-side field at the API boundary, update the tests to assert the real request shape, and account for the fact that Visualizer currently gates tag writes to premium users.

tadelv:
One remaining blocker from the re-review: the new tag ownership tracking compares tag strings exactly, but Visualizer canonicalizes `tag_list` values when it stores them (squishes whitespace, strips unsupported characters, lowercases, then deduplicates).

That means a local tag such as `High-Speed` is recorded in `managedLocalTags` as `High-Speed`, while Visualizer subsequently returns it as `high-speed`. If the user then removes the local tag, the next merge sees `high-speed` as a remote-only tag because it no longer matches the managed spelling, so the plugin preserves/re-adds it instead of removing it.

Please make ownership comparisons use the same canonical representation Visualizer uses for tags, and add a regression test that writes a mixed-case/non-canonical local tag, has the mocked Visualizer return its canonicalized form, then verifies that removing the local tag actually removes it remotely. The existing lowercase-only tests do not exercise this behavior.

tadelv:
Two findings from the latest re-review:

1. **P1 — persist `managedLocalTags` across restart.** The new canonicalization fix is correct, but ownership is still memory-only. `managedLocalTags` starts empty on load and is not restored from plugin storage. After a normal app/plugin restart, a tag that Decaid previously wrote to Visualizer is therefore indistinguishable from a Visualizer-only tag. If the user removes that tag locally after restart, `mergeCurrentVisualizerTags()` sees the remote copy as `remoteOnlyTags` and preserves/re-adds it instead of removing it.

   Example: sync local `washed` → restart → Visualizer still has `washed` → remove it locally → ownership map is empty, so `washed` is classified as remote-only and survives. Please persist the canonicalized ownership map alongside the existing back-sync state, restore it in `handleStorageRead`/`onLoad`, and add a regression test that unloads/reloads the plugin between initial sync and removal.

2. **Minor — PR description has a stale plugin version.** The latest code/manifest bump the bundled Visualizer plugin to `1.5.5`, but the PR body still says `Bundled plugin version: 1.5.4`. Please update that line for consistency.

bsnelson:
All outstanding review points are addressed as of ff33d11a (tag canonicalization, ownership persisted across restart, `tag_list` write-side translation, premium-gate handling). Tests, `flutter analyze`, and `dart format` are clean. Ready for re-review.

tadelv (approved):

PR #586, opened 2026-08-07
Simulator: replay device that plays a real shot matched to the selected profile

johnbuckman:
> ⚠️ **Vibe-coded, needs review.** Built with Claude Code at John's direction. Functional and fully tested, but treat the design as a proposal.

Reworked per @tadelv's review: replay is now its **own device**, not baked into `MockDe1`. Rebased onto `main` (the earlier in-`MockDe1` approach and the plugin-toggle commit are gone), so `MockDe1`/`MockScale` are untouched and keep the recent simulator-realism work.

## Summary

- **Problem:** the simulated DE1 synthesizes telemetry — never looks like a real pull, identical every run.
- **What changed:** new `simulate=replay` devices **`MockReplayDe1` + `MockReplayScale`** that play back a real recorded shot. On shot start `MockReplayDe1` asks `SimulatedShotLibrary` for a bundled recording **made with the currently selected profile** (normalized-title match); if none exists it falls back to a generic shot. It streams the recording at 10 Hz; `MockReplayScale` reports the recording's **real weight** (no flow integration).
- **SRP:** `MockReplayDe1 extends MockDe1` only to inherit the large `De1Interface` stub surface; it runs its own connection/stream/timer and never uses the puck sim. `MockDe1`/`MockScale` remain the pure puck simulator.
- **What did NOT change:** `MockDe1`/`MockScale`, real device paths, REST/WS API, storage.

## The corpus (visualizer.coffee)

Profile-matched shots were pulled from the public visualizer.coffee firehose — one real shot per bundled profile that had a public match (**14 of 71** at build time: the popular Decent defaults; the long tail has no public shots → generic fallback). Each was reconstructed into a `.shot` and run through the existing `TclShotParser` → 10 Hz pipeline. `assets/simulations/manifest.json` maps profile title → shot.

## Change Type

- [x] Feature
- [x] Docs

## Scope

- [x] Scale / weight / flow
- [x] Docs / specs

## Linked Issues

- Related: John Todo 10181321154; supersedes this PR's original in-`MockDe1` approach per review.

## Documentation Obligations

- [x] N/A for API/skin/profile/device-flow specs. Internal notes updated: `doc/AI_BUILD_NOTES.md`, `doc/AI_TESTING_NOTES.md`; design doc at `doc/plans/archive/simulator-shot-replay/design.md`.

## Security Impact

- REST/WS/network/BLE/USB/filesystem/plugin-sandbox: **No** on all (simulated devices only; reads a bundled asset via `rootBundle`).

## User-Visible Changes

- In `simulate=replay`, starting an espresso replays a real recorded shot matched to the selected profile (or a generic fallback) at 10 Hz, with the recording's real weight. No change with real hardware or the other simulate modes.

## Verification

- [x] `dart format lib test` clean · [x] `flutter analyze` clean · [x] `flutter test` — **all 2888 pass**
- New tests: `ShotReplayer` mapping; `SimulatedShotLibrary` profile match + normalization + fallback; `MockReplayDe1` picks the profile-matched shot, falls back when unmatched, and the scale reports recorded weight; asset guard/regenerator.
- Manual: not yet re-run on macOS after the device split (previous versions were). Will attach a screenshot.

## Compatibility & Migration

- Backward compatible? **Yes.** No config/env/DB migration.

## Risks & Mitigations / rework spots (please scrutinize)

- `MockReplayDe1 extends MockDe1` for interface reuse but never uses the puck code — a shared `SimulatedDe1Base` would be cleaner but means refactoring `MockDe1`.
- Injected single "Replay" profile step (so the JSON round-trips) is a placeholder; real advanced_shot frames not reconstructed.
- Large committed corpus (~17 shot JSONs + `.shot` sources).
- Coverage is 14/71 profiles; the rest replay a generic fallback.


tadelv:
Closing this because we want to keep Decaid's simulator profile-driven rather than replace it with historical-shot replay.

The simulator is useful not just for producing realistic-looking telemetry, but for exercising the selected profile and the normal Decaid shot-control path: profile steps and transitions, limiters, target-weight behavior, MockScale/ShotSequencer interaction, and other profile-dependent behavior. Replaying a fixed historical shot would largely bypass those semantics, so although the resulting telemetry is more realistic, it makes `simulate=1` less representative of how Decaid itself behaves for a given profile.

The direction in #590 is a better fit for what we want: retain the synthetic/profile-driven model and improve its realism, including honoring profile limiters and introducing bounded puck variation.

Thanks for putting this together — the replay approach is useful in its own right, but it isn't the direction we want for the default Decaid simulator.

johnbuckman:
Reworked per your review, @tadelv: replay is now its own device pair (`MockReplayDe1` + `MockReplayScale`, `simulate=replay`) instead of being baked into `MockDe1` — so `MockDe1`/`MockScale` stay the single-responsibility puck simulator and keep the recent realism work. I rebased onto `main` and force-pushed, dropping the old in-`MockDe1` approach and the plugin-toggle commit.

New capability on top of your suggestion: the replay device picks a bundled recording **made with the currently selected profile** when one exists (pulled from visualizer.coffee — 14 of the 71 bundled profiles had a public match), else a generic fallback. `MockReplayDe1` reuses `MockDe1` only for the `De1Interface` stub surface; it runs its own stream/timer and never touches the puck sim.

Still vibe-coded — the extends-for-reuse seam and the placeholder profile step are the spots I'd most want your eyes on. All 2888 tests pass.

johnbuckman:
Built @ODevStudio's #590 goals into the replay path.

#590 (merged) argued replay 'discards the selected profile and target-weight dynamics.' Replay now answers both:
- **Selected profile** — it already plays a recording made with the current profile.
- **Target weight** — the recorded weight flows through the normal `ScaleController`/`ShotSequencer`, so a replayed shot now **stops at the profile's target yield via the exact stop-at-weight path a real shot uses** — no replay-specific stop logic (respecting #590's 'no duplicate stop-at-weight' boundary). The only change needed: recordings start at the pour, so `MockReplayDe1` first emits a brief `preparingForShot` frame to kick off the sequencer's lifecycle. Locked in by `test/integration/replay_target_weight_test.dart`.

#590's per-shot puck-resistance jitter is inherently synthetic and doesn't apply to replaying real recordings.

Full suite green (2889). Note: #590's realism work is on `main` and untouched by this branch.

johnbuckman:
@vid please reopen this and let's discuss.  It's a genuinely useful feature that I'd like to have optional in decaid, but which is acceptable to your requirements.  So, if it's still not acceptable, let's discuss.

tadelv:
@johnbuckman instead of creating a device pair, a single device should be used, just like a Bengle, which has Scale and steam temp and other mixins

tadelv:
the current approach is still incorrect, either a commit was not pushed or is wrongly implemented. ShotReplayer should be a stand alone device, not injected into MockDe1

tadelv:
@johnbuckman  also please open a new PR, I seem to be unable to reopen this one

PR #605, opened 2026-08-12
fix: align Bengle MMR units with firmware

ODevStudio:
## Summary

- Problem: Decaid encoded Bengle `MatSetPoint` in tenths, while firmware consumes whole degrees; the REST contract still accepted unrepresentable fractions.
- Why it matters: a 60 C target was sent as raw 600, and fractional API values would be silently truncated after correcting the wire scale.
- What changed: use the firmware's x1 units, expose integer cup-warmer values, reject fractional or out-of-range input, and document the 0..80 C range with 1 C resolution.
- What did NOT change (scope boundary): byte-level cup-warmer wire regressions remain; the duplicated firmware fixture and contract test were removed.

## Change Type (select all)

- [x] Bug fix
- [ ] Feature
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [x] REST API / handlers
- [ ] WebSocket API
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

- Related #462
- Related #601

## Root Cause (if bug fix)

- Root cause: `MatSetPoint` copied the x10 convention used by other temperature registers even though this firmware register uses multiplier 1.
- Missing detection or guardrail: the public API did not validate values against the register's whole-degree resolution.
- Contributing context (if known): the shared scaled encoder truncates values that are not representable at the configured scale.

  - [x] Unit test
- Target test or file: `test/models/device/bengle_cup_warmer_test.dart` and `test/services/webserver/de1handler_cup_warmer_test.dart`
- Scenario the test should lock in: raw wire writes use whole degrees, GET returns an integer, and `{temperature: 60.5}` returns 400.
- If no new test added, why not: N/A

- [x] API spec updated: `assets/api/rest_v1.yml`
- [ ] N/A - no docs affected

## Security Impact (required)

- New or changed REST endpoints? (`Yes/No`): Yes, validation and schema for the existing cup-warmer endpoint.
- New or changed WebSocket topics? (`Yes/No`): No
- New or changed network calls? (`Yes/No`): No
- BLE/USB surface changed? (`Yes/No`): Yes, the existing Bengle cup-warmer write now uses x1 firmware units.
- File system access changed? (`Yes/No`): No
- Plugin sandbox boundary changed? (`Yes/No`): No
- If any `Yes`, explain risk and mitigation: validation rejects values the hardware cannot represent, range checks remain 0..80 C, and byte-level tests assert the raw write/read behavior.

## User-Visible Changes

The cup-warmer REST API now returns whole-degree integers and rejects fractional targets with HTTP 400.

- [x] `dart format lib test` - changed Dart files formatted
- [x] `flutter analyze` - no issues found
- [ ] `flutter test` - full suite completed with 2,791 passes and 116 unrelated Windows/CRLF and timing failures
- [ ] `./scripts/fetch_dye2_plugin.sh` - not applicable; plugin assets are unchanged

- OS / platform tested: Windows
- Simulated devices? (`simulate=1`): No
- What you personally verified and how: ran the focused REST and byte-level cup-warmer tests and parsed the updated OpenAPI document through the existing schema test.
- Edge cases checked: integer response, fractional rejection, 0..80 bounds, raw 60 write, raw 80 clamp.
- What you did **not** verify: live app or real Bengle hardware.

### Evidence

- [ ] Log snippets

Focused result: 15 tests passed. OpenAPI schema test passed. `flutter analyze` reported no issues.

## Compatibility & Migration

- Backward compatible? (`Yes/No`): No for clients sending fractional cup-warmer targets; those requests now return 400 instead of truncating.
- Config / env changes needed? (`Yes/No`): No
- Database migration needed? (`Yes/No`): No
- If any `No` or `Yes`, explain exact steps: clients must send integer targets from 0 through 80.

## Risks & Mitigations

- Risk: clients that send values such as 60.5 will receive 400.
  - Mitigation: OpenAPI and `doc/Api.md` now state integer type and 1 C resolution.


tadelv (changes requested):
The register fix itself is correct, but I’d like two changes before merge:

1. **Update the public cup-warmer API to match the firmware’s whole-degree resolution.** `MatSetPoint` is now correctly modeled with multiplier 1, so Decaid should no longer advertise arbitrary floating-point precision that the machine cannot represent. Please make the REST/API contract explicitly whole-degree only, including request validation and OpenAPI. For example, reject fractional values rather than silently truncating them, and add the corresponding API regression coverage. `0..80 °C` remains the valid range; the important change is the 1 °C step/resolution.

2. **Remove the firmware contract fixture/test from this PR entirely.** We do not want a duplicated/pinned firmware contract maintained in the Decaid source tree. Firmware behavior and register definitions will be derived from the actual firmware repositories/source when needed. Please remove `test/fixtures/bengle_mmr_contract.json` and `test/unit/models/device/impl/bengle/mmr_contract_test.dart` rather than revising or expanding them.

Keep the byte-level cup-warmer regression tests that directly exercise Decaid’s actual wire behavior; those are useful and should remain.

tadelv (approved):

PR #604, opened 2026-08-12
fix: round Bengle scaled MMR writes

ODevStudio:
## Summary

- Problem: protected capability MMR writes truncated scaled doubles, so a Bengle 2.3 g stop target could serialize as raw 229 instead of 230.
- Why it matters: firmware should receive the nearest representable scaled value; floating-point representation must not bias targets downward.
- What changed: round protected scaled writes, lock Bengle and private DE1 behavior with regressions, correct the capability description, and revise the MockBengle integrated-scale scenario for firmware-autonomous SAW.
- What did NOT change (scope boundary): no SAW range, readback, tare, or A013 work already delivered by #601; no public API shape or private DE1 serialization change.

## Change Type (select all)

- [x] Bug fix
- [ ] Feature
- [ ] Refactor required for the fix
- [x] Docs
- [ ] Security hardening
- [ ] Chore / infra
- [ ] Plugin (DYE2 or bundled skin)

## Scope (select all touched areas)

- [x] BLE transport / device comms
- [ ] REST API / handlers
- [ ] WebSocket API
- [ ] Machine state / shot logic
- [x] Scale / weight / flow
- [ ] Profiles / beans / grinders / workflows
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] Storage / Drift database
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

- Related #462
- Related #601

## Root Cause (if bug fix)

- Root cause: `(value * scale).toInt()` truncates binary floating-point products such as `2.3 * 100`, which can be represented just below 230.
- Missing detection or guardrail: no regression distinguished the protected capability writer from the legacy private DE1 writer.
- Contributing context (if known): #601 established the current Bengle SAW path, leaving only this serialization residual from #462.

## Regression Test Plan (if bug fix or refactor)

- Coverage level that should have caught this:
  - [x] Unit test
  - [ ] Integration test (mock transport edge)
  - [x] End-to-end test (`simulate=1` + curl/websocat)
  - [ ] Existing coverage already sufficient
- Target test or file: `test/unit/models/device/impl/de1/unified_de1/protected_surface_test.dart`
- Scenario the test should lock in: Bengle `endOfShotWeight = 2.3` writes raw 230 while plain DE1 `setSteamFlow(2.3)` remains raw 229.
- If no new test added, why not: N/A

## Documentation Obligations (required)

[...]

Issue #469, opened 2026-07-16
Bengle machine support (Decent-hardware variant) — overview + PR index


ChampionDesigns:
This is the umbrella overview for Bengle machine support in reaprime, now open as a foundation PR
plus a stack of small, severable, capability-gated PRs (indexed below), alongside a few independent
DE1 fixes. Posting per CONTRIBUTING to align on scope, sequencing, and the two open questions at the
end — happy to adjust scope, split further, restructure (including core vs plugin), or provide the
per-PR hardware-validation evidence.

## What Bengle is

Bengle is a machine built on Decent hardware that identifies itself over the standard `v13Model`
MMR with a value `>= 128`. It speaks the DE1 protocol as a strict superset, and adds:

- an **integrated scale** (load cells in the drip tray; weight rides a new BLE characteristic
  `0xA013` carrying an extended shot sample),
- **LED strips**, a **cup warmer** with a heated mat, and a **milk temperature probe**,
- a **USB-serial transport** as a full alternative to BLE (works with Bluetooth off).

Everything here is hardware-validated against a live machine running pinned firmware, with **one**
exception on the record: the cup-warmer **mat-temperature readout** is unit/mock-tested only
(severable). The load-cell **calibration wizard** (isolated-cell auto-detect) is hardware-validated
end-to-end — the empty-platform zero and both per-cell latches. Everything ships with a full
unit/integration test suite plus a mock device (`simulate=bengle`) so CI and reviewers need no
hardware.

## PR index

**Bengle capability stack** — foundation first, then additive capability PRs gated on Bengle
detection. They are a stack; **please review/merge in dependency order**. Each PR bases on the
previous branch, so until the earlier ones merge the later diffs are cumulative (noted in each PR).

1. #459 — B-1 · `feat(bengle):` model-authoritative machine class, MTU 517, and a firmware↔app MMR contract
2. #460 — B-2 · `feat(serial):` DE1-family USB-serial parity, and let a wired-only setup reach its machine
3. #461 — B-3 · `feat(bengle):` 0xA013 as the sole shot source, and report the firmware's own gravimetric flow
4. #462 — B-4 · `feat(bengle):` fill in the real EndOfShotWeight and ScaleTare registers
5. #463 — B-5 · `feat(bengle):` isolated-cell auto-detect load-cell calibration wizard
6. #464 — B-6 · `feat(bengle):` LED strip palettes + live preview over MMR
[...]

PR #612, opened 2026-08-13
feat: complete Bengle support for the current firmware (scale cal, LEDs, cup warmer + preheat, wake schedule)

tadelv:
## Summary

Completes Decaid's Bengle support for the current `BengleMainCPUFirmware`
snapshot (tadelv/Bengle `master` @ `2377c7e0e48e9ee2c43cf02ad2f82028252f56e8`),
replacing the remaining open historical Bengle PRs. Every feature was
re-derived from the current firmware (MMR.def rows 39-61 + the consuming
C/C++ in System.cpp / System.hpp / APIView.cpp / ShotMachine.cpp /
CLoadCellCal.hpp), never copied from the historical PRs. Decaid `main` base:
`bc56025ee4e454c9910dd876897ef674bea535a9`.

- **Scale calibration** (firmware rows 39-41): `zero` / `latch` (isolated-cell,
  auto-detect, fractional grams ×10 on the wire) / `abort`, full packed
  `ScaleCalState` decode (step, detected cell, sub-state, seconds remaining,
  E_CalStatus incl. NotIsolated), acceptance observed as the step leaving its
  terminal value (guards the firmware's latched-terminal race), single-flight,
  REST `GET`/`PUT /api/v1/machine/scaleCalibration` (202 accepted / 409
  rejected).
- **LED strips** (rows 43-48): real persisted palette implementation replacing
  the stub — hydration once per connection (failed hydration → 503, never
  fabricated black), write-through `PUT` (firmware persists immediately),
  `frontSwitch` derived from the front strip exactly like the firmware (black
  → product defaults; ignored on write — no independent switch register),
  `commit` = documented compatibility no-op, `reset` = truthful firmware
  reload.
- **Cup warmer manual mode + live temperature** (rows 50, 58): explicit
  `enabled` without destroying the persisted `MatSetPoint`;
  `{"temperature": 45}` keeps its historical meaning (set AND enable manual
  heating); RAM-only mode is never re-enabled on reconnect;
  `currentTemperature` nullable (raw 0 = no valid reading).
- **Scheduled pre-warm** (rows 59-61): firmware-owned timing, persisted
  `enabled`/`leadMinutes` (0-120), read-only `active`; the app stores nothing
  and never reasserts on connect. New `GET`/`PUT /cupWarmer/preheat`.
- **Firmware wake schedule + inactivity timeout** (rows 54-57): app
  `sleepTimeoutMinutes` pushed 1:1 (0..240, 0 = disabled — **no** #467
  60-minute floor); local wall-clock (DST-safe) + wake table pushed on every
  connect and on settings change (RAM-only); windows translated app-side
  (`keepAwakeFor=N` → `[start, start+N)`, otherwise `[start, start+240)` per
  product decision; Dart weekday → firmware dow; midnight splits; 32-entry
  cap); rewrite sequence is edge-safe against spurious wakes. No
  `ScheduleStatus` invention.
- **Firmware surface**: no probe and no per-endpoint gating — every
  Bengle is assumed to run the current firmware. There are no
  pre-`2377c7e0` Bengles in the field (decision 2026-08-15), so the
  old-firmware compat layer (empty capability list, 404s, schema
  nullability, mock flag) was removed. Bengle endpoints 404 only on
  non-Bengle machines; plain DE1s get zero Bengle-only writes.

## Linked Issue

N/A — supersedes the remaining historical Bengle PRs (see Historical
disposition below). Umbrella tracker: #469 (not touched).

## Historical disposition

- **#461** → superseded by #601 (A013 telemetry, integrated weight, GFlow,
  milk probe, `TargetMilkTemp`, autonomous steam stop, `EndOfShotWeight`,
  `ScaleTare`). No work here.
- **#462** → superseded by #601 + #604 (SAW, tare, protected scaled writes).
  Zero new implementation code.
- **#463** → superseded by this PR's current-firmware calibration.
  Deliberately dropped: explicit left/right cell commands (retired in
  firmware, cmds 4/5), whole-gram-only weight restriction (firmware is
  float), historical terminal-word constants (acceptance is observed via
  step transitions), second tare API (existing integrated-scale tare path).
- **#464** → superseded by this PR's LED implementation. Deliberately
  dropped/redesigned: live-register preview endpoints (live registers are
  persisted snapshots clobbered by state changes — not a config surface),
  fake independent `frontSwitch` control (derived, not a register),
  commit/reset as persist/rollback (no commit latch; reset = truthful
  reload).
- **#465** → `TargetMilkTemp` via #601 + `MatSetPoint` via #605 + this PR's
  manual mode / live temperature / preheat work. No auto-re-enable on
  reconnect (mode is RAM-only and boots off by design).
- **#466** → superseded by #486 (Bengle v2 profile upload). No work here.
- **#467** → superseded by this PR's schedule sync. Removed: `ScheduleStatus`
  invention (no such MMR), forced 60-minute floor (firmware 0 = disabled,
  matching Decaid), arbitrary 30-minute wake window (240-min firmware-max
  fallback per product decision). Safety note: `InactivitySleepTimeout` is
  persisted — the app's 0..240 value now survives tablet disconnects.
- **#468** → superseded by this PR's pre-warm. Removed: contract v2,
  additive pin, two-firmware-source contract model,
  `assets/api/bengle_hw_v1.yml`. The registers live in the same coherent MMR
  table as the rest of the current surface.

## Second review disposition (merge blockers fixed)

| Finding | Fix commit |
| --- | --- |
| High: schedule sync never retries a failed push on its own (firmware can keep the old/partially rewritten table, incl. the master-disable path) | `4c8656e3` — the drain now auto-retries the newest desired generation up to 3 attempts with a 2 s backoff; a newer generation during backoff supersedes (stale state never pushed alone); dirty state persists past the bound until the next trigger. Controller tests: failed connect push converges with no further changes; failed master-disable converges to timeout 0 + cleared table; supersede-during-backoff |
| Low/Med: OpenAPI still listed `ledStrip` as an unconditional Bengle capability | `4c8656e3` — `ledStrip` moved into the post-0x00803880 surface-gated group in `rest_v1.yml` and `doc/Api.md` |

## Old-firmware compat layer removal (2026-08-15)

Decision (see `doc/plans/archive/bengle-old-firmware-removal.md`): no
old-firmware Bengles exist, so the compatibility probe and the entire
outdated-firmware path were deleted. Every Bengle advertises the
complete capability set; `supportsCurrentBengleFirmwareSurface` and
`extra.bengleFirmwareSurface` are gone from the interface, mocks,
controllers, handlers and specs.
## Style cleanup (AGENTS.md: no explanatory comments in source)

Per AGENTS.md ("Do not add explanatory comments to new or substantially
rewritten code; put rationale, hardware constraints, and debugging history
in the matching `doc/AI_*_NOTES.md` file"), the comment essays added by
this branch were stripped from code and tests and consolidated into a new
`doc/AI_BENGLE_NOTES.md` (single-surface contract + probe semantics, MMR
rows 39-61 map with addresses/wire formats, scale calibration, LED
palette, cup warmer, wake-schedule translation + rewrite order,
presence-sync design). `doc/AI_REPO_MAP.md` routes Bengle work to it.
Code retains at most a one-line pointer to the notes file.

## Verification

- `dart format lib test` clean; `flutter analyze` clean.
- Full `flutter test`: 2998 passed, 1 pre-existing skip, 0 failures.
- Byte-exact FakeBleTransport tests for every new MMR address/payload
  (cal state packing, ×10 weight, palette RGB 8-bit encoding, mode/lead
  writes, clock/clear/entries/enable sequence).
- State-machine tests: cal acceptance/rejection, stale-terminal guard,
  single-flight, hydration failure (no fabricated black), old-firmware
  probe latch, plain-DE1 negative tests (404s, no writes).
- Schedule translation tests: weekday mapping, midnight splits, 240-min
  fallback, 32-entry cap, DST-safe clock.
- sb-dev smoke tests against MockBengle (curl): capabilities, cup-warmer
  setpoint/enable/temperature, preheat, LED hydrated read/write/commit/
  reset, scale-calibration zero/latch/abort + 400 validation — all as
  specified in the updated decent-app scenarios.
- **Hardware: no real Bengle hardware available.** Physical smoke tests
  (cal zero + latch A/B, LED palettes, manual warmer + mat temp,
  tablet-absent schedule wake, scheduled pre-warm) were NOT performed;
  byte-level wire tests + simulation are the only verification. This is
  stated explicitly rather than implied.

## Impact

- New REST surface: `GET`/`PUT /api/v1/machine/scaleCalibration`,
  `GET`/`PUT /api/v1/machine/cupWarmer/preheat`; `GET /cupWarmer` gains
  `enabled` + `currentTemperature` (nullable; null on old firmware);
  `PUT /cupWarmer` accepts `enabled` (temperature-only requests unchanged);
  `/ledStrip` semantics are now truthful (write-through, no commit latch,
  derived switch, 503 on failed hydration).
- Capabilities list: the full 7-item set for every Bengle
  (`cupWarmer`, `integratedScale`, `stopAtWeight`, `ledStrip`,
  `scaleCalibration`, `preheat`, `wakeSchedule`), empty for plain DE1s.
- `assets/api/rest_v1.yml` and `doc/Api.md` updated in the same commits;
  `doc/DeviceManagement.md` documents the firmware-synced state flow;
  decent-app scenarios updated/added.
- Design doc archived at `doc/plans/archive/bengle-feature-completion/`.

## Review disposition (2026-08-14 review)

All findings addressed on the current head (68c2982e); each fix is one
logical commit:

| Finding | Fix commit |
| --- | --- |
| H1 master toggle not mirrored to firmware | `cdf35631` (+`64c92ac3`) — master flag is part of the desired state: disabling pushes timeout 0 + cleared table; re-enabling restores |
| H2 `{temperature, enabled:false}` intermediate enable | `363c0849` — auto-enable only when `enabled` is omitted; byte-order + injected-failure tests |
| H3 unserialized / non-convergent schedule sync | `cdf35631` — device-scoped coalescing single-flight drain; cache commits only on full success; dirty state retried; mid-flight replacement cannot mark the new machine synced |
| M1 clock written before old table disabled | `cdf35631` — rewrite now disables first (control 0 -> clock -> entries -> control 1); byte-exact order test |
| M2 preheat enable before lead | `363c0849` — lead first when enabling, disable first when disabling |
| M3 LED cache/reset truthfulness | `7c772565` — quantized cache, unknown state on partial write, failed reset returns null -> 503; MockBengle mirrors 8-bit quantization |
| M4 ledStrip advertised on old firmware | `7c772565` — capability gated on the surface probe like the other new features |
| M5 one dropped probe response disables surface | `da1b89cf` — bounded two-attempt probe |
| M6 DST claim only true at push time | `bd20f26f` — docs narrowed to "safe while synchronized" (no tz-watcher; noted as operational limitation) |
| L1 swapped 404 feature labels | `bc96ce12` — labels swapped; body assertions in old-firmware route tests |

Fault-injection and concurrency tests added for every multi-register
transaction (MMR write-failure injection at each sequence point, overlap/
coalescing, failure convergence, device replacement mid-push). Full
`flutter test` (2,998+), analyzer, formatter and the CI 20 s active-time
gate all pass. Physical smoke tests remain outstanding (no Bengle
hardware available).

## Contributor Responsibility


ODevStudio (changes requested):
PR #612

Verdict: Request changes — do not merge the reviewed revision as-is.

The firmware register definitions, scaling, packing, and most success-path behavior are generally consistent with the referenced Bengle firmware. The principal defects are not address-map errors; they are state-synchronization and transactionality errors around autonomous sleep/wake, the cup warmer, preheat, and LED state reporting.

Reviewed revisions

Pull request: decentespresso/decaid#612

PR head: d5a60041983ac02314f8e04585a0475a083a0541

PR base: bc56025ee4e454c9910dd876897ef674bea535a9

Review date: 2026-08-14

PR size at review: 48 changed files, +3901 / -429

The head SHA was rechecked immediately before this report was generated.

Review method and limits

I inspected the complete PR change set, the relevant call sites and tests, and the reference firmware paths named by the PR: the MMR map and the consuming code in System.cpp, System.hpp, ShotMachine.cpp, APIView.cpp, and load-cell calibration code. I also checked the current GitHub Actions result for the reviewed head.

I did not run the Flutter test suite locally and did not test real Bengle hardware. The PR's own disclosure likewise says no physical Bengle was available. Consequently, the findings below are static/protocol review findings with explicit failure sequences; hardware timing observations should still be validated physically.

Findings

H1 — The master “Auto sleep & wake” setting is not mirrored to Bengle firmware

Severity: High — merge blocker
Primary anchors:

lib/src/controllers/presence_controller.dart, _onSettingsChanged() around lines 156–166

lib/src/controllers/presence_controller.dart, _syncFirmwareScheduleAndTimeout() around lines 273–321

test/controllers/presence_controller_bengle_sync_test.dart, the case treating userPresenceEnabled = false as an unrelated change

What happens

The app-side master enable is userPresenceEnabled. The local timer and heartbeat paths respect it, but _syncFirmwareScheduleAndTimeout() does not read it. That method always derives firmware state from:

sleepTimeoutMinutes; and

wakeSchedules.

Its deduplication key also contains only those two values.

Failure sequences

Disabled before connection

The user disables Auto sleep & wake but leaves the previously configured timeout and schedules in settings.

A Bengle connects.

The controller clears its push cache and invokes _syncFirmwareScheduleAndTimeout().

The nonzero timeout and schedule table are pushed anyway.

After the tablet disconnects, firmware-owned sleep/wake remains active although the master feature is visibly disabled.

Disabled after connection

Enabled state has already been mirrored to firmware.

The user turns the master switch off.

_onSettingsChanged() calls the sync method.

Because timeout and schedule JSON did not change, the cache check returns immediately.

The persisted inactivity timeout and the RAM wake table remain active.

The new test suite reinforces the wrong contract by changing userPresenceEnabled to false and expecting no firmware re-push.

Why this matters

InactivitySleepTimeout is deliberately persisted by firmware, and the wake table exists specifically to operate without the tablet. This bug turns an explicit user disable into an app-only disable while unattended firmware behavior remains enabled.

Required correction

Build and cache an effective desired firmware state, not raw settings:

final enabled = _settingsController.userPresenceEnabled;
final effectiveTimeout = enabled
    ? _settingsController.sleepTimeoutMinutes
    : 0;
final effectiveWindows = enabled
    ? translateWakeSchedules(...)
    : const <FirmwareWakeWindow>[];

The master flag must participate in the desired-state key. A true → false transition must push timeout 0 and ScheduleControl=0; false → true must restore the current configured values.

Required tests

Connect while the master flag is already false.

Toggle true → false without changing timeout or schedule JSON.

Toggle false → true.

Verify tablet disconnect leaves no firmware-owned timeout or wake table after disable.

H2 — PUT /cupWarmer can enable heating for a request that explicitly says enabled: false

Severity: High — merge blocker
Primary anchor: lib/src/services/webserver/de1handler.dart, cup-warmer PUT handler, approximately lines 99–125

What happens

For every request containing temperature, the handler writes the new setpoint and then writes CupWarmerMode=1 for backward compatibility. Only afterward does it apply an explicit enabled field.

For this valid request:

{"temperature": 55, "enabled": false}

the wire sequence is effectively:

MatSetPoint = 55

CupWarmerMode = 1

CupWarmerMode = 0

Failure impact

The firmware's control loop may observe mode 1 between the two writes and energize the mat while the machine is idle. More importantly, if the final write times out or the connection drops, the operation leaves manual heating enabled even though the request explicitly required it to be disabled.

This is a heater-control ordering defect, not merely a response-code issue.

Required correction

Backward-compatible auto-enable should occur only when enabled is omitted:

if (temperature != null) {
  await de1.setCupWarmerTemperature(temperature);
}
if (enabled != null) {
  await de1.setCupWarmerEnabled(enabled);
} else if (temperature != null && de1.bengleFeatureSurfaceSupported) {
  await de1.setCupWarmerEnabled(true);
}

An explicit false must never cause an intermediate true write.

Required tests

Byte-order test for {temperature, enabled:false} proving no enable write occurs.

Inject failure after the setpoint write and confirm the mode remains disabled.

Preserve the historical temperature-only behavior by testing {temperature:45} still enables manual mode.

H3 — Firmware schedule replacement is neither serialized nor convergent after failure

Severity: High — merge blocker
Primary anchors:

lib/src/controllers/presence_controller.dart, _syncFirmwareScheduleAndTimeout() around lines 273–321

lib/src/models/device/impl/de1/unified_de1/wake_schedule_capability.dart, pushFirmwareWakeSchedule() around lines 31–52

lib/src/models/device/impl/de1/unified_de1/firmware_mmr_gate.dart

What happens

Every relevant settings notification starts an unawaited asynchronous sync. There is no schedule-transaction queue or generation guard. The firmware MMR gate prevents collisions with firmware upload, but it does not serialize ordinary MMR transactions.

A schedule replacement itself consists of multiple independent writes:

set local clock;

disable/clear table;

append zero or more entries;

enable table.

Two replacements can therefore interleave.

Example interleaving

Update A: clear → append A1

Update B: clear → append B1 → enable

Update A resumes: append A2 → enable

The resulting table is neither A nor B. It can contain a mixed set of windows and can drive both autonomous wake and scheduled preheat.

Failure convergence defect

_lastPushedSleepTimeout and _lastPushedSchedulesJson are assigned before any I/O completes. A timeout after clear or after a partial append can leave the table disabled or incomplete while the controller records the desired state as already pushed. No retry occurs until a relevant setting changes again or the machine reconnects.

Required correction

Use one device-scoped, coalescing single-flight synchronizer:

Keep a monotonically increasing desired generation.

Serialize the complete timeout + clock + table transaction.

Coalesce intermediate updates and apply only the newest desired state next.

Update “last successfully applied” only after the full sequence succeeds.

On failure, retain dirty state and retry the newest desired generation.

Cancel or generation-check work when the connected device changes.

Required tests

Force two replacements to overlap at each await boundary and assert the final table equals only the newest desired schedule.

Inject failure after clear, after each append, and before enable; verify a retry converges.

Replace the machine while a push is in flight and prove the old task cannot mark the new device synchronized.

M1 — The wake rewrite corrects the clock before disabling the old table

Severity: Medium-High
Primary anchor: wake_schedule_capability.dart, pushFirmwareWakeSchedule() around lines 35–52

The source comment describes the safe sequence as ScheduleControl=0 → entries → ScheduleControl=1, but the implementation first writes SetLocalTimeOfWeek and only then clears/disables the table.

If a reconnect or time correction moves the firmware clock into a window from the old table, ShotMachine::checkSchedule() can observe that old window before the disable write arrives and wake the machine. The risk is small in duration but directly contradicts the “edge-safe” claim.

Correction: disable/clear first, then set the clock, append entries, and enable. Add a byte-exact order assertion.

This should be fixed as part of H3's serialized transaction.

M2 — Enabling preheat writes the enable bit before the new lead time

Severity: Medium-High
Primary anchor: lib/src/models/device/impl/de1/unified_de1/cup_warmer_capability.dart, setCupWarmerPreheat() around lines 31–40

The implementation writes:

MatPreheatEnable

MatPreheatLeadMin

When enabling with a changed lead time, firmware can begin scheduled heating using the previous persisted lead. If the second write fails, preheat remains persistently enabled with the old lead.

Correction:

Enabling: write lead first, then enable.

Disabling: write disable first.

Changing lead while already enabled: either write the lead alone or use disable → lead → enable when strict transition safety is required.

Add injected-failure tests. The current byte test expects the unsafe enable-first order and must be updated.

M3 — LED GET/reset can report state the firmware does not contain

Severity: Medium
Primary anchors:

lib/src/models/device/impl/de1/unified_de1/led_strip_capability.dart, approximately lines 64–142

lib/src/services/webserver/de1handler.dart, LED reset handler

test/unit/models/device/impl/de1/unified_de1/led_strip_capability_test.dart

There are three related consistency defects.

1. Quantization is hidden

The firmware stores 8 bits per RGB channel. _toFirmwareRgb() discards each 16-bit channel's low byte. After a successful write, however, the cache publishes the original 16-bit request rather than the quantized value actually written. A PUT followed by GET can therefore return channel values that do not exist in firmware until reset/reconnect rehydrates them.

The unit test that expects full 16-bit local precision codifies this discrepancy.

2. Partial writes leave a stale authoritative cache

Four palette registers are written sequentially. If write 2, 3, or 4 fails, firmware contains a partially updated palette while the cache still contains the old complete palette. Subsequent GET returns 200 with stale state.

3. A failed reset can return stale data as success

_hydrateLedStrip() catches read failures and leaves the previous subject value untouched. resetLedStrip() then returns that previous value. Once an initial hydration has succeeded, a later reset failure is therefore indistinguishable from a successful reload and the REST handler returns 200 instead of the promised 503.

Required correction

Cache the quantized representation that was actually sent.

Make hydration return an explicit success/result value.

On partial write or refresh failure, mark the cache unknown or rehydrate before serving it as authoritative.

A failed reset must return null/error even when an older cached value exists.

Required tests

Non-byte-aligned 16-bit input followed by GET.

Failure at each of the four palette writes.

Successful initial hydration followed by failed reset; expected REST status 503.

M4 — ledStrip is advertised on firmware for which every LED endpoint returns 404

Severity: Medium
Primary anchor: lib/src/services/webserver/de1handler.dart, capability list and LED endpoint gates

The capability handler adds ledStrip for every BengleInterface. The LED endpoints separately require bengleFeatureSurfaceSupported, because the real palette registers are part of the post-0x00803880 surface.

An older Bengle therefore receives:

{"capabilities":[...,"ledStrip",...]}

but GET, PUT, commit, and reset all return 404. Capability discovery is not truthful.

Correction: gate ledStrip with the same surface check, or retain a genuinely supported old-firmware implementation. Add an old-firmware capabilities test.

M5 — One dropped probe response disables all new Bengle features for the whole connection

Severity: Medium
Primary anchor: lib/src/models/device/impl/de1/unified_de1/bengle_feature_probe.dart, approximately lines 18–68

Feature detection performs one two-second read, latches _bengleProbeDone before the attempt, and treats every timeout or exception as proof that the surface is unsupported. This intentionally avoids repeatedly probing old firmware, but it also makes a transient BLE loss indistinguishable from an old image.

The normal MMR read path has retries; this probe has none. A single dropped response causes:

new endpoints to return 404;

LED hydration to be skipped; and

firmware timeout/schedule mirroring to be skipped

until reconnect.

Correction: use a small bounded retry policy or distinguish a deterministic unsupported response from transport failure. At minimum, test “first response dropped, second response succeeds.”

M6 — “DST-safe” is only true at push time

Severity: Medium-Low / documented operational limitation
Primary anchors:

lib/src/models/firmware_wake_window.dart, localSecondsSinceSunday()

presence_controller.dart, the connect/settings-only synchronization triggers

The local-time calculation itself is correct: it uses local calendar fields instead of subtracting epochs. But the firmware clock is pushed only on connection or relevant settings changes. A DST or timezone change during a long-lived connection leaves firmware free-running from the old local offset.

The app may continue to wake correctly while connected, but if the tablet disconnects after the offset change and before another sync, autonomous wake and preheat can be shifted by an hour.

Correction: detect timezone-offset/local-day changes and resynchronize through the same serialized mechanism from H3, or narrow the documentation and PR claim to “DST-safe when synchronized.”

L1 — Preheat and scale-calibration 404 messages name the wrong feature

Severity: Low
Primary anchor: lib/src/services/webserver/de1handler.dart

The feature labels passed to _bengleSurfaceGate() are swapped:

preheat routes pass scaleCalibration;

scale-calibration routes pass cupWarmer/preheat.

The status code is still 404, but clients and operators receive misleading error bodies.

Correction: swap the strings and assert the response body in old-firmware route tests.

Reference-firmware parity assessment

The following portions appear consistent with the reviewed Bengle reference revision:

MMR addresses and ranges for scale calibration, LED palettes, cup-warmer mode/current temperature, inactivity timeout, schedule registers, and preheat.

Scale-calibration packed-field layout and command values exposed by the PR.

Tenths-of-a-gram scaling for the calibration reference weight.

Firmware weekday mapping, midnight splitting, and 32-entry table limit.

The absence of an independent front-switch palette register and the front-strip/default-color derivation.

Write-through persistence of the LED palette and the absence of a real commit/rollback latch.

RAM-only manual cup-warmer mode and persisted setpoint/preheat configuration.

Two suspected defects were specifically ruled out:

Connection publication order: De1Controller awaits onConnect() before publishing the machine, so the connect-time controller sync observes the completed Bengle feature probe.

Immediate calibration acceptance read: the referenced firmware stages the calibration procedure synchronously from the command handler, so the immediate post-write state read is defensible. The larger risk is physical timing/hardware behavior, not an obviously asynchronous command-dispatch race.

CI and test assessment

The current PR workflow is green for head d5a60041983ac02314f8e04585a0475a083a0541. The PR reports formatting, analysis, and 2,998 passing Flutter tests; GitHub Actions reports the PR-check workflow completed successfully.

The test suite is broad on byte packing and all-success behavior. Its principal gap is fault injection and overlap testing for multi-register operations. Several tests currently encode the unsafe behavior:

master enable treated as unrelated to firmware schedule state;

preheat enable written before lead time; and

LED cache preserving precision not present in firmware.

Green CI therefore does not invalidate the findings above.

Required pre-merge checklist

Make the master power-management toggle disable firmware timeout and wake scheduling.

Remove the intermediate cup-warmer enable for explicit enabled:false requests.

Serialize/coalesce schedule synchronization and commit cache state only after success.

Disable the old schedule before changing its clock.

Use safe preheat write ordering.

Make LED cache/reset responses represent confirmed firmware state.

Align ledStrip capability discovery with route availability.

Add fault-injection and concurrency tests for all multi-register transactions.

Recommended physical smoke tests

Because the changed behavior controls heaters and unattended wake/sleep, merge confidence should include a short real-Bengle pass:

temperature + enabled:false never energizes the mat;

preheat enable/lead updates around an active wake window;

rapid repeated schedule edits converge to the final table only;

disabling Auto sleep & wake, then disconnecting the tablet, produces neither autonomous sleep timeout nor scheduled wake;

reconnect/time correction does not trigger an old schedule window;

LED PUT/reset behavior after an induced BLE interruption;

calibration zero and latch on both isolated cells.

ODevStudio:
Re-reviewed PR #612 at the current head 3f13c269ec8271679f3f891a937b025607e1d363. It is four commits ahead of the head from my previous review, remains open/mergeable, and the head did not move while I reviewed it. CI is green on this exact SHA: format, analyzer, full Flutter test job, and Linux build all passed.

Verdict: request changes

There is one remaining high-severity concurrency bug in the new retry implementation. The rest of the previous functional findings look resolved.

High — a newer schedule generation can be lost when the retry budget is exhausted

The new retry loop keeps a single attempts counter across generations. More importantly, on failure it does this:

attempts++;
if (attempts >= _maxFirmwareSyncAttempts) break;

The check for whether _firmwareSyncGeneration changed happens only later.

That creates a concrete lost-update race. Suppose generation A has already failed twice. The third attempt starts, and while it is in flight the user changes settings — especially userPresenceEnabled=false — which creates generation B. _syncFirmwareScheduleAndTimeout() increments the generation, but because _firmwareSyncInFlight already exists it correctly does not start another drain.

If that third A attempt then fails, attempts becomes 3 and the loop immediately breaks. The drain clears _firmwareSyncInFlight in finally, but nothing restarts it for generation B. The B trigger already occurred while the old drain was active.

So a master-disable request can disappear entirely:

A failure #1
A failure #2
A failure #3 starts
    user disables Auto sleep & wake -> generation B
A failure #3 completes
    attempts == 3 -> break
drain exits
generation B never pushed

There is a related weaker case: if B arrives during the second backoff, B inherits attempts == 2, so the new desired state receives only one attempt instead of its own three-attempt budget.

The added test covers a newer generation after one failure, so it does not exercise the retry-exhaustion boundary where this happens.

I would fix this before merge. Retry accounting needs to be generation-scoped. On every iteration, if the desired generation differs from the generation whose failures are being counted, reset attempts. And before giving up at the retry bound, check whether the failed generation is still current; if a newer generation exists, continue with a fresh budget instead of breaking.

A regression test should specifically hold/fail the third attempt, change the master toggle while that final attempt is in flight, and verify that the newer generation still gets pushed.

Low — the OpenAPI cup-warmer schema still describes the removed partial-firmware behavior

The new single-surface contract is clear at the route level: outdated Bengle firmware gets 404 rather than a partial cup-warmer response.

But CupWarmerState still declares only temperature as required, marks enabled nullable, and says enabled can be null on firmware predating the mode register. That state is no longer reachable with a 200 response: current firmware always returns an enabled boolean; outdated firmware is rejected by the compatibility gate.

I would change the schema to require temperature, enabled, and currentTemperature; make enabled non-nullable; keep currentTemperature nullable only because raw 0 means there is no valid sensor reading. The PR description's older “null on old firmware” wording should be cleaned up at the same time.

Previous probe-lifecycle finding

I withdraw my previous objection that the compatibility probe must explicitly reset on Bengle.onDisconnect().

After tracing the device lifecycle, a real BLE disconnect removes that device from the discovery service, and a subsequent connection is instantiated through DeviceFactory, producing a fresh Bengle object. Bengle.onDisconnect() also disposes/closes object-owned streams, which is consistent with the object not being reused for a later external connection. The probe's per-object latch is therefore compatible with the actual external connection lifecycle; the same-object transport recovery path does not represent a different firmware installation. The bounded two-attempt probe itself now looks reasonable.

The earlier master-toggle, cup-warmer ordering, schedule transaction ordering, LED cache truthfulness, preheat ordering, capability gating, label, and DST-documentation findings remain resolved in the current code. The all-or-nothing Bengle firmware-surface refactor is also substantially cleaner than the previous partial-capability model.

So the current position is: one real merge blocker remains in _drainFirmwareSync() plus a small OpenAPI cleanup. After the generation/retry race is fixed and tested, I would be comfortable approving this from static/code-review perspective. The existing limitation remains that the calibration, unattended wake, preheat, and warmer paths still have not been smoke-tested on physical Bengle hardware.

PR #619, opened 2026-08-17
feat(api): add plugin source update endpoint

allofmeng:
## Summary

Adds a REST endpoint for updating a plugin's manifest and source without going through a directory install.

- `PUT /api/v1/plugins/<id>/source` takes `{"manifest": {...}, "plugin": "<js source>"}` and writes `manifest.json` + `plugin.js` for that id, creating the plugin if it is not installed yet.
- Partial update by design: every other file in the plugin directory is left untouched.
- Downgrades are rejected (409). The submitted version must be >= the installed version; an older version requires removing the plugin first. Enforced in `PluginLoaderService` for both `addPlugin` and the update path, using one semantic-version comparison (`comparePluginVersions`) that also drives the bundled-plugin startup copy.
- Transactional for a loaded plugin: unload, replace, reload; if the new source fails to load, the previous manifest, source, and running plugin are restored and the request returns 500. An unloaded plugin stays unloaded and its source is not executed.
- `PluginLoaderService.updatePluginSource` owns the whole operation (loaded-state preservation, swap, rollback). Per-plugin mutations - install, update, remove, load, unload, reload - are serialized behind a per-plugin lock.
- Settings and the auto-load preference survive an update.

## Linked Issue

Fixes # N/A

## Verification

- `test/webserver/plugins_handler_update_test.dart` (14 tests): overwrite, fresh install, extra-file preservation, settings + auto-load preservation, reload-if-loaded, stay-unloaded, rollback on failed reload, downgrade rejection (409), same-version accept, id mismatch, missing source, invalid manifest, unsafe id (also asserts no directory escape).
- `test/plugins/plugin_loader_update_test.dart` (7 tests): service-level downgrade rejection through `addPlugin`, same/newer accept, prerelease ordering, manifest id mismatch, concurrent updates, update racing an unload, update racing a load.
- `test/plugins/plugin_version_test.dart` (6 tests): version comparison including prerelease, build metadata, short versions.
- `flutter analyze`: `No issues found!`
- `flutter test`: 3043 passing. The 4 failures in `test/webui_support/webui_token_injection_test.dart` are pre-existing on this checkout - verified by stashing this branch's changes and re-running that file, same 4 failures.

## Impact

- API: new `PUT /api/v1/plugins/{id}/source`; `assets/api/rest_v1.yml` updated in the same commit. Like `/settings`, the path shadows a plugin-declared endpoint named `source`.
- Behavior change beyond the new endpoint: `addPlugin` now rejects installing an older version over a newer one (previously it replaced silently). The UI sideload path hits this too - remove the plugin first to go back to an older version.
- Docs: `doc/Plugins.md` REST table plus a section with a curl example, the rules above, and the security note.
- Plugins that need new extra files still have to be installed from a directory.
- Overwriting a bundled plugin is temporary: on next app start the bundled copy wins when its version is higher.
- Security: this is the first implemented REST path that accepts arbitrary plugin JS, and plugin routes are not covered by the account-proxy bearer-token middleware, so anyone who can reach the API server on the LAN can install or replace plugin code that Decaid then runs in-process. That matches the existing trusted-LAN model for the API server rather than tightening it; `doc/Plugins.md` now states it explicitly. A shared write-authorization policy for plugin and WebUI/skin mutation endpoints is worth a follow-up.

## Contributor Responsibility

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -2715,6 +2715,73 @@ paths:
                     type: string
 
   /api/v1/plugins/{id}:
+    put:
+      summary: Create or overwrite a plugin
+      description: >
+        Writes manifest.json and plugin.js for the given plugin id, replacing
+        the installed plugin when one exists. The manifest id must match the
+        path id. A plugin that was loaded is unloaded before the write and
+        reloaded afterwards; an unloaded plugin stays unloaded. Overwriting a
+        bundled plugin is temporary: on the next app start the bundled copy
+        wins whenever its version is higher.
+      tags: [Plugins]
+      parameters:
+        - name: id
+          in: path
+          required: true
+          schema:
+            type: string
+          description: Plugin ID
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              required: [manifest, plugin]
+              properties:
+                manifest:
+                  type: object
+                  description: Contents of manifest.json, including a matching id
+                plugin:
+                  type: string
+                  description: JavaScript source written to plugin.js
+      responses:
+        "200":
+          description: Plugin written
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  message:
+                    type: string
+                  id:
+                    type: string
+                  version:
+                    type: string
+                  loaded:
+                    type: boolean
+        "400":
+          description: >
+            Invalid JSON body, missing manifest or plugin source, invalid
+            manifest, manifest id mismatch, or unsafe plugin id
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+        "500":
+          description: Plugin could not be written, or was written but failed to reload
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
     delete:
       summary: Delete a plugin
       description: Removes a plugin from the system
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -2714,16 +2714,23 @@ paths:
                   id:
                     type: string
 
-  /api/v1/plugins/{id}:
+  /api/v1/plugins/{id}/source:
     put:
-      summary: Create or overwrite a plugin
+      summary: Create or overwrite a plugin's manifest and source
       description: >
-        Writes manifest.json and plugin.js for the given plugin id, replacing
-        the installed plugin when one exists. The manifest id must match the
-        path id. A plugin that was loaded is unloaded before the write and
-        reloaded afterwards; an unloaded plugin stays unloaded. Overwriting a
-        bundled plugin is temporary: on the next app start the bundled copy
-        wins whenever its version is higher.
+        Writes manifest.json and plugin.js for the given plugin id, creating the
+        plugin when it is not installed yet. This is a partial update: every
+        other file in the plugin directory is left untouched. The manifest id
+        must match the path id, and the manifest version may not be lower than
+        the installed version - an older version can only be installed after the
+        plugin has been removed. A plugin that was loaded is unloaded, replaced,
+        and reloaded; if the new source fails to load, the previous manifest,
+        source, and running plugin are restored and the request fails. An
+        unloaded plugin stays unloaded and its source is not validated.
+        Settings and the auto-load preference are preserved. Overwriting a
+        bundled plugin is temporary: on the next app start the bundled copy wins
+        whenever its version is higher. This path shadows a plugin-declared
+        endpoint named "source", the same way /settings does.
       tags: [Plugins]
       parameters:
         - name: id
@@ -2748,7 +2755,7 @@ paths:
                   description: JavaScript source written to plugin.js
       responses:
         "200":
-          description: Plugin written
+          description: Plugin source updated
           content:
             application/json:
               schema:
@@ -2773,8 +2780,10 @@ paths:
                 properties:
                   error:
                     type: string
-        "500":
-          description: Plugin could not be written, or was written but failed to reload
+        "409":
+          description: >
+            The submitted version is lower than the installed version; remove
+            the plugin before installing an older version
           content:
             application/json:
               schema:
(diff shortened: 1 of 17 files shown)


tadelv (changes requested):
The API direction looks reasonable, but I think the write/update semantics need hardening before merge.

Blocking changes requested:

1. **Do not destroy unrelated plugin files.** `writePlugin()` stages only `manifest.json` + `plugin.js` and then calls `addPlugin()`, which deletes the whole existing plugin directory. Updating a plugin that has assets/other files therefore silently removes them. If this endpoint is intended as a source update, please preserve all other files (and consider naming it `/source` to make the partial-update semantics explicit). If it is intended as full replacement, the request needs to represent the complete plugin package.

2. **Make replacement transactional for loaded plugins.** The current flow unloads the working plugin, deletes/replaces it, then tries to load the new version. A syntax/runtime/onLoad failure leaves the user with the bad version persisted and the previously-working plugin gone. Please stage/validate/swap with rollback semantics so that a failed update restores the previous files/cache and reloads the previous plugin where possible. A failed PUT should leave the installed/running plugin unchanged.

3. **Reject downgrades while a plugin is installed.** Right now `addPlugin()` does not compare versions, so this endpoint can replace e.g. `2.0.0` with `1.0.0`. An older version must only be installable after the existing plugin has first been uninstalled. Please enforce this in `PluginLoaderService` rather than only in the HTTP handler, and add tests for lower/equal/higher versions. Note that bundled-plugin startup has a version comparison already, but normal installs do not; ideally use one consistent version-comparison policy. Please also handle prerelease/semantic versions correctly rather than relying on integer-only dotted components if those are valid manifest versions.

4. **Keep lifecycle/update ownership inside `PluginLoaderService` and serialize per-plugin mutations.** The handler currently captures `wasLoaded`, `addPlugin()` may unload, and then the handler decides whether to reload. That splits one logical operation across layers and leaves races with concurrent PUT/enable/disable/load operations. Please expose a service-level update operation that owns preserve-loaded-state, swap/rollback, and per-plugin locking.

Please add failure-path coverage for: failed reload of a previously loaded plugin with rollback to the old version; preservation of extra files (if this remains a partial source update); settings + auto-load preservation; downgrade rejection; and concurrent update/lifecycle operations.

One security/documentation point as well: this is the first implemented REST path that accepts arbitrary plugin JS, whereas `/plugins/install` is still unimplemented. The API server listens on the LAN and plugin routes are not covered by the account-proxy bearer-token middleware. If the trusted-LAN model is intentional, that is fine for this PR, but the description should not characterize the network security impact as entirely unchanged. Longer term, code/content mutation endpoints for Plugins and WebUI/Skins should probably share an explicit write-authorization policy.

allofmeng:
Thanks - all four blocking points addressed in 14f2f2b8.

1. **No longer destroys unrelated files.** The endpoint moved to `PUT /api/v1/plugins/<id>/source` and is an explicit partial update: it writes `manifest.json` and `plugin.js` into the existing plugin directory instead of staging a directory and calling `addPlugin()`, so assets and any other files survive. Covered by 'keeps files the update does not carry'.

2. **Transactional swap with rollback.** `updatePluginSource` captures the previous manifest, previous source, and loaded state, writes, then reloads. On a load failure it restores both files and the manifest cache and reloads the previous version, then rethrows (500). A newly created plugin is removed instead. Covered by 'rolls back to the running version when the new source fails to load', which asserts the old version is back on disk, in the cache, and running. An unloaded plugin stays unloaded and its source is not executed - I did not add speculative validation for that case.

3. **Downgrades rejected in the service.** New `comparePluginVersions` (semver: numeric core, prerelease below release, numeric-before-alpha identifiers, build metadata ignored) is now the single comparison used by both the bundled-plugin startup copy and `_ensureNotDowngrade`, which `addPlugin` and `updatePluginSource` both call. Lower rejects with `PluginDowngradeException` -> 409; equal and higher are accepted, so republishing the same version during development still works. Say the word if equal should be rejected too. Tests cover lower/equal/higher plus prerelease ordering, at both the service and HTTP level. Note this also changes the UI sideload path: reinstalling an older version now requires removing the plugin first.

4. **Ownership and locking in the service.** `updatePluginSource` owns preserve-loaded-state, swap, and rollback; the handler only parses the body and maps errors. `addPlugin`, `updatePluginSource`, `removePlugin`, `loadPlugin`, `unloadPlugin`, and `reloadPlugin` all run under a per-plugin mutation lock (internal unlocked variants avoid reentrancy), so concurrent PUT/enable/disable/load calls for one id serialize. Tests cover concurrent updates, update racing an unload, and update racing a load.

Failure-path coverage requested: rollback on failed reload, extra-file preservation, settings + auto-load preservation, downgrade rejection, and concurrent update/lifecycle are all in `test/webserver/plugins_handler_update_test.dart` and `test/plugins/plugin_loader_update_test.dart`.

On the security point - you are right, the original description overstated it. The PR description now says plainly that this is the first implemented REST path accepting arbitrary plugin JS, that plugin routes are not behind the account-proxy bearer token, and that anyone on the LAN who can reach the API server can install or replace in-process plugin code; `doc/Plugins.md` carries the same warning. I left the trusted-LAN model as is rather than inventing an auth scheme here - a shared write-authorization policy for plugin and WebUI/skin mutation endpoints seems like its own PR.

tadelv (changes requested):
Thanks, the previous blockers are largely addressed. Two remaining issues still need tightening before merge:

1. **Make the source update transactional across the write phase, not only reload failure.** `updatePluginSource()` captures the old state and unloads the running plugin, but the rollback path currently only surrounds `_queuedLoad()`. If writing `manifest.json` succeeds and writing `plugin.js` fails (disk full, I/O error, permission failure, interruption, etc.), the plugin can be left with mismatched files and, if it was running, unloaded. Please make every mutation after the old state is captured/unloaded recoverable. Prefer staging the new core files first and swapping only once both are complete; otherwise ensure write failures restore the previous files/cache and running state too. Add failure-path coverage if practical.

2. **Make enable/disable atomic service-level lifecycle operations.** The per-plugin mutation lock now correctly serializes install/update/remove/load/unload/reload, but the REST enable/disable flows still perform `setPluginAutoLoad()` plus loaded-state checks plus `loadPlugin()`/`unloadPlugin()` as separate operations. That leaves the auto-load preference outside the mutation lock and allows concurrent enable/disable/update calls to produce inconsistent runtime vs auto-load state. Please move these into `PluginLoaderService.enablePlugin(id)` / `disablePlugin(id)` (or equivalent), each holding the per-plugin mutation lock for the complete state transition, and have the HTTP handler delegate to those methods. Please add a concurrent enable/disable/lifecycle test.

Everything else from the earlier review now looks addressed: `/source` partial-update semantics preserve extra files, downgrade prevention is enforced service-side (including `addPlugin`), semantic prerelease ordering is shared with bundled-plugin handling, failed reload restores the prior running plugin, settings/autoload survive source updates, and the security/trusted-LAN behavior is documented.

tadelv (approved):

Issue #626, opened 2026-08-18
Bring plugin installation and updates to parity with WebUI skins


tadelv:
## Why

Plugin runtime/lifecycle support is mature, but plugin distribution is still mostly manual. WebUI skins already have source-aware installation and update tracking; plugins should have the same basic management model so users can install and keep third-party plugins current without replacing folders by hand.

## What

Add first-class plugin installation from:

- GitHub Release — stable/user distribution channel
- GitHub branch — development/bleeding-edge channel
- local ZIP — manual snapshot, no updates
- local folder — existing native Flutter behavior: copy a snapshot into Decaid-managed plugin storage, no live execution from the selected folder

`manifest.json` remains required for every plugin and remains authoritative for plugin identity, version, API version and permissions.

GitHub-backed installs must persist Decaid-owned source metadata and participate in automatic update checks on the same cadence as WebUI skins/app update checks. Local ZIP/folder installs are untracked snapshots and are not auto-updated.

### Version/update rules

- GitHub Release installs are strict: release tags must be `X.Y.Z` or `vX.Y.Z` and must exactly match `manifest.json.version` after removing the optional leading `v`.
- Release assets must contain exactly one valid plugin payload. If a release has multiple ZIP assets, require an explicit asset selection instead of guessing.
- GitHub branch installs track the branch and resolved commit/revision. A changed branch commit is an update even when `manifest.version` is unchanged.
- Existing downgrade protection for released/versioned plugins must remain intact.
- Preserve enabled/disabled state across updates. A running plugin should be transactionally replaced and restarted; a disabled plugin should remain disabled.
- Failed replacement must leave the previous installed/running version intact.

### Permission escalation

Automatic updates may proceed only when the candidate manifest requests the same permissions or a strict subset of the currently installed permissions.

If an update adds any permission:

- do not install it automatically;
- record/surface that an update is available but requires approval;
- show the added permission(s) to the user;
- install only after explicit user approval.

Repository trust does not imply permission-escalation trust.

## Agent instructions

[...]

PR #627, opened 2026-08-18
feat(plugins): install and update plugins from GitHub, ZIP and folders

allofmeng:
## Summary

Brings plugin installation and updates to parity with WebUI skins: install from a GitHub release, a GitHub branch, a local ZIP or a local folder, with provenance and automatic update checks for the GitHub-backed sources.

- **One validation path** (`lib/src/plugins/plugin_package.dart`). Any staged directory resolves to a single plugin root - flat, or one wrapper directory as GitHub archives and release ZIPs produce - and must carry a parseable `manifest.json`, a `plugin.js` and a safe id. Several candidate roots are rejected rather than guessed. Nothing installed is touched until validation passes.
- **Transactional install** (`PluginLoaderService.installPluginPackage`). Under the existing per-plugin mutation lock: rename the installed directory aside, copy the candidate in, reload if it was running; on any failure restore the previous files, manifest cache and runtime. `addPlugin` delegates to it, so the folder picker path is unchanged.
- **Provenance and updates** (`PluginSourceService`). Release installs require a `X.Y.Z`/`vX.Y.Z` tag equal to the manifest version and exactly one `.zip` asset unless the caller names one. Branch installs record the resolved commit, so a moved branch is an update even when the manifest version does not change. Local ZIP/folder installs are untracked snapshots and never auto-update.
- **Metadata storage.** A Decaid-owned `.rea_source.json` inside the plugin directory (the same shape of decision as WebUI's `.rea_metadata.json`). Removal deletes the directory, so removal clears the metadata with no extra wiring; the partial `PUT /source` write leaves it alone.
- **Permission escalation.** An automatic update installs only when the candidate keeps the same permissions or fewer. Anything added is recorded as a `pendingUpdate` carrying the candidate version and the permission delta, surfaced in the UI and the API, and installed only after explicit approval.
- **Update cadence.** `updateAllPlugins()` runs beside `WebUIStorage.updateAllSkins()` in `UpdateCheckService` - no new timer. One plugin's failure is recorded in `lastError` and never stops the others.
- **REST.** `POST /api/v1/plugins/install/github-release`, `/install/github-branch`, `/update`, `/{id}/update/approve`; `GET /api/v1/plugins` now reports `source` and `pendingUpdate`. The old URL installer stub still returns 501 but names the endpoints to use; arbitrary remote ZIP URLs stay out of scope per the issue.
- **Native UI.** The install button becomes a four-source menu, plus a "Check for updates" action, a per-plugin provenance line, and a permission-delta dialog before an escalating update installs.

## Linked Issue

Refs #626

## Verification

- `flutter analyze`: `No issues found!`
- `flutter test`: 3151 passing. The 4 failures in `test/webui_support/webui_token_injection_test.dart` are environmental on this machine, not from this branch - a running Decaid instance holds port 3000 (`SocketException: Address already in use, port = 3000`); stashing the branch and re-running that file reproduces them identically.
- New tests:
  - `test/plugins/plugin_package_test.dart` (7): flat root, single wrapper, several roots rejected, missing manifest, missing `plugin.js`, malformed manifest, unsafe id.
  - `test/plugins/plugin_source_service_test.dart` (20): release install + provenance, non-semver tag rejected, tag/manifest mismatch rejected, `v`-less tag accepted, several zip assets require explicit selection, branch install + commit, branch update with unchanged manifest version, unchanged commit only moves `lastChecked`, reduced permissions auto-update, added permission blocks and records approval-required, approval installs, running plugin restarted, disabled stays disabled, settings survive, one failure does not stop others, local zip (flat and wrapped) not auto-updated, removal clears metadata, failed install leaves the previous version running.
  - `test/webserver/plugins_handler_install_test.dart` (9): both install endpoints, listed source fields, 400 on missing repo and on tag mismatch, 501 URL stub, update-all, pending update listed then approved, 409 without a pending update, 404 for unknown plugin.
  - `test/plugins/plugins_settings_view_sources_test.dart` (4): install menu, manual update check, provenance line, permission-delta confirmation (cancel installs nothing, approve installs).
  - `test/update_check_service_test.dart` (+2): managed plugins updated on the normal cadence, including macOS.

## Impact

- API: four new endpoints; two new fields on `GET /api/v1/plugins`. `assets/api/rest_v1.yml` updated in the same commit (`PluginSource`, `PluginPendingUpdate`, `PluginInstallResult` schemas), `doc/Api.md` table updated.
- Docs: `doc/Plugins.md` gains a Distribution and Updates section - the four sources, the release packaging contract (tag == manifest version, one zip asset), update rules, and the permission-escalation rule.
- Behavior: `addPlugin` is now transactional and reloads a plugin that was running; previously it deleted the directory first and left a running plugin unloaded.
- Security: automatic updates can no longer widen a plugin's permissions. Plugin routes remain outside the account-proxy bearer token, unchanged by this PR and already documented.
- Not included, per the issue: arbitrary remote ZIP URL installation; bundled plugins keep the app-owned startup copy path.
- Design rationale archived at `doc/plans/archive/plugin-distribution/plugin-distribution-parity.md`.

Note: the GitHub release/branch helpers are new (`lib/src/util/github_archive.dart`) and used by plugins only. `WebUIStorage` still has its own inline HTTP/release parsing - rewiring skins onto the shared helper is worth doing but risks skin regressions, so I left it for a follow-up rather than bundling it here.

## Contributor Responsibility

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ Requirements: Flutter (stable), the [GitHub CLI](https://cli.github.com) (`gh`, 
 
 ```bash
 flutter pub get
-./scripts/fetch_dye2_plugin.sh   # installs assets/plugins/dye2.reaplugin/ from the pinned allofmeng/dye2 release (scripts/fetch_dye2_plugin.sh)
+./scripts/fetch_dye2_plugin.sh   # installs assets/plugins/dye2.reaplugin/ from the pinned decentespresso/dye2 release (scripts/fetch_dye2_plugin.sh)
 
 # Run with simulated hardware (no DE1 / scale required):
 flutter run --dart-define=simulate=1
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -3115,8 +3115,10 @@ paths:
     post:
       summary: Approve and install a permission-escalating update
       description: >
-        Installs the update recorded in pendingUpdate, accepting the added
-        permissions.
+        Installs the exact candidate recorded in pendingUpdate - the stored
+        release tag or commit - accepting its added permissions. If the source
+        moved since detection, the fetched candidate is refused, recorded as
+        the new pendingUpdate and has to be approved again.
       tags: [Plugins]
       parameters:
         - name: id
@@ -3144,7 +3146,9 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
         "409":
-          description: No update is awaiting approval
+          description: >
+            No update is awaiting approval, or the candidate no longer matches
+            the approved one
           content:
             application/json:
               schema:
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -2956,8 +2956,41 @@ paths:
 
   /api/v1/plugins/install:
     post:
-      summary: Install a plugin from URL
-      description: Installs a plugin from a remote URL (not yet implemented)
+      summary: Install a plugin from URL (not supported)
+      description: >
+        Installing from an arbitrary URL is not supported. Use
+        /api/v1/plugins/install/github-release or
+        /api/v1/plugins/install/github-branch.
+      tags: [Plugins]
+      requestBody:
+        required: false
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                url:
+                  type: string
+      responses:
+        "501":
+          description: Not supported; the response names the endpoints to use
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+
+  /api/v1/plugins/install/github-release:
+    post:
+      summary: Install a plugin from a GitHub release
+      description: >
+        Downloads a release asset, validates the package, and installs it. The
+        release tag must be X.Y.Z or vX.Y.Z and must equal the manifest version
+        after stripping the optional leading v. When the release carries several
+        .zip assets, assetName must name the one to install. Downgrades are
+        rejected; remove the plugin first to go back to an older version.
       tags: [Plugins]
       requestBody:
         required: true
@@ -2965,31 +2998,163 @@ paths:
           application/json:
             schema:
               type: object
-              required:
-                - url
+              required: [repo]
               properties:
-                url:
+                repo:
                   type: string
-                  description: URL to install the plugin from
+                  description: owner/repo
+                assetName:
+                  type: string
+                  description: Release asset to install; required when several .zip assets exist
+                includePrerelease:
+                  type: boolean
+                  default: false
       responses:
-        "501":
-          description: Not yet implemented
+        "200":
+          description: Plugin installed
           content:
             application/json:
               schema:
-                type: object
-                properties:
-                  error:
-                    type: string
+                $ref: '#/components/schemas/PluginInstallResult'
         "400":
-          description: Missing url field
+          description: Invalid body, invalid package, or tag/version mismatch
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "409":
+          description: The release is older than the installed version
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "500":
+          description: Download or installation failed
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+
+  /api/v1/plugins/install/github-branch:
+    post:
+      summary: Install a plugin from a GitHub branch
+      description: >
+        Resolves the branch head, downloads the branch archive, validates the
+        package, and installs it. The resolved commit is stored and later drives
+        updates, so a branch-backed plugin updates even when its manifest
+        version does not change.
+      tags: [Plugins]
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              required: [repo]
+              properties:
+                repo:
+                  type: string
+                  description: owner/repo
+                branch:
+                  type: string
+                  default: main
+      responses:
+        "200":
+          description: Plugin installed
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PluginInstallResult'
+        "400":
+          description: Invalid body or invalid package
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "409":
+          description: The branch carries an older version than the installed one
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "500":
+          description: Download or installation failed
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+
+  /api/v1/plugins/update:
+    post:
+      summary: Check every GitHub-backed plugin for updates
+      description: >
+        Runs the same check the app runs on its update cadence. Unchanged
+        plugins only get a new lastChecked; changed plugins are validated and
+        installed when the candidate keeps the same permissions or fewer, and
+        recorded as a pendingUpdate when it asks for more. One plugin failing
+        does not stop the others.
+      tags: [Plugins]
+      responses:
+        "200":
+          description: Update check finished
           content:
             application/json:
               schema:
                 type: object
                 properties:
-                  error:
+                  message:
                     type: string
+        "500":
+          description: The update check could not run
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+
+  /api/v1/plugins/{id}/update/approve:
+    post:
+      summary: Approve and install a permission-escalating update
+      description: >
+        Installs the update recorded in pendingUpdate, accepting the added
+        permissions.
+      tags: [Plugins]
+      parameters:
+        - name: id
+          in: path
+          required: true
+          schema:
+            type: string
+      responses:
+        "200":
+          description: Plugin updated
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PluginInstallResult'
+        "400":
+          description: The candidate package is invalid
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "404":
+          description: Plugin not found
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "409":
+          description: No update is awaiting approval
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "500":
+          description: Installation failed and the previous version was restored
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
 
   # Profiles API
   # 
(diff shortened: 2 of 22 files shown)


tadelv (changes requested):
The overall structure matches #626 well, but there are two update-integrity/security issues that need fixing before merge, plus one policy mismatch.

1. **[P1] Bind every update candidate to the plugin being updated.** `_updatePlugin()` resolves a candidate package and compares its permissions, but never verifies `package.manifest.id == installed.id`. `_install()` then delegates to `installPluginPackage()`, which derives its mutation lock/destination from the *candidate* manifest id. A release/branch for plugin A can therefore change its manifest id to B and the update path will install/replace B instead of rejecting the update. This also means permission comparison is being done between unrelated plugin identities. Reject an update before any mutation unless the candidate manifest id exactly matches the installed plugin id/source owner. Please add release + branch regression tests, including the case where B already exists.

2. **[P1] Approval must be bound to the exact pending candidate the user reviewed.** `approvePendingUpdate()` only checks that a pending update exists, then calls `installFromGitHubRelease()` / `installFromGitHubBranch()`, which resolve the current latest release/branch again. If the release selection or branch head changes between detection and approval, the user can approve permissions for candidate N but Decaid installs candidate N+1 without comparing that candidate's added permissions to what was approved. For branches this is especially easy because the head is mutable. Persist/use the pending release tag or commit as the approval target, fetch that exact revision, and revalidate id/version/permissions before mutation. If the candidate differs from what was approved (including any additional permission), require a new approval. Add tests where the remote revision moves after `pendingUpdate` is recorded.

3. **[P2] Keep downgrade protection for branch updates.** The branch path passes `allowDowngrade: true`. A changed commit with an unchanged manifest version already passes the existing comparator, so bypassing downgrade protection is unnecessary for the required branch behavior and lets a force-pushed/changed branch silently roll the plugin back to a lower manifest version. #626 explicitly says existing downgrade protection must remain intact. Remove this bypass unless we deliberately change that policy.

Smaller consistency gap: the native folder-snapshot path still calls `pluginLoaderService.addPlugin(...)`, so it never uses `PluginSourceService.installFromFolder()` and therefore never records/displays the `local_folder` source metadata even though that source kind/UI rendering now exists. This is not as important as the three items above, but it looks unintended and leaves `installFromFolder()` effectively unused by the actual native flow.

CI being green is useful, but the current tests do not cover candidate-id changes or remote movement between permission detection and approval, so these invariants currently pass unnoticed.

tadelv:
Additional blocker from checking bundled DYE2 behavior:

Bundled `dye2.reaplugin` never enters the new managed-update system. `PluginSourceService.updateAllPlugins()` only considers plugins with source metadata, but `_copyBundledPlugins()` copies only `manifest.json` and `plugin.js`; it never seeds `.rea_source.json`. Therefore a normal Decaid install will not check bundled DYE2 for updates.

Please make DYE2 a GitHub-release-backed bundled seed:

- canonical repo must be `decentespresso/dye2` (the current `scripts/fetch_dye2_plugin.sh` still defaults to `allofmeng/dye2`);
- when the bundled DYE2 is installed, persist equivalent source metadata for `dye2.reaplugin`: `kind=github_release`, `repo=decentespresso/dye2`, the exact bundled release tag/version, the release asset name when known, and `includePrerelease=false`;
- this metadata must be seeded/migrated even when the plugin directory already exists and the bundled version is not copied. Otherwise existing Decaid users upgrading to this feature will still have DYE2 with no provenance and it will remain invisible to `updateAllPlugins()`;
- after provenance is established, a newer GitHub release should flow through the normal permission-escalation/update path. The existing bundled-version startup behavior can remain the app-owned seed/floor and should not overwrite a newer already-updated DYE2.

Please add tests for both a fresh bundled DYE2 install and migration of an already-present DYE2 with no source metadata, asserting that `sourceFor('dye2.reaplugin')` points at `decentespresso/dye2` and that the update checker subsequently queries it.

allofmeng:
Addressed in 529f3159.

**Bundled DYE2 is now release-backed.** `bundledPluginRepos` (in `plugin_source_service.dart`) maps `dye2.reaplugin` -> `decentespresso/dye2`, and `seedBundledSources()` writes `kind=github_release`, `repo=decentespresso/dye2`, `releaseTag=v<installed manifest version>`, `includePrerelease=false`. It runs at the top of `updateAllPlugins()` and when the Plugins screen loads, so it covers both a fresh bundled install and an existing DYE2 with no metadata - no reinstall needed, and no dependency on whether the bundled copy was written this launch.

**Precedence rules:** metadata pointing anywhere else (a fork, a branch) is never touched. Only a source already pointing at the bundled repo gets its tag realigned when a newer bundled copy lands, which keeps provenance honest without changing the startup copy's app-owned seed/floor behaviour. A newer GitHub release then flows through the normal path, permission gate included.

**Canonical repo:** `scripts/fetch_dye2_plugin.sh` now defaults to `decentespresso/dye2`, and `doc/Plugins.md`, `doc/AI_REPO_MAP.md` and `CONTRIBUTING.md` follow. No `allofmeng/dye2` references remain.

**One deviation, deliberate:** I do not seed an asset name. DYE2 names its asset after the version (`dye2.reaplugin-<version>.zip`), so pinning today's name would make the *next* release un-installable. The installer picks a release's single `.zip` and demands an explicit choice only when there are several, which is the correct behaviour here. Say the word if you would rather store a pattern.

**Extra fix the tests turned up:** an update now requires the candidate package to carry the plugin id being updated. Without it, a release fetched for plugin A could install plugin B and overwrite B's provenance. Covered by 'an update carrying a different plugin is rejected'.

**Tests** (`test/plugins/plugin_source_service_test.dart`, +6): fresh bundled DYE2 seeded with `decentespresso/dye2`; existing DYE2 with no metadata migrated through `updateAllPlugins()`; the checker then queries `https://api.github.com/repos/decentespresso/dye2/releases/latest` and installs v0.2.0; tag realigned after a newer bundled copy; a user-chosen fork/branch source left alone; id-mismatch rejected.

`flutter analyze` clean; `flutter test` 3157 passing (the 4 `webui_token_injection_test.dart` failures are a local port-3000 conflict on my machine, green on CI).

tadelv (changes requested):
Re-review of current head `ed4bb349`.

The DYE2 follow-up is substantially correct: bundled `dye2.reaplugin` is now seeded/migrated to `github_release` provenance for `decentespresso/dye2`, update checks subsequently query that repo, user-selected alternate provenance is left alone, and the fetch script uses the canonical repo. The candidate-ID substitution issue is also fixed by `_requireSamePlugin()` before permission comparison/install.

There are still two merge blockers from the previous review that remain unchanged:

1. **[P1] Approval is still not bound to the exact candidate the user reviewed.** `approvePendingUpdate()` still only checks that `pendingUpdate` exists and then calls `installFromGitHubRelease()` / `installFromGitHubBranch()`, which resolve the current latest release or current branch head again. The stored `pendingUpdate.releaseTag` / `commit` is not used as the fetch target and the freshly fetched candidate is not checked against the permissions/revision that were approved. A branch can therefore move after the dialog is shown and before approval, and Decaid can install a different candidate than the one whose permission delta the user accepted. Please fetch the exact stored release tag/commit (or reject if it is no longer obtainable), revalidate plugin id/version and permission delta, and require a new approval if the candidate differs. Add release + branch tests where the remote source moves after `pendingUpdate` is recorded.

2. **[P2] Branch updates still bypass downgrade protection.** The branch update path still passes `allowDowngrade: true`. Same-version branch commits already work without this bypass, so this still permits a changed/force-pushed branch to silently move from e.g. manifest 2.0.0 back to 1.0.0, contrary to #626's explicit downgrade rule. Remove the bypass and add a regression test for a branch commit whose manifest version decreases.

One additional issue in the new bundled-source migration:

3. **[P2] Clear stale pending approval when a bundled DYE2 upgrade has already satisfied it.** `seedBundledSources()` realigns the canonical source with `existing.copyWith(releaseTag: tag)`, and `PluginSource.copyWith()` preserves `pendingUpdate` by default. Example: GitHub v0.2.0 was blocked for a new permission and recorded as pending; a later Decaid release bundles DYE2 v0.2.0 and installs it as the app-owned seed/floor; source seeding then records releaseTag v0.2.0 but leaves the old 'update to 0.2.0 needs approval' state visible/actionable. When the installed bundled version is >= the pending version/revision, clear that pending state (and test it). Do not clear a genuinely newer pending update, e.g. bundled 0.2.0 with pending 0.3.0.

Minor item still present: native folder snapshot installation continues through `pluginLoaderService.addPlugin(...)`, so `PluginSourceService.installFromFolder()` remains unused and the UI does not actually record/display `local_folder` provenance for that install path.

Current PR CI is green; these are missing behavioral invariants rather than failing checks.

allofmeng:
Addressed in c1776d6b.

**1 [P1] Approval is now bound to the exact candidate.** `approvePendingUpdate()` no longer calls the install helpers. It fetches the stored revision:

- release: `fetchGitHubReleaseByTag(repo, pendingUpdate.releaseTag)` (`GET /repos/{repo}/releases/tags/{tag}`);
- branch: `gitHubCommitArchiveUrl(repo, pendingUpdate.commit)` (`https://github.com/{repo}/archive/{sha}.zip`).

Both then revalidate before any mutation: `_requireSamePlugin()`, tag/manifest agreement for releases, and a new `_requireApprovedCandidate()` that requires the candidate manifest version to equal `pendingUpdate.version` and every added permission to be one the user already approved. A candidate that differs is recorded as the new `pendingUpdate` (carrying its own version, revision and full permission delta) and rejected with `PluginApprovalRequiredException`, so the fresh delta has to be reviewed and approved in turn. A tag or commit that is no longer obtainable fails the fetch and nothing is touched.

**2 [P2] Branch downgrade bypass removed.** `allowDowngrade: true` is gone from the branch update path, and `_install()` lost the parameter entirely. A moved branch whose manifest version is unchanged still updates - the comparator is `>= 0` - and a commit whose version drops is refused and recorded in `lastError`.

**3 [P2] Stale pending approval cleared on bundled realign.** `seedBundledSources()` compares the installed manifest version against `pendingUpdate.version`: at or above it, the realign writes `clearPendingUpdate: true`; below it, the pending update is kept. Bundled 0.2.0 with pending 0.2.0 clears; bundled 0.2.0 with pending 0.3.0 keeps.

**Minor.** The native folder snapshot path now calls `PluginSourceService.installFromFolder()` instead of `pluginLoaderService.addPlugin()`, so it records and displays `local_folder` provenance.

**Tests** (`test/plugins/plugin_source_service_test.dart`, +6):

- approval installs the reviewed release when latest has moved v1.1.0 -> v1.2.0;
- a release re-cut at the approved tag with an extra permission is refused, the plugin stays on 1.0.0 and the new delta is recorded as pending;
- branch approval installs the reviewed commit while the head has moved to commit-3;
- a branch commit dropping 2.0.0 -> 1.0.0 is refused, the commit and version are unchanged and `lastError` is set;
- bundled seeding clears a pending update the bundled copy satisfies, and keeps a newer one.

The bundled tests now apply the newer bundled copy in place via `updatePluginSource()`, which is what `_copyBundledPlugins()` does. `installPluginPackage()` replaces the whole directory and was wiping `.rea_source.json`, so the pre-existing 'seeding realigns the tag' test was passing through the fresh-seed branch rather than the realign branch; it exercises the realign path now.

**Docs/spec.** `doc/Plugins.md` update rules gain the candidate-id and branch-downgrade lines, and the escalation section describes the approval binding and the stale-pending clearing. `assets/api/rest_v1.yml` describes the approve endpoint as installing the stored revision, and its 409 now covers "the candidate no longer matches the approved one".

`flutter analyze`: `No issues found!`. `flutter test`: 3163 passing; the 4 `webui_token_injection_test.dart` failures are the same local port-3000 conflict (`Address already in use, errno = 48`), green on CI.

allofmeng:
Good catch - fixed in 813f6a49. The TOCTOU was real: both branch paths resolved the head SHA and then downloaded `.../archive/refs/heads/{branch}.zip`, so a branch moving between the two requests installed newer contents under the older recorded SHA.

`installFromGitHubBranch()` and the branch arm of `_updatePlugin()` now download `gitHubCommitArchiveUrl(repo, commit)` after resolving the SHA, the same target the approval path uses. `gitHubBranchArchiveUrl()` has no callers left and is removed.

Regression tests (`test/plugins/plugin_source_service_test.dart`, +2): 'a branch install downloads the commit it records' and 'a branch update downloads the commit it records'. Both mocks resolve a SHA, serve the matching version at `/archive/{sha}.zip` and a different version at `/archive/refs/heads/{branch}.zip`, then assert the installed manifest version is the one belonging to the recorded SHA. Stashing the fix and re-running the pair fails both, so they hold the invariant rather than passing by construction:

```
00:00 +0 -2: Some tests failed.
  PluginSourceService a branch install downloads the commit it records
  PluginSourceService a branch update downloads the commit it records
```

`doc/Plugins.md` update rules note that a branch is resolved to a commit first and that commit's archive is what gets downloaded.

`flutter analyze`: `No issues found!`. `flutter test`: 3165 passing, same 4 local port-3000 failures in `webui_token_injection_test.dart`.

tadelv (approved):

ODevStudio:
Seems like i was a bit late to the party.

[P1] Transaction backup naming can delete a different installed plugin. installPluginPackage() uses <pluginId>.previous as its backup directory and unconditionally deletes that directory before determining whether the plugin being installed even has an existing version. Thus installing/updating foo deletes plugins/foo.previous if it exists. .previous is not reserved by plugin ID validation—dots are allowed—so foo.previous can be a legitimate plugin ID. This is cross-plugin data loss, and the cache can subsequently still contain the now-deleted plugin.
Fix: keep transaction state outside the plugin namespace, e.g. a dedicated transaction/temp directory with a unique name, rather than deriving a sibling directory from the plugin ID.
[P2] .rea_source.json is app-owned but not reserved from package contents. Package validation only requires manifest.json, plugin.js, and a safe ID; arbitrary additional files/directories are accepted. A package can therefore contain a directory named .rea_source.json. The loader successfully commits that package, after which _writeSource() treats the same path as a file and throws. This means an install can return failure after the new plugin has actually been committed, with no usable provenance. Worse, during updateAllPlugins() the error handler calls _writeSource() again; that second failure escapes the per-plugin catch and stops processing subsequent plugins, contradicting the PR's stated “one plugin failure never stops the others” invariant.
Fix: either reject packages containing the reserved metadata path or, preferably, store Decaid-owned provenance outside the plugin-controlled directory.
[P2] The rollback boundary does not cover the whole mutation. The running plugin is unloaded, any old backup is deleted, and the installed directory is renamed before entering the try that restores state. The final backup deletion is also after that try. Consequently, a filesystem error deleting/renaming the backup can leave the previous plugin installed but unexpectedly unloaded, while a failure deleting the backup after a successful replacement reports the install as failed even though the new plugin is already active. In the managed-source path, that latter case also prevents provenance from being written.
Fix: put unload → backup creation → replacement → reload → cleanup under one transaction/error-handling boundary, with rollback appropriate to the point at which failure occurred. A crash-recovery strategy for leftover transaction directories would also make the “transactional” guarantee substantially stronger.

tadelv:
@ODevStudio feel free to open a new follow-up issue for this. Judging by the Ps I see no serious ones though, so probably after 1.0

PR #628, opened 2026-08-18
fix(api): support tri-state partial updates

### LINKED ISSUE #556: Allow partial-update APIs to clear nullable fields
## Problem

Several partial-update handlers use nullable Dart parameters to represent both “the JSON field was omitted” and “the JSON field was explicitly set to `null`.” Those are different operations:

- omitted field: preserve the stored value;
- explicit `null`: clear a nullable stored value.

Because the distinction is lost, clients cannot reliably clear previously stored optional data. The audit identified this in bean, bean-batch, grinder, profile-metadata, and steam-setting updates. For example, [the profile update handler](https://github.com/decentespresso/decaid/blob/f8cd95535c10215630fd60787a1633e4524a427c/lib/src/services/webserver/profile_handler.dart#L154-L173) reads `json['metadata'] as Map<String, dynamic>?`; both an absent key and `"metadata": null` become `null` before reaching the controller.

## Expected behavior

All partial-update endpoints should implement consistent tri-state patch semantics:

| JSON input | Meaning |
|---|---|
| key omitted | Leave the current value unchanged |
| key present with `null` | Clear the value, if the field is nullable |
| key present with a value | Validate and replace the value |

Required or non-nullable fields should reject explicit `null` with a clear 400 response.

## Acceptance criteria

- Audit the affected bean, bean-batch, grinder, profile-metadata, and steam-setting update paths.
- Preserve field presence through the handler/controller/storage layers instead of collapsing omitted and explicit-null values.
- Apply the same semantics consistently across affected endpoints.
- Add endpoint tests proving preserve/clear/replace behavior for each affected resource and rejection for non-nullable fields.
- Update `assets/api/rest_v1.yml` and `doc/Api.md` so nullability and partial-update behavior are explicit.
- Preserve backward compatibility for requests that omit optional fields.

tadelv:
## Summary

What changed, and why?

- distinguish omitted fields from explicit `null` in bean, bean-batch, grinder, and profile metadata updates
- reject explicit `null` for non-nullable profile and steam-setting fields with `400`
- strictly validate HTTP numeric/list patch values before permissive import parsers can coerce them
- document tri-state partial-update semantics in OpenAPI and API/profile docs

## Linked Issue

Fixes #556

## Verification

- `dart format --output=none --set-exit-if-changed` on changed Dart files
- focused handler/controller tests: 82 passed
- `flutter test` — 3,121 passed, 1 skipped
- parsed `assets/api/rest_v1.yml` with PyYAML
- simulated macOS REST smoke with MockDe1 across bean, batch, grinder, profile metadata, and workflow steam settings; repeated after hot reload
- Drift-backed REST smoke proving malformed bean altitude/variety, batch numeric, and grinder numeric/list patches return `400` without mutation

## Impact

Affected partial-update endpoints now preserve omitted fields, clear nullable fields on explicit `null`, and return `400` for explicit `null` on non-nullable fields. Supplied numeric/list values must match the OpenAPI types; malformed strings and list elements are rejected before persistence. OpenAPI and user-facing API/profile documentation are updated. No database migration or security impact.

## Contributor Responsibility

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -1198,8 +1198,11 @@ paths:
 
         Each HTTP request is an independent mutation. Requests are processed in FIFO order,
         and partial request bodies are deep-merged against the workflow state when that request
-        reaches the front of the queue. The server does not debounce or combine separate
-        requests. Machine side effects from workflow PUT requests do not execute concurrently,
+        reaches the front of the queue. Omitted steam-setting fields keep their current values;
+        supplied values replace them. Steam settings are non-nullable, so explicit `null` for
+        `steamSettings` or any of its fields returns `400`.
+        The server does not debounce or combine separate requests. Machine side effects from
+        workflow PUT requests do not execute concurrently,
         so a request may wait behind an in-flight machine operation. Each response describes
         the workflow resulting from that request. High-frequency clients should throttle input
         on the client side rather than relying on server-side debounce.
@@ -1231,7 +1234,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/WorkflowRequest"
+              $ref: "#/components/schemas/WorkflowPatch"
             examples:
               updateContext:
                 summary: Update dose and grinder via context (recommended)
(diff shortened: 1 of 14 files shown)


tadelv:
Addressed the blocking review in 67f875e. HTTP patch handlers now validate all affected numeric and list fields against the OpenAPI types before calling the permissive import/persistence parsers. Added router regressions covering every affected numeric field plus bean/grinder lists; each asserts 400 and unchanged stored data. Focused tests, analyze, the full 3,121-test suite, and a Drift-backed REST smoke all pass.

PR #631, opened 2026-08-19
fix(scales): log why a BLE scale connect failed

allofmeng:
## Summary

What changed, and why?

- Every BLE scale's `onConnect` swallowed its connect exception with a bare `catch`, so a failed connect surfaced only as `WARNING ScaleController - Scale failed to connect (state: disconnected)` with the cause discarded. Transport-level detail is logged at `FINE`, which end users never run at, so a failing scale leaves no diagnosable trace at all.
- Log the exception at `WARNING` before reporting the disconnected state, in the 11 scale implementations that were silent (Atomheart, BlackCoffee, Bookoo, Difluid, Eureka, Felicita, Hiroia, Skale2, SmartChef, Varia, WeighMaster). Acaia, DecentScale and DecentTemp already logged theirs.
- Behavior is otherwise unchanged: the catch still swallows and still reports `disconnected`.

This is the missing instrumentation for #629, not a fix for it. Log analysis for that issue: the Atomheart Eclair is discovered by name and keeps advertising throughout, while each connect attempt dies at a consistent ~20.4s — matching `UniversalBle.connect(timeout: 20s)` in `universal_ble_transport.dart` plus the 300ms Android pre-connect settle. Whether that is the GATT connect timing out, service discovery failing, or the subscribe call throwing cannot be told apart from the submitted log, because of the bug this PR fixes. Next step is a repro log at `FINE`.

## Linked Issue

Refs #629 (adds the diagnostics needed to root-cause it; does not close it)

## Verification

- New `test/unit/models/atomheart_scale_connect_failure_test.dart`: two cases (transport `connect()` throws; expected service missing from discovery) assert the state goes to `disconnected` *and* a `WARNING` record carrying the reason is emitted. Written first, confirmed red before the change, green after.
- `flutter analyze` → `No issues found!`
- `flutter test` → `+3111 ~1 -4`. The 4 failures are all in `test/webui_support/webui_token_injection_test.dart` with `SocketException: Failed to create server socket (OS Error: Address already in use, errno = 48), address = 0.0.0.0, port = 3000` — a locally running Decaid instance holding port 3000 (`lsof -nP -iTCP:3000` confirms), unrelated to this change and untouched by it.
- No hardware testing: the reporter's Eclair is the only known unit and it is not on hand.

## Impact

- No behavior change. One additional `WARNING` log line per failed scale connect, which also means failed scale connects now reach the Crashlytics log buffer (`main.dart` forwards records at `>= WARNING`). Device IDs in those records go through the existing `Anonymization.scrubString`.

## Contributor Responsibility

### DIFF
--- a/lib/src/models/device/impl/atomheart/atomheart_scale.dart
+++ b/lib/src/models/device/impl/atomheart/atomheart_scale.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:typed_data';
+import 'package:logging/logging.dart';
 import 'package:reaprime/src/models/device/ble_service_identifier.dart';
 import 'package:reaprime/src/models/device/device_implementation.dart';
 import 'package:reaprime/src/models/device/transport/ble_transport.dart';
@@ -12,6 +13,8 @@ import 'package:reaprime/src/models/errors.dart';
 import '../../scale.dart';
 
 class AtomheartScale implements Scale {
+  final Logger _log = Logger('AtomheartScale');
+
   static final BleServiceIdentifier serviceIdentifier =
       BleServiceIdentifier.long('b905eaea-6c7e-4f73-b43d-2cdfcab29570');
   static final BleServiceIdentifier dataCharacteristic =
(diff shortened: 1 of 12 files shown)


tadelv (approved):
Approved. The change is appropriately scoped as generic BLE-scale connection diagnostics for #629 and preserves existing behavior while making swallowed connect failures diagnosable. The Eclair-specific protocol/reliability work identified while investigating #629 should remain separate.

PR #632, opened 2026-08-19
feat(api): log client IP and scrub serial from exported logs

### LINKED ISSUE #500: Feature request: Add client IP address to logs
I know, this would mean a potential "breaking change" if there's any automation or something reading logs but, it would be swell if the logs showed client IP address for API calls (such as the rogue tare in my other request), so that one could get a better idea of who's sending what.

tadelv:
## Summary

Two changes addressing API observability and log-export privacy:

- **Client IP in API logs** — replaced shelf's `logRequests` with a middleware that includes the client's remote address, so API call logs show who is sending what (e.g. the rogue tare from the related issue).
- **Sensitive info scrubbing on log export** — the in-app feedback flow now scrubs the machine serial number (and already-scrubbed MAC/IP) from `log.txt` and `webview_console.log` before uploading to a gist or building the HTML report.

## Linked Issue

Fixes #500

## Verification

- `flutter analyze` clean.
- New unit tests: `test/services/telemetry/anonymization_test.dart` (serial scrubbing), `test/services/webserver/request_log_middleware_test.dart` (IP extraction + logging).
- Full `flutter test`: 3185 passed, 1 skipped, 0 failures.

## Impact

- API access-log lines now include the client IP (cosmetic change to log format).
- Exported feedback logs no longer contain the machine serial number.
- No API/spec, DB, or endpoint changes.

### DIFF
--- a/lib/src/app.dart
+++ b/lib/src/app.dart
@@ -452,6 +452,7 @@ class _MyAppState extends State<MyApp> {
                       return DataManagementPage(
                         controller: widget.settingsController,
                         persistenceController: widget.persistenceController,
+                        de1Controller: widget.de1Controller,
                         profileStorageService: widget.profileStorageService,
                         beanStorageService: widget.beanStorage,
                         grinderStorageService: widget.grinderStorage,
--- a/lib/src/controllers/de1_controller.dart
+++ b/lib/src/controllers/de1_controller.dart
@@ -21,7 +21,10 @@ class De1Controller {
   Workflow? defaultWorkflow;
 
   De1Interface? _de1;
+  final Set<String> _seenSerials = {};
   final Logger _log = Logger("De1Controller");
+
+  List<String> get seenSerials => List.unmodifiable(_seenSerials);
 
   final BehaviorSubject<De1Interface?> _de1Controller = BehaviorSubject.seeded(
     null,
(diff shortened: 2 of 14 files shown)


ODevStudio (changes requested):
Findings
[P1] The actual feedback dialog never supplies the serial number, so its exports still leak it.
lib/src/feedback_feature/feedback_view.dart:47 still creates FeedbackService(githubToken: widget.githubToken) without currentSerialNumber. _scrubSensitive() can only redact the serial when that optional callback returns one, so both Gist uploads and generateHtmlReport() from this dialog get no serial redaction.
Fix: inject the serial provider into this service too—preferably make the dependency non-optional so future call sites cannot silently disable the privacy feature. Add an integration-level test around the actual FeedbackService used by the dialog.
[P2] Scrubbing only the currently connected machine's serial leaves old serials in persistent logs.
The webserver service supplies de1Controller.connectedDe1OrNull?.machineInfo.serialNumber. On disconnect, De1Controller explicitly clears _de1, so the callback becomes null. The feedback endpoint itself does not require a connected machine. Therefore feedback sent after a disconnect—or after switching from machine A to machine B—can export serials already present in log.txt from earlier connections.
Fix: scrub serial-number log fields structurally, or retain all serials observed during the lifetime covered by the log rather than querying only the current device.
[P2] IPv6 client addresses introduced by the new logger are not scrubbed on export.
clientIpFromRequest() returns HttpConnectionInfo.remoteAddress.address, which can be an IPv6 address. But scrubString() only applies MAC and IPv4 regexes; there is no IPv6 pass. Thus a request from something like fe80::1234:abcd is deliberately added to the application log by this PR and subsequently survives feedback export unchanged.
Fix: add IPv6 scrubbing and an IPv6 request/export test.
[P2] The replacement request logger drops exception stack traces.
In the new onError branch, the logged message ends with error: $error; the supplied stackTrace is never logged. Shelf's logRequests, which this replaces, includes the exception's folded stack trace in error messages. So the observability improvement simultaneously removes one of the most useful pieces of information for diagnosing server exceptions.
Fix: retain Shelf's error formatting/stack handling and only augment it with the client IP.

PR #567, opened 2026-08-06
feat: bundle the MSL skin

allofmeng:
## Summary

- Problem: MSL ([allofmeng/MSL](https://github.com/allofmeng/MSL)) is a streamline.js-inspired WebUI skin that ships tagged releases, but it is not offered as a skin option in the app.
- Why it matters: users have to sideload it instead of picking it from the bundled skin list.
- What changed: one `github_release` entry for `allofmeng/MSL` in `skin_sources.json`. `bundle_skins.sh` picks it up with no code change — same path as the other seven sources.
- What did NOT change (scope boundary): no bundler logic, no skin install/serve code, no other skin sources. The skin is bundled at its latest release, unpinned, like every other entry.

## Change Type (select all)

- [ ] Bug fix
- [x] Feature
- [ ] Refactor required for the fix
- [ ] Docs
- [ ] Security hardening
- [ ] Chore / infra
- [x] Plugin (DYE2 or bundled skin)

## Scope (select all touched areas)

- [ ] BLE transport / device comms
- [ ] REST API / handlers
- [ ] WebSocket API
- [ ] Machine state / shot logic
- [ ] Scale / weight / flow
- [ ] Profiles / beans / grinders / workflows
- [x] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] Storage / Drift database
- [x] CI / build / infra
- [ ] Docs / specs

## Linked Issues

- Closes #
- Related #

## Root Cause (if bug fix)

N/A

## Regression Test Plan (if bug fix or refactor)

N/A — data-only config addition. Existing coverage already applies: `test/webui_storage_bundled_test.dart` asserts `skin_sources.json` parses into typed source configs and that the generated `assets/bundled_skins/manifest.json` is a non-empty list of skin IDs. `bundle_skins.sh` itself exits non-zero if any source fails to fetch, so a bad repo/asset breaks the build loudly rather than silently shipping seven skins.

## Documentation Obligations (required)

- [ ] API spec updated
- [ ] API docs updated: `doc/Api.md`
- [ ] Plugin docs updated: `doc/Plugins.md`
- [ ] Skin docs updated: `doc/Skins.md`
- [ ] Profile docs updated: `doc/Profiles.md`
- [ ] Device docs updated: `doc/DeviceManagement.md`
- [x] N/A — no docs affected (`doc/Skins.md` documents the skin API, not the bundled-skin roster; no doc enumerates the sources)

## Security Impact (required)

[...]

PR #633, opened 2026-08-19
feat(skins): add native GitHub release/branch skin install (gh-571)

### LINKED ISSUE #571: Add manual skin install option to native UI Skins settings
Allow users to install skins from the Launcher -> Skins view

tadelv:
## Summary

Adds a native **Install skin** (+) action to the Skins settings page (Launcher → Skins). Users can now install a skin from:

- a **GitHub Release** (`owner/repo`, optional `.zip` asset name)
- a **GitHub Branch** (`owner/repo`, branch, default `main`)
- a **ZIP file** (existing flow, moved from the hidden dropdown item into the menu)

All installs drive the existing `WebUIStorage` installers, which already extract, validate skin IDs, rescan the list, and persist source metadata — so **Check for updates** keeps refreshing GitHub-installed skins from their original source.

## Linked Issue

Fixes #571

## Verification

- `dart format lib test` — clean
- `flutter analyze` — No issues found
- New widget tests (`test/skin_selector_install_test.dart`): release install (repo + asset), branch install (custom branch), empty-branch-defaults-to-main, failure surfaces storage error — all pass
- Related suites pass: `skin_selector_update_refresh_test`, `skin_selector_go_to_skin_test`, `webui_storage_install_test`, `webui_handler_appstore_test`
- Full `flutter test`: 3179 pass; 4 failures in `webui_support/webui_token_injection_test.dart` are pre-existing on clean `main` (port 3000 occupied — documented in PR #567 evidence; verified `Address already in use, errno = 48`)
- Manual smoke test in a simulated app (installing a real public repo) not run — requires a running app + network

## Impact

- User-visible: new + menu on the Skins page; the old hidden "Install from .zip..." dropdown item is consolidated into it
- Docs: `doc/Skins.md` updated with the native install options
- No API/spec, storage, database, or transport changes
- Behavior note: GitHub installs do **not** auto-activate the skin (installers return no skin ID); the user picks it from the list — ZIP install keeps its existing install-and-load behavior

## Contributor Responsibility


ODevStudio (changes requested):
I found one correctness regression; CI is green, and the GitHub install paths otherwise look consistent with the existing storage APIs. The PR is open/mergeable, changes 4 files, and the PR-check workflow completed successfully.

Finding — P2: ZIP cancellation now clobbers the selected skin

_pickCustomSkinZip() still resets _selectedSkinId to webUIStorage.defaultSkin?.id when the picker is cancelled, returns no path, or installation fails. Previously that made sense because ZIP installation was represented by the synthetic __custom__ dropdown value: cancelling needed to get the dropdown out of that sentinel state.

This PR removes the __custom__ dropdown item and invokes _pickCustomSkinZip() from the new + menu instead. As a result, those resets now overwrite a legitimate current selection.

Repro:

Stop the skin server.
Select skin B while skin A remains the persisted default.
Open + → ZIP file.
Cancel the picker.
Selection unexpectedly jumps back to skin A.

The same happens after a ZIP installation error. Because selection changes while the server is stopped do not update the persisted default, this is observable normal behavior rather than a theoretical edge case. The fix should be to remove those default-selection resets from the cancel/null/error paths now that there is no _customSkinId sentinel to clean up.

I would request changes for that regression.

There are also two non-blocking test gaps. The test named empty branch defaults to main never empties the branch field—the dialog already initializes it to main—so the detail.isEmpty ? 'main' : detail branch isn't actually exercised. And the fake storage always exposes the same single skin, so the new tests don't verify the PR's stated requirement that a successful GitHub install becomes visible after the selector rebuilds.

Otherwise, the implementation is appropriately scoped to issue #571, and reusing WebUIStorage.installFromGitHub* preserves the existing extraction/provenance/update machinery.

tadelv:
Addressed all three points in a47df71:

- **P2 regression:** removed the `_selectedSkinId` resets from `_pickCustomSkinZip`'s cancel, null-path, and error branches. With the `__custom__` sentinel gone, those resets clobbered a legitimate selection; they now leave the selection untouched and only show the snackbar on error. The success path still selects + loads the installed skin.
- **Test gap 1:** 'empty branch defaults to main' now explicitly clears the pre-filled branch field (`enterText('')`) so the `detail.isEmpty ? 'main' : detail` fallback is actually exercised.
- **Test gap 2:** the fake storage now appends a skin on successful install, and the release + branch tests open the selector dropdown to assert the newly installed skin is visible after the rebuild.

PR #635, opened 2026-08-20
fix(de1): bound shot-settings reads so steam writes cannot wedge the workflow queue

### LINKED ISSUE #634: [Bug Report] **Steam didn't not stop automatically**
## Description
**Steam didn't not stop automatically**

steam didn't stop per time request

## System Info
**System Information:**
- App Version: 0.7.17 (build 2269)
- Commit: 2e9e6b39
- Branch: HEAD
- Platform: android
- OS Version: BP2A.250605.031.A3.X210XXSCEZE9

## Attachments
Logs and screenshots: https://gist.github.com/reaprime-bot/38a73870f0fe66148a44e4e41649c2c9

---
_Submitted via in-app feedback on 2026-08-20T09:54:11.186_

allofmeng:
## Summary

What changed, and why?

- `UnifiedDe1Transport._shotSettingsSubject` is an unseeded `BehaviorSubject`, seeded only by the connect-time characteristic read. When that read fails, the subject stays empty for the whole connection and `shotSettings.first` never completes.
- Every steam and hot-water write reads the current `De1ShotSettings` first, so an empty subject hung the write forever. That hang wedged the `De1Controller` device-write queue, and every later `PUT /api/v1/workflow` sat behind it until the 30 s queue wait expired with 503. Field symptom: the user's steam duration never reached the DE1 and the app reported an error 30 s later.
- `De1Controller._readShotSettings` now bounds every shot-settings read with `ConnectionTimings.initialShotSettingsTimeout`, and maps a closed subject (`StateError`) to `DeviceNotConnectedException`.
- Device-level `TimeoutException` in `WorkflowHandler` now maps to 503 rather than 500.
- No generic queue-release timeout is added. `Future.timeout()` does not cancel the underlying future, so releasing the queue would let a stalled write resume and overlap a newer one. A real anti-wedge mechanism needs explicit cancellation or fencing and belongs in its own PR.
- A connect-time read timeout no longer skips startup defaults permanently. `_deferStartupDefaults` re-arms on the first frame that does arrive, so a transient MMR timeout at connect no longer leaves the machine unconfigured until app restart. The deferred defaults run through `runDeviceWrite`, so they cannot overlap a workflow write that started after initialization settled.

## Linked Issue

Fixes #634

## Verification

- New `test/controllers/de1_controller_shotsettings_stall_test.dart`: a DE1 that never emits shot settings makes `updateWorkflowSettings` fail fast with `TimeoutException`; a following `updateFlushSettings` still runs (queue not wedged); startup defaults still run when shot settings arrive after the initial read timed out.
- `de1_controller_shotsettings_stall_test.dart` also covers the race: a flush write held open in the device layer, then a late shot-settings frame - startup defaults wait for the queue and run after the flush write completes. The test fails without the `runDeviceWrite` fencing.
- `flutter analyze` — No issues found.
- `flutter test` — 3183 passed, 1 skipped, 4 failures, all in `test/webui_support/webui_token_injection_test.dart` and all reproduced on `main` before this branch (pre-existing, unrelated).

## Impact

- Behavior: steam/hot-water/rinse writes now fail with an error instead of hanging when the DE1 never reports shot settings. Clients see a prompt 503 rather than a 30 s stall followed by 503 on every subsequent request.
- API: device read timeouts return 503 `Machine unavailable` instead of 500.
- Docs: new "shotSettings Never Arrives (gh-634)" section in `doc/AI_BLE_NOTES.md`.

## Contributor Responsibility


tadelv (changes requested):
The bounded `shotSettings` fix is the right direction and directly addresses the reproduced #634 failure. Please keep `_readShotSettings()` and use it in the steam/hot-water/default paths, including the `StateError` -> `DeviceNotConnectedException` mapping and the 503 mapping for device read timeouts.

For this PR, please remove the generic queue-release timeouts (`deviceWriteStallTimeout` and `WorkflowHandler.applyTimeout`). `Future.timeout()` does not cancel the underlying future, so after the timeout the queue can advance while the original machine/workflow write is still alive. If that old operation later resumes, it can overlap with or overwrite newer writes, which breaks the serialization guarantee and can leave machine state out of order.

The generic anti-wedge mechanism should be handled separately with explicit cancellation/fencing if we need it. For #635, bounding the actual unbounded `shotSettings.first` read is sufficient and much lower risk.

One additional ordering issue to address while keeping the deferred-startup-default behavior: `_deferStartupDefaults()` currently runs outside the device-write queue after initialization is marked settled. A normal workflow write can therefore start, then late startup defaults can arrive and write concurrently. Please preserve serialization here as well (or otherwise fence the deferred defaults), and add a regression test covering a normal write racing with late shot-settings arrival.

allofmeng:
Addressed in 6af72af.

- Removed `De1Controller.deviceWriteStallTimeout` and `WorkflowHandler.applyTimeout`, plus `ConnectionTimings.deviceWriteStallTimeout` and the `rest_v1.yml` paragraph describing the 120 s release. Agreed on the reasoning: `Future.timeout()` leaves the original write alive, so releasing the queue breaks serialization. Deleted `test/webserver/workflow_handler_queue_test.dart`, which existed only to cover that behavior.
- Kept `_readShotSettings()` and all its call sites (steam, hot water, defaults), the `StateError` -> `DeviceNotConnectedException` mapping, and the `TimeoutException` -> 503 mapping in `WorkflowHandler._applyPayload`.
- `_deferStartupDefaults` now runs `_setDe1DefaultsFor` inside `runDeviceWrite`, with an identity check on the device handed back by the queue, so late defaults are serialized against normal writes instead of racing them.
- New regression test `deferred startup defaults do not overlap an in-flight write`: a flush write is held open in the device layer, a late shot-settings frame arrives, and the defaults must not touch the device until the flush write finishes. Verified it fails without the fencing (`Expected: <0> Actual: <1>`) and passes with it.

`flutter analyze` clean. `flutter test`: 3183 passed, 1 skipped, 4 pre-existing failures in `test/webui_support/webui_token_injection_test.dart` that also fail on `main`.

tadelv (approved):
Re-reviewed after 6af72af. The previous blockers are addressed: the generic queue-release timeouts are removed, bounded `shotSettings` reads remain, device-level timeout errors map correctly, and deferred startup defaults are serialized through `runDeviceWrite` with regression coverage for the late-frame/in-flight-write race.

The remaining active serial/USB recovery work is now tracked separately in #660. With that follow-up carved out, this PR is appropriately scoped to preventing missing shot-settings frames from wedging the device/workflow write queues and is good to merge.

PR #641, opened 2026-08-20
feat(account): require native consent for proxy callers

### LINKED ISSUE #300: Account proxy: client consent prompt over active view
## Goal
When a client first requests use of the linked Decent account, show a consent prompt over the active view ("\<caller\> wants to use your Decent account") so account use is user-approved, not silent. Remember the decision per caller.

## Where it fits
Part of the Decent account auth-proxy feature — see **#296**. Branch `feature/account-proxy`, design in `doc/plans/account-proxy-design.md` ("Consent prompt").

## Reuse these
- Caller identity already exists: `proxyCallerOf(request)` (`lib/src/services/webserver/proxy_auth_middleware.dart`) returns the `ProxyCaller` (skin / `api:*` / `plugin:*`).
- `NavigationService.navigatorKey` (see `lib/src/app.dart`) for showing a dialog over any route, including the full-screen `SkinView` (`lib/src/skin_feature/skin_view.dart`).
- `ShadDialog` (used in `lib/src/settings/data_management_page.dart`) for the prompt.

## Locked decisions
- **Pause + native Flutter dialog via `navigatorKey`.** Do **not** inject the prompt into the webview via `evaluateJavascript` — keep consent in trusted native UI.
- Consent is remembered per caller (token / plugin id) so the prompt is not repeated.

## Scope
- Intercept the first proxied request per caller; gate on a stored consent decision.
- Native consent dialog (allow / deny), shown even when a skin is full-screen.
- Persist per-caller consent; a "manage / revoke consent" surface in settings is a nice-to-have.

## Open design questions (resolve before/while building)
- UX when the request originates from a remote-browser skin (the prompt appears on the device running the app, not the browser) — how to make that obvious.
- Timeout / default-deny behaviour while awaiting consent.

## Acceptance criteria
- [ ] First proxied request from a new caller prompts before forwarding.
- [ ] Deny -> request fails (403); allow -> forwards and is remembered.
- [ ] Prompt renders over `SkinView`.

## Tests
- Widget: consent dialog allow/deny.
- Unit: per-caller consent store + gate.

## Deps
Builds on #296 (caller identity). Pairs with the plugin path issue (plugins are a natural first consumer).

ODevStudio:
## Summary

- Require native consent before skins, plugins, or named API clients can use the linked Decent account through the shared proxy.
- Persist an immutable UUID for each API token and use `api:<id>` as its consent key. Labels remain presentation-only.
- Rotate skin tokens and origins together, fail closed when consent cannot be obtained, and add session-only trust flags for headless use.
- Rebase the branch onto current `main`, preserving the newer SkinView/navigation behavior.

## Linked Issue

Fixes #300

## Verification

- `dart format lib test`
- `flutter analyze --no-pub`: no issues
- Focused consent, token-storage, and skin tests: 57 passed
- Full Windows test run: 3,273 passed, 1 skipped; the only failure is the existing current-`main` LF/CRLF assertion in `connect_result_schema_test.dart`
- Linux build smoke and contribution-policy checks passed
- Regression coverage verifies duplicate labels require independent consent, revoke/recreate with the same label prompts again, and migrated legacy token IDs remain stable

## Impact

- API-token consent is now bound to a persisted token UUID rather than a reusable label. Existing label-keyed consent is intentionally not inherited.
- Legacy persisted tokens receive and save an immutable ID on first load; no database migration is required.
- Unapproved account-proxy requests now return `403` before credentials are sent or the upstream service is contacted.
- Interactive callers receive a first-use native prompt. Headless deployments must opt in with `--trust-consent=<caller-key>` or `--trust-all-consent`.
- REST/API, skin, plugin, build, and verification documentation has been updated.

## Contributor Responsibility


ODevStudio:
[P1] A stale skin page can inherit another skin’s consent. AccountConsentGate receives the generic caller ID skin and determines its identity from the currently served skin (_activeSkin()). But ProxyTokenService creates one process-wide skin token whose caller identity is always skin. Switching skins only stops/restarts the server with a new folder; it does not invalidate the document already loaded in a remote browser. Consequently, a malicious skin A can retain its bearer token in an open browser tab, the user can switch to already-approved skin B, and A can then make a proxy call that the gate attributes to skin:B. It either inherits B’s stored allow or causes a prompt labelled as B. That defeats the PR’s central per-skin consent boundary. The caller identity needs to be bound to the requesting skin/session rather than inferred from global current-server state. Note that merely rotating the token on a switch may still be insufficient for remote tabs if old and new skins retain the same :3000 origin, since stale JavaScript can potentially fetch newly served same-origin HTML.
[P2] Concurrent prompts for different callers can break the 30-second timeout. The gate coalesces only requests sharing the same consent key, so two different new callers can simultaneously invoke _promptAndPersist. Each dialog then starts its own timer, but the timeout calls Navigator.of(context).pop() on the shared navigator. If A’s dialog is underneath B’s, A’s timer fires first and pops topmost B. A remains on the navigator, but its one-shot timer has already fired, so A can now remain open indefinitely and its HTTP request exceeds the specified timeout. Serialize consent dialogs across keys, or dismiss the specific dialog route rather than blindly popping the navigator’s top route. A two-distinct-callers timeout widget test should reproduce this.

tadelv:
fyi: will need rebase due to #650

tadelv (changes requested):
One blocker before merge:

API-client consent is currently keyed by the mutable token label (`api:<label>`), not by an immutable identity belonging to that specific token. Since labels can be reused, this allows consent to outlive token revocation: create token `station` → allow it → revoke it → create a new token also named `station`, and the new bearer token inherits the old consent without a first-use prompt. Two simultaneously valid tokens with the same label would also share consent.

Please give persisted API tokens an immutable non-secret ID (for example a UUID) and use `api:<id>` as the consent key, keeping `label` presentation-only. A digest of the bearer token would also avoid collisions, but an explicit persisted ID is cleaner.

Please add regressions covering duplicate labels and revoke/recreate-with-same-label requiring independent consent.

Separately, this PR also needs a rebase onto current `main` before merge; the branch is currently non-mergeable and there are overlapping SkinView/navigation changes since its last merge from main.

tadelv (approved):

PR #642, opened 2026-08-20
fix: report Bengle virtual scale as connected in device APIs

### LINKED ISSUE #630: Bengle integrated scale is reported disconnected while active
## Summary

When a Bengle is connected, its integrated `BengleVirtualScale` is attached to `ScaleController` and streams valid data, but both device-list surfaces report it as unavailable and disconnected:

- `GET /api/v1/devices`
- `/ws/v1/devices`

This makes API clients and skins treat the working integrated scale as absent.

## Observed behavior

With a Bengle connected:

```json
[
  {
    "name": "Bengle",
    "id": "<machine-id>",
    "state": "connected",
    "type": "machine",
    "available": true
  },
  {
    "name": "Bengle scale",
    "id": "bengle-internal-<machine-id>",
    "state": "disconnected",
    "type": "scale",
    "available": false
  }
]
```

At the same time, `GET /api/v1/machine/capabilities` includes `integratedScale`, and `/ws/v1/scale/snapshot` emits:

```json
{"status":"connected"}
{"timestamp":"...","weight":0.0,"weightFlow":0.0,"battery":100,"timerValue":null}
```

Reproduced on 2026-08-18 with a real Bengle.

## Expected behavior

While the Bengle and its virtual scale are connected, the device list should include:

```json
{
  "name": "Bengle scale",
  "id": "bengle-internal-<machine-id>",
  "state": "connected",
  "type": "scale",
  "available": true
}
```

The REST and WebSocket device snapshots should transition consistently when the Bengle connects or disconnects.

## Likely cause

`ConnectionManager._attachBengleVirtualScale()` creates the virtual scale and passes it directly to `ScaleController.connectToScale()`. It is not added to `DeviceController.devices`.

`DevicesHandler` builds both REST and WebSocket inventories from `DeviceController.devices` plus remembered devices via `buildAvailabilityDeviceList()`. Because the active virtual scale is missing from the live list, its remembered entry falls through as `disconnected` / `available:false`.

Relevant code:

- `lib/src/controllers/connection_manager.dart` — `_attachBengleVirtualScale`
- `lib/src/services/webserver/devices_handler.dart` — `_deviceList`, `_buildSnapshot`, `buildAvailabilityDeviceList`
- `lib/src/models/device/impl/bengle/bengle_virtual_scale.dart`

## Impact

Passione reproduced this by gating scale UI/telemetry on the devices API: Decaid was streaming weight successfully, but the skin displayed no scale because the inventory said it was disconnected. Clients can work around this by treating `/ws/v1/scale/snapshot` status as authoritative, but the device inventory should still truthfully represent the active integrated scale.

ODevStudio:
## Summary

What changed, and why?

- Include the scale currently owned by `ScaleController` in the REST and
  WebSocket device inventories when it is absent from discovery. This makes a
  connected `BengleVirtualScale` visible as connected and available.
- Rebuild WebSocket snapshots when the scale-controller connection state
  changes, including removal or remembered-offline state after disconnect.
- Reject REST and WebSocket connect/disconnect commands for inventory-only
  devices explicitly instead of reporting their listed IDs as not found.
- Update the REST, WebSocket, API, and device-management contracts for the
  clarified `available` and command-ownership meanings.

## Change Type (select all)

- [x] Bug fix
- [ ] Feature
- [ ] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [x] REST API / handlers
- [x] WebSocket API
- [x] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

Fixes #630

> **Remaining hardware validation before merge:** Verify REST and WebSocket inventory transitions with a physical Bengle and its virtual scale.

## Root Cause (if bug fix)

- Root cause: Device inventories were built only from `DeviceController`
  discovery, while a `BengleVirtualScale` can be connected and owned solely by
  `ScaleController`.
- Missing detection or guardrail: REST and WebSocket coverage did not exercise
  a real `BengleVirtualScale` absent from discovery or its command contract.
- Contributing context: The two controllers intentionally have different
  ownership responsibilities.

  - [ ] Unit test
- Target test or file: `test/devices_handler_test.dart` and
  `test/devices_ws_test.dart`.
- Scenario the test should lock in: A connected Bengle virtual scale appears
  once as connected and available when discovery does not contain it, while
  connect/disconnect commands return an explicit inventory-only error.
- If no new test added, why not: N/A; regression tests were added first.

- [x] Device docs updated: `doc/DeviceManagement.md` (if device flows changed)
- [ ] N/A - no docs affected

## Security Impact (required)

- New or changed REST endpoints? `Yes`, inventory inclusion changed.
- New or changed WebSocket topics? `Yes`, inventory inclusion changed.
- BLE/USB surface changed? `No`.
- Risk and mitigation: The APIs expose no new class of device data; they merge
  an already connected scale by ID, reject independent lifecycle commands, and
  integration tests cover de-duplication and offline transitions.

## User-Visible Changes

- `GET /api/v1/devices` and `/ws/v1/devices` now report an actively connected
  scale even when it did not originate from discovery.
- `available: true` now explicitly means present in discovery or actively
  connected. Remembered offline devices remain unavailable and disconnected.
- Inventory-only devices return REST 409 or an explicit WebSocket error for
  connect/disconnect because their lifecycle belongs to another controller.

- [x] `flutter test` - 3,181 passed, 1 skipped

- OS / platform tested: Windows test host; live app launched.
- Simulated devices? (`simulate=1`): `Yes`.
- What you personally verified and how: Computer Use confirmed the live Devices
  inventory renders MockBengle under machines and Mock Scale under scales; real
  Shelf and WebSocket handler paths remain covered by integration tests.
- Edge cases checked: Missing discovery entry, both command transports,
  disconnect, remembered-offline state, and duplicate prevention.
- What you did **not** verify: Physical Bengle behavior or command execution
  against Bengle hardware.

### Evidence

- [ ] Log snippets

Detailed evidence:

- Regression-first proof: before the implementation, the new REST test returned
  an empty list and the new WebSocket test timed out.
- Follow-up regression proof: the real Bengle command tests failed with REST
  404 and WebSocket `Device not found` before the inventory-only contract fix.
- `dart format --output=none --set-exit-if-changed lib/src/services/webserver/devices_handler.dart test/devices_handler_test.dart test/devices_ws_test.dart`
- `flutter test --no-pub test/connect_result_schema_test.dart test/devices_handler_test.dart test/devices_ws_test.dart` (61 passed)
- `flutter test --no-pub` with the package-matched QuickJS DLL on `PATH`
  (3,181 passed, 1 skipped)
- Flutter's default Windows path still selected CMake 3.20, but the unchanged
  source built with installed CMake 3.28 and Visual Studio 2022. Computer Use
  verified the simulated device inventory; changed REST and WebSocket paths
  remain exercised through real Shelf and WebSocket handler integration tests.

## Compatibility & Migration

- Backward compatible? `Yes`; existing objects and routes are unchanged. The
  newly exposed inventory-only IDs now return explicit 409 instead of 404.
- Database migration needed? `No`.
- Exact steps: None.

## Risks & Mitigations

- Risk: A controller-owned scale could be emitted twice or remain available
  after disconnect.
  - Mitigation: Merge by device ID and rebuild on scale-controller state
    changes; REST and WebSocket tests cover both cases.
- Risk: Clients may treat `available` as permission to send lifecycle commands.
  - Mitigation: Specs and docs separate presence from command ownership, and
    both command transports return an explicit inventory-only error.

## Contributor Responsibility

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -39,11 +39,12 @@ paths:
                     available:
                       type: boolean
                       description: >
-                        true if the device is currently present in discovery;
-                        false for a remembered device that is not present
-                        (reported with state "disconnected"). Remembered devices
-                        are those the user has connected to; they persist across
-                        restarts until forgotten via /api/v1/devices/forget.
+                        true if the device is currently present in discovery or
+                        is actively connected; false for a remembered device
+                        that is not present (reported with state "disconnected").
+                        Remembered devices are those the user has connected to;
+                        they persist across restarts until forgotten via
+                        /api/v1/devices/forget.
         "500":
           description: Internal Server Error
           content:
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -121,6 +121,8 @@ paths:
         Accepts deviceId via JSON body (preferred) or query parameter (backward compatible).
         Device IDs may contain special characters (e.g. colons in BLE MAC addresses,
         slashes in serial port paths), so the JSON body approach avoids URL encoding issues.
+        Controller-owned inventory devices such as Bengle's integrated virtual scale
+        are not independent command targets and return 409.
       tags: [Devices]
       parameters:
         - name: deviceId
@@ -154,11 +156,13 @@ paths:
         "404":
           description: "Device not found"
         "409":
-          description: "Connection conflicts with another operation or a stale selection"
+          description: "Connection conflicts with another operation, a stale selection, or an inventory-only device"
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/DeviceConnectResult"
+                oneOf:
+                  - $ref: "#/components/schemas/DeviceConnectResult"
+                  - $ref: "#/components/schemas/Error"
         "503":
           description: "Device transport currently unavailable"
           content:
@@ -177,6 +181,8 @@ paths:
       summary: Disconnect a specified device
       description: |
         Accepts deviceId via JSON body (preferred) or query parameter (backward compatible).
+        Controller-owned inventory devices such as Bengle's integrated virtual scale
+        follow their machine lifecycle and return 409.
       tags: [Devices]
       parameters:
         - name: deviceId
@@ -205,6 +211,12 @@ paths:
           description: "Missing deviceId"
         "404":
           description: "Device not found"
+        "409":
+          description: "Device is inventory-only and cannot be disconnected independently"
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
 
   /api/v1/devices/wifi:
     get:
--- a/assets/api/websocket_v1.yml
+++ b/assets/api/websocket_v1.yml
@@ -849,7 +849,8 @@ components:
             connected; false for a remembered device that is not present
             (reported with state "disconnected"). Remembered devices persist
             across restarts until forgotten via the REST
-            /api/v1/devices/forget endpoint.
+            /api/v1/devices/forget endpoint. Availability does not imply command
+            ownership; controller-owned devices may be inventory-only.
           example: true
 
     DeviceConnectResult:
@@ -890,7 +891,11 @@ components:
           description: The command to execute.
         deviceId:
           type: string
-          description: Device identifier. Required for connect and disconnect commands.
+          description: >
+            Device identifier. Required for connect and disconnect commands.
+            Controller-owned inventory devices such as Bengle's integrated
+            virtual scale reject both commands because their lifecycle follows
+            the machine.
         connect:
           type: boolean
           description: Whether to scan first and automatically fill missing machine and scale slots without replacing connected devices (scan command only).
--- a/assets/api/websocket_v1.yml
+++ b/assets/api/websocket_v1.yml
@@ -845,10 +845,11 @@ components:
         available:
           type: boolean
           description: >
-            true if the device is currently present in discovery; false for a
-            remembered device that is not present (reported with state
-            "disconnected"). Remembered devices persist across restarts until
-            forgotten via the REST /api/v1/devices/forget endpoint.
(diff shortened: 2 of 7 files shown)


ODevStudio:
[P2] The newly exposed virtual scale cannot be controlled through the same Devices API that reports it as connected.
_devicesForInventory() now appends the ScaleController-owned scale when it is missing from discovery, which correctly makes the Bengle virtual scale appear as state: "connected", available: true.
However, both REST control handlers still resolve devices exclusively through _controller.devices. Therefore:
GET /api/v1/devices can return bengle-internal-<id> as connected/available.
PUT /api/v1/devices/disconnect with that exact ID returns 404 Device not found.
The /ws/v1/devices disconnect command likewise returns Device not found.
connect has the same lookup discrepancy.
This matters because the PR explicitly broadens available:true from “present in discovery” to “present in discovery or actively connected,” while the public API still describes the device endpoints/channel generically as supporting device connect/disconnect operations.
I would not necessarily make the virtual scale independently disconnectable. BengleVirtualScale.disconnect() is intentionally a no-op and its connection state is inherited from the Bengle machine, so its lifecycle is clearly machine-owned.  A clean fix could instead make that contract explicit—for example, document that integrated/controller-owned devices are inventory-only and cannot receive device connect/disconnect commands, or add an explicit capability/ownership field. Otherwise clients can reasonably act on a device the API says is connected only to have the same API reject its ID.
Test gap
The new WebSocket regression test uses a generic TestScale, manually changes its connection state, and verifies inventory removal. It does not exercise the actual BengleVirtualScale, whose state is coupled to the machine, nor does it test a disconnect command against the newly surfaced ID.  I would add at least the command-contract case chosen above.

tadelv (approved):

PR #644, opened 2026-08-20
feat(debug): add DE1 calibration controls

### LINKED ISSUE #620: Native debug: add DE1 calibration controls
Depends on #613.

## Goal

Extend Decaid's native debug view to inspect and modify DE1 calibration using the calibration API introduced by #613.

This is a developer/debugging interface, not a guided end-user calibration workflow.

## Requirements

- Add a calibration section to the native debug view.
- Support all calibration targets exposed by the machine API:
  - flow
  - pressure
  - temperature
- Read and display the machine's current calibration values.
- Read and display factory calibration values.
- Allow calibration values to be written back to the machine.
- Refresh/re-read the calibration after a successful write so the displayed state reflects the machine's authoritative value.
- Show read/write failures using the existing native-debug error/status conventions.
- Use only the typed machine API from #613; there should be no A012 packet construction or BLE protocol logic in the debug UI.

## Non-goals

- Guided pressure or temperature calibration procedures.
- General user-facing calibration settings.
- Duplicating calibration protocol logic from the device layer.

## Acceptance criteria

From the native debug view, a developer can inspect current and factory calibration and write calibration values for flow, pressure and temperature, with all communication going through the API introduced by #613.

ODevStudio:
## Summary

What changed, and why?

- Add native debug controls for reading the current and factory flow, pressure,
  and temperature calibration values through the existing typed `De1Interface`.
- Let a technician submit reported and measured values for each target, then
  re-read the machine so the UI shows the value the DE1 accepted.
- Keep the calibration state alive while scrolling and make the wide debug view
  scrollable so the new controls remain usable at narrow and wide widths.
- Wait for machine connection setup before creating the calibration controls; the shared A012 codec rejects flow or pressure divisors that encode as zero in Q16.16 for every caller.

## Change Type (select all)

- [ ] Bug fix
- [x] Feature
- [ ] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [ ] REST API / handlers
- [ ] WebSocket API
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [x] UI / Flutter widgets
- [ ] CI / build / infra
- [ ] Docs / specs

## Linked Issues

Fixes #620

> **Remaining hardware validation before merge:** Validate calibration reads and writes on a physical DE1. Incorrect calibration can affect machine readings.

## Root Cause (if bug fix)

N/A - this exposes existing typed calibration operations in the native debug UI.

N/A - feature coverage lives in
`test/debug_feature/debug_item_details_view_test.dart` and locks in reads,
writes, connection gating, refresh, failures, and
responsive layouts.

- [x] N/A - no docs affected

## Security Impact (required)

- New or changed REST endpoints? `No`.
- New or changed WebSocket topics? `No`.
- BLE/USB surface changed? `Yes`; existing typed calibration reads and writes
  are now reachable from the native debug UI.
- Risk and mitigation: Bad calibration can affect machine readings. Controls
  remain in the debug screen, wait for machine setup, use the typed `De1Interface`, validate zero wire divisors once in the shared codec, and re-read the accepted value after writes.

## User-Visible Changes

- The DE1 debug screen shows current and factory flow, pressure, and
  temperature calibration values and can write corrections.
- Both inputs start at the current value; an unchanged submission is a no-op.
- Calibration controls stay unavailable until machine connection setup
  completes, and unsafe zero flow/pressure divisors are rejected by the shared machine codec.

- [x] `flutter test` - 3,187 passed, 1 skipped

- OS / platform tested: Windows test host; live app launched.
- Simulated devices? (`simulate=1`): `Yes`.
- What you personally verified and how: Computer Use opened MockDe1 details,
  waited for live calibration reads, and confirmed the flow, pressure, and
  temperature controls fit and scroll coherently at 1268 by 795.
- Edge cases checked: Delayed connection, zero and sub-LSB flow/pressure
  divisors, temperature zero, unchanged values, invalid input, read/write
  failures, post-write refresh, scrolling, and retained state.
- What you did **not** verify: Calibration reads or writes on a physical DE1.

### Evidence

- [ ] Log snippets

Detailed evidence:

- Regression-first codec coverage checks zero and sub-LSB flow/pressure divisors while permitting temperature zero. Widget coverage checks delayed connection setup, all current and factory
  reads, write arguments, post-write refresh, failures, and responsive layouts.
- `flutter test --no-pub test/debug_feature/debug_item_details_view_test.dart`
  (11 passed).
- Widget, calibration codec/transport/API, Bengle scale calibration, and REST
  calibration tests passed together (81 tests).
- `flutter analyze --no-pub` (no issues).
- `flutter test --no-pub` with the package-matched QuickJS DLL on `PATH`
  (3,187 passed, 1 skipped).
- No DE1 was available for a hardware calibration smoke test.
- Review follow-up commit c04d8594 moved zero-divisor validation into De1CalibrationCodec. The codec and widget suites passed 28 tests, and flutter analyze was clean.
- The Windows full suite reached 3,185 passed and 1 skipped; the two failures were the existing CRLF-sensitive AsyncAPI assertion and the ignored bundled-skin fixture not entering Flutter's cached asset bundle.

## Compatibility & Migration

- Backward compatible? `Yes`; the change adds controls to an existing debug
  surface and reuses the current typed API.
- Database migration needed? `No`.
- Exact steps: None.

## Risks & Mitigations

- Risk: An incorrect technician entry can miscalibrate the DE1.
  - Mitigation: Require physical-DE1 validation before merge; wait for connection setup,
    validate zero wire divisors once in the shared codec, retain typed writes, and re-read after writes.

## Contributor Responsibility

### DIFF
--- a/lib/src/debug_feature/calibration_debug_card.dart
+++ b/lib/src/debug_feature/calibration_debug_card.dart
@@ -13,8 +13,6 @@ class CalibrationDebugCard extends StatefulWidget {
 
 class _CalibrationDebugCardState extends State<CalibrationDebugCard>
     with AutomaticKeepAliveClientMixin {
-  static const _fixedPointScale = 65536.0;
-
   final _reportedControllers = {
     for (final target in De1CalibrationTarget.values)
       target: TextEditingController(),
@@ -201,12 +199,6 @@ class _CalibrationDebugCardState extends State<CalibrationDebugCard>
       _showStatus('Enter valid reported and measured values');
       return;
     }
-    if (target != De1CalibrationTarget.temperature &&
-        (reported * _fixedPointScale).round() == 0) {
-      _showStatus('Reported value must be non-zero for flow and pressure');
-      return;
-    }
-
     setState(() => _writing = {target});
     try {
       await widget.machine.writeCalibration(
--- a/lib/src/debug_feature/calibration_debug_card.dart
+++ b/lib/src/debug_feature/calibration_debug_card.dart
@@ -13,6 +13,8 @@ class CalibrationDebugCard extends StatefulWidget {
 
 class _CalibrationDebugCardState extends State<CalibrationDebugCard>
     with AutomaticKeepAliveClientMixin {
+  static const _fixedPointScale = 65536.0;
+
   final _reportedControllers = {
     for (final target in De1CalibrationTarget.values)
       target: TextEditingController(),
@@ -197,6 +199,11 @@ class _CalibrationDebugCardState extends State<CalibrationDebugCard>
         !reported.isFinite ||
         !measured.isFinite) {
       _showStatus('Enter valid reported and measured values');
+      return;
+    }
+    if (target != De1CalibrationTarget.temperature &&
+        (reported * _fixedPointScale).round() == 0) {
+      _showStatus('Reported value must be non-zero for flow and pressure');
       return;
     }
 
--- a/lib/src/debug_feature/calibration_debug_card.dart
+++ b/lib/src/debug_feature/calibration_debug_card.dart
@@ -1,0 +1,288 @@
+import 'package:flutter/material.dart';
+import 'package:reaprime/src/models/device/de1_interface.dart';
+import 'package:shadcn_ui/shadcn_ui.dart';
+
+class CalibrationDebugCard extends StatefulWidget {
+  const CalibrationDebugCard({super.key, required this.machine});
+
+  final De1Interface machine;
+
+  @override
+  State<CalibrationDebugCard> createState() => _CalibrationDebugCardState();
+}
+
+class _CalibrationDebugCardState extends State<CalibrationDebugCard>
+    with AutomaticKeepAliveClientMixin {
+  final _reportedControllers = {
+    for (final target in De1CalibrationTarget.values)
+      target: TextEditingController(),
+  };
+  final _measuredControllers = {
+    for (final target in De1CalibrationTarget.values)
+      target: TextEditingController(),
+  };
+
+  Map<De1CalibrationTarget, De1Calibration> _current = const {};
+  Map<De1CalibrationTarget, De1Calibration> _factory = const {};
+  Set<De1CalibrationTarget> _writing = const {};
+  var _loading = true;
+
+  @override
+  void initState() {
+    super.initState();
+    _refreshAll();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    super.build(context);
+    return ShadCard(
+      title: Row(
+        children: [
+          const Expanded(child: Text('Calibration')),
+          Tooltip(
+            message: 'Refresh calibration',
+            child: ShadButton.ghost(
+              size: ShadButtonSize.sm,
+              onPressed: _loading || _writing.isNotEmpty ? null : _refreshAll,
+              child: const Icon(LucideIcons.refreshCw, size: 16),
+            ),
+          ),
+        ],
+      ),
+      child: _loading
+          ? const Center(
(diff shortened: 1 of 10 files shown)


ODevStudio:
High — initial calibration reads race machine connection (calibration_debug_card.dart:33). The normal entry path is the debug device list’s Connect button. De1DebugView.initState() then starts machine.onConnect() without awaiting it, while the newly mounted CalibrationDebugCard immediately calls _refreshAll() from its own initState. On real transports, writeWithResponse() explicitly rejects requests unless the connection state is already connected; BLE setup also installs the calibration subscription during the asynchronous connection sequence. This means the six initial calibration reads can run before connection/setup completes, producing a “Calibration read failed” dialog and leaving the card unavailable until the technician manually retries.
Fix: gate creation/refresh of the card on completion of the onConnect() future, rather than merely connection-state readiness. Add a test double whose onConnect() is delayed and whose calibration methods fail before it completes.
Medium-high — flow/pressure accept a zero divisor (calibration_debug_card.dart:195). Validation only requires finite doubles, so reported = 0 is accepted and sent. The PR’s own calibration test model defines flow/pressure correction as previous * measured / reported, making zero explicitly invalid; the lower-level Q16.16 codec also accepts and encodes zero. A sufficiently tiny nonzero value can additionally quantize to zero on the wire. For calibration writes, this is an unsafe input to allow through the UI.
Fix: for flow and pressure, reject reported values that are zero—or close enough to encode as zero in Q16.16—and cover that case with a widget test. Temperature can remain additive and therefore permit zero.

lib/src/debug_feature/calibration_debug_card.dart
```diff
@@ -0,0 +1,287 @@
...
+      setState(() {
+        _current = {..._current, target: current};
+        _writing = const {};
+      });
+      _setInputValues(current);
+      _showStatus('${_targetLabel(target)} calibration updated');
+    } catch (error) {
+      if (!mounted) return;
+      setState(() => _writing = const {});
+      _showFailure('Calibration refresh failed', error);
```
tadelv, line 239: After `writeCalibration()` succeeds, this re-read can still fail. In that case the machine may already contain a new calibration, but `_current[target]` remains the old value and the UI continues to label it as `Current` after the dialog is dismissed. A failed manual refresh has the same stale-state problem for the whole card. Please invalidate/mark the affected displayed value as unavailable/stale when verification fails (and ideally prevent another write for that target until a successful refresh), so we never present a known-non-authoritative value as current.

lib/src/debug_feature/calibration_debug_card.dart
```diff
@@ -0,0 +1,287 @@
...
+      ],
+    );
+  }
+
+  Future<void> _refreshAll() async {
+    if (mounted) {
+      setState(() => _loading = true);
+    }
+    try {
+      final reads = await Future.wait([
```
tadelv: `readCalibration()` serializes A012 requests and each read can take the 4s calibration timeout. Creating all six futures here therefore pre-queues six operations, and `Future.wait` (without `eagerError`) won't report the first failure until the whole queue drains — roughly 24s when A012 is unavailable. Please issue these reads sequentially and stop on the first failure (or otherwise avoid pre-queuing the full batch). `eagerError: true` alone would surface the error sooner but would still leave the remaining doomed operations occupying the calibration queue.

tadelv (changes requested):
Two issues to address before merge. The overall implementation and calibration semantics look sound; these are both failure-path problems in the debug UI.

tadelv (approved):

PR #645, opened 2026-08-20
feat(plugins): reconcile missed shot uploads

### LINKED ISSUE #603: Make shot-upload plugin eventually reconcile the full local shot library
## Goal

Make `assets/plugins/shot-upload.reaplugin` best-effort and eventually consistent with the user's Decent backend shot history: when auto-upload is enabled and the account/backend are reachable, **all eligible local shots should eventually be uploaded**, not only shots recorded while the plugin is active.

This should mirror the intent of de1app's current `shot_upload` plugin, which has both a live upload fast path and a throttled backlog drain.

## Current gap

The Decaid plugin currently:

- reacts to `shotStored` and immediately calls `autoUpload(id)`;
- retries that one POST up to 3 times with a fixed delay;
- persists only `lastUploadedShot` as a fast-path dedup value;
- marks a successfully uploaded record with `annotations.extras.uploaded_to_decent`.

There is no startup/idle/periodic reconciliation of the existing shot library and no durable queue/backlog scan.

Consequently, eligible shots can remain permanently absent from the backend when they were:

- recorded before the plugin was enabled/installed;
- recorded while auto-upload was disabled and later enabled;
- missed because Decaid/plugin execution was interrupted;
- not uploaded because the network/backend was unavailable beyond the immediate retries.

Taking another shot does not repair those older gaps.

By contrast, de1app's `de1plus/plugins/shot_upload/plugin.tcl` explicitly implements an uploaded marker/ledger, attempt tracking, backoff, and a backlog drain triggered while the machine is asleep, specifically to backfill offline and pre-enable shots.

## Suggested approach

Keep `shotStored` as the low-latency fast path, but add a **low-priority reconciliation worker**:

1. Trigger reconciliation when appropriate, e.g. after plugin load/account enable, when entering an idle/safe machine state, and after a failed live upload. A modest periodic idle trigger is also reasonable so missed events eventually heal.
2. Enumerate local shots incrementally/paginated rather than loading the entire library into memory.
3. Treat `annotations.extras.uploaded_to_decent` as the durable per-shot success marker. Select eligible shots without that marker and process only a small bounded batch per drain opportunity.
4. Upload sequentially/throttled so catch-up work cannot interfere with brewing, UI responsiveness, or normal device traffic. Stop/pause the drain when the machine becomes active.
5. On successful upload, persist the existing `uploaded_to_decent` marker immediately. Retries must be idempotent from the user's perspective.
6. Handle transient network/5xx failures with backoff and resume on a later reconciliation opportunity rather than permanently abandoning the shot after one event's retry window. Permanent 4xx/data errors should not create a tight retry loop; persist/report enough failure state to avoid repeatedly hammering the same bad record.
7. Re-scan/reconcile over time instead of relying solely on an in-memory queue. This is what provides eventual consistency across app restarts and lost events.

### Historical machine identity

The current live path injects identity from `/machine/info`, i.e. the currently connected machine. A backlog uploader must **not attribute arbitrary historical shots to whichever machine happens to be connected during reconciliation**. Use machine identity stored with the shot at capture time if available; if Decaid's shot API does not currently expose enough historical identity/provenance, extend that API/model first (or otherwise skip records whose ownership cannot be established safely).

## Acceptance criteria

- Enabling auto-upload with an existing local library eventually uploads all eligible, not-yet-uploaded shots without requiring new shots to be taken.
- Shots missed during an offline/backend outage are retried and eventually uploaded after connectivity recovers.
- Restarting Decaid during catch-up does not lose reconciliation progress; the next run discovers remaining unmarked shots again.
- Already uploaded shots are not needlessly re-uploaded during normal reconciliation.
- Reconciliation is bounded/throttled and does not compete with an active espresso workflow.
- A permanently bad/rejected shot cannot cause a busy retry loop or block all later shots from converging.
- Historical shots are uploaded with their correct captured machine identity/provenance, never the identity of an unrelated currently connected machine.

## Reference behavior

`decentespresso/de1app/de1plus/plugins/shot_upload/plugin.tcl` documents and implements this as a live post-shot path plus a Sleep-only backlog drain for offline/pre-enable shots, with uploaded markers/ledger, per-shot attempt tracking and backoff. Decaid should implement the equivalent semantics using its own shot database/API and plugin lifecycle rather than copying the Tcl mechanism literally.

ODevStudio:
## Summary

What changed, and why?

- Capture machine provenance at shot start and reconcile missed uploads in bounded newest-first batches while the machine is safe.
- Classify account-consent denial for plugins and pause reconciliation without repeated prompts.
- Preserve 0.2.0 legacy-history behavior: records without `provenanceStatus` may use the currently connected real machine; `captured` uses only captured identity; `unavailable` blocks automatic fallback.
- Allow explicit manual upload to override `unavailable` provenance while preserving simulated-machine safeguards.
- Preserve legacy `upload_skipped: mock-device` exclusions, stop an active reconciliation immediately on consent pause, and avoid repeat remote POSTs when local upload marking fails.
- Suppress genuinely permanent remote rejections for the plugin session when local annotation fails, while allowing manual retry and one later idempotent retry after restart.
- Make the 0.2.0 to 0.2.1 `AutoUpload`/`DrainHistory` migration semantics explicit.

## Linked Issue

Fixes #603

## Verification

- `dart format lib test`
- `flutter analyze --no-pub` (clean)
- Focused provenance and reconciliation suites: 47 passed
- Full Windows suite: 3,285 passed and 1 skipped. Seven unrelated failures remain: one existing CRLF-sensitive AsyncAPI assertion and six WebUI tests unable to bind occupied port 3000.
- No production-backend or real-machine upload was run.

## Impact

- Enabling `AutoUpload` in 0.2.1 also enables backlog reconciliation; the removed 0.2.0 `DrainHistory` setting no longer gates it.
- Recent shots reconcile before ancient backlog while the persisted offset still converges over older records.
- Consent denial or timeout pauses automatic reconciliation until upload is explicitly re-enabled or the plugin reloads, including an already-running pass.
- A successful remote upload remains successful if its local annotation PUT fails; the plugin suppresses repeat POSTs for that shot for the rest of the session.
- Permanent remote rejections receive the same session suppression if their annotation PUT fails. Manual retry bypasses it; after restart, one later idempotent retry is possible.
- Legacy records with absent provenance status retain the 0.2.0 connected-real-machine fallback and its possible multi-machine misattribution tradeoff. Captured identity is authoritative, and newly recorded capture failures do not fall back automatically.
- Plugin documentation and the REST schema describe the new provenance states.

## Contributor Responsibility


ODevStudio:
P1 — a failed machine snapshot can persist stale machine identity. In lib/src/controllers/de1_state_manager.dart around 602–617, the code reads the currently connected DE1 and replaces baseWorkflow.machine only when that read succeeds. On failure it persists baseWorkflow unchanged. But Workflow.machine can now contain serialNumber loaded through Workflow.fromJson, and /api/v1/workflow accepts that structure into the current workflow. The reconciler then treats any workflow.machine.serialNumber as trustworthy capture-time provenance.
That means if the fresh machine read fails while the active workflow contains identity from a previously loaded workflow/shot, this new shot can inherit the old serial and later be uploaded under the wrong machine—the exact failure #603 is intended to prevent. I would capture an immutable machine snapshot when the shot sequencer starts, or at minimum explicitly clear provenance when fresh identity cannot be established. Note that the current Workflow.copyWith(machine: null) cannot clear the field because it uses machine ?? this.machine. Add a regression test with a stale serial in baseWorkflow and a failing connectedDe1() read; the resulting shot must have no trusted serial and must not reconcile.
P1 — disabling automatic upload does not stop already-scheduled POST retries. reconcile() checks state.autoUpload, unloaded, and machine safety between records, but once uploadShot() enters postShot(), the three-attempt retry loop never checks any of those conditions.
Reproduction: the first backend POST returns 503; before the retry timer fires, dispatch settingsUpdated with AutoUpload: false. The retry loop will still issue the second—and potentially third—POST, so a later retry can upload the shot after the user has disabled uploads. The same problem applies when stateUpdate changes to espresso: additional retry attempts can start while brewing even though the PR says reconciliation pauses when brewing starts. postShot() should receive/check a continuation predicate before every new POST and after each retry delay. Add tests that inject a transient failure, disable auto-upload or enter espresso before the next timer, and assert only the first request occurred.

tadelv:
@ODevStudio  please rebase on main and verify there is still anything useful to pull out from this PR. 
ref commit: https://github.com/decentespresso/decaid/commit/0ea60340c9cb8642c337e841d331cfee486d5235

tadelv (changes requested):
Requesting changes before merge. The overall reconcile architecture is preferable to the old DrainHistory approach, and the earlier stale-machine and retry-after-disable findings appear fixed, but there are two remaining issues I consider merge blockers / near-blockers:

**P1 — rebase/merge current `main` and account for the new account-proxy consent contract.** `main` has moved since this PR's last merge and `DecentProxyService` now requires `requireConsent`; the new reconciliation test harness still constructs it using the old API, so the currently-green CI is no longer representative of current `main`.

More importantly, consent denial/timeout currently propagates through `host.decentProxy()` as an exception. `postShot()` treats thrown proxy failures as transient unless they carry its own `permanent` flag, so a denied or timed-out consent request will be retried three times and then scheduled for future reconciliation again. A timeout/non-decision is not persisted by the consent layer, which can therefore lead to repeated native consent prompts on later reconciliation passes. Please merge/rebase current `main`, expose/classify consent denial in a stable way, and make reconciliation pause rather than treating it like a transport failure. Add tests for explicit denial and consent timeout/non-decision.

**P2 — reconcile recent history before ancient backlog.** The current reconciler scans `order=asc`, at most 5 pages / 100 rows per run, and normally waits five minutes between scans. On a migrated library with thousands of old records, recent missed shots can therefore take a very long time to surface. This has already been observed with 2000+ shot histories. Prefer newest-first traversal (ideally using a stable timestamp/id cursor rather than relying indefinitely on offset pagination), while still allowing eventual convergence of the full backlog.

Please also make the 0.2.0 -> 0.2.1 settings migration semantics explicit in tests: 0.2.0 had separate `AutoUpload` and `DrainHistory` opt-ins, whereas 0.2.1 removes `DrainHistory` and makes backlog reconciliation follow `AutoUpload`. That may be the intended product change, but it should be deliberately specified rather than implicit.

Once these are addressed, I expect this to be close to approval.

tadelv:
One additional compatibility requirement after comparing the old `DrainHistory` behavior with the new reconciliation path:

**Please preserve legacy-history upload when capture-time machine provenance is absent.** The new `workflow.machine` provenance is the right primary source and should always win when present. However, shots created before this change generally have no `workflow.machine.serialNumber`, and #645 currently makes those records permanently ineligible for reconciliation.

For those legacy records only, please fall back to the previous behavior: use the currently connected real machine's identity, subject to the same no-machine / simulated-machine safeguards as 0.2.0. Do not use the fallback when captured provenance exists.

In other words:

- captured machine identity present -> use it;
- captured identity absent -> fall back to the currently connected real machine;
- captured identity present but invalid/mock -> do not silently replace it with the connected machine.

This keeps the correctness improvement for all newly recorded shots while avoiding a regression where existing Decaid history can no longer be uploaded at all. The fallback retains the same possible multi-machine misattribution tradeoff that users already had with `DrainHistory`, rather than silently removing that capability during the 0.2.0 -> 0.2.1 migration.

Please cover the fallback and precedence rules with tests.

tadelv:
One remaining blocker from the legacy-fallback change:

**Shots already marked `annotations.extras.upload_skipped = "mock-device"` must remain permanently excluded from reconciliation.**

0.2.0 deliberately wrote that marker when a shot came from the simulator, and the old drain explicitly skipped those records so they could never later be attributed to a real connected DE1. With the new legacy fallback, ignoring that marker means an old simulated shot with no captured `workflow.machine.serialNumber` can become eligible, fall back to the currently connected real machine, and be uploaded under the wrong serial.

Please preserve the existing safety semantics by making reconciliation reject/skip `upload_skipped: "mock-device"` before attempting any connected-machine fallback. More generally, if `upload_skipped` is intended as a durable no-upload marker, it would be reasonable to honor it consistently rather than only special-case this value.

Please add a regression test covering:

- legacy shot with no captured serial;
- `annotations.extras.upload_skipped = "mock-device"`;
- a real machine connected;
- reconciliation performs zero proxy uploads and does not resurrect the shot through the legacy fallback.

Once this is fixed, the other changes from the previous review look resolved.

tadelv (approved):

PR #646, opened 2026-08-20
feat(api): list bean batches across beans

### LINKED ISSUE #564: N+1 query pattern: no bulk "list all batches" endpoint forces clients into one GET per bean
## N+1 query pattern: no bulk "list all batches" endpoint forces clients into one GET per bean

**Summary:** The batches API is per-bean only (`GET /api/v1/beans/<beanId>/batches`), so any client that needs all batches must issue one request per bean — an N+1 pattern. Observed from the WorkFlow-Skin, where this is a significant contributor to request bursts on low-power hosts.

### Context

In [`beans_handler.dart`](https://github.com/decentespresso/decaid/blob/main/lib/src/services/webserver/beans_handler.dart) the only ways to list batches are:

- `GET /api/v1/beans/<beanId>/batches` — batches for a single bean
- `GET /api/v1/bean-batches/<id>` — a single batch

There is no way to fetch all batches (or batches for a set of beans) in one call.

Clients that show a bags/inventory view (e.g. WorkFlow-Skin's [`useReaData.ts`](https://github.com/Sabotage1/WorkFlow-Skin/blob/main/skin/workflow-skin/src/state/useReaData.ts) → `beanList.map(api.listBatches)`) therefore do:

```
GET /api/v1/beans/{bean1}/batches
GET /api/v1/beans/{bean2}/batches
...
GET /api/v1/beans/{beanN}/batches
```

### Measured impact

With 9 beans, every WorkFlow-Skin refresh cycle (every 30s) fires 9 batch requests alongside ~21 other GETs, all at once. On a low-power host (Orange Pi, 4-core ARM, 2 GB RAM) serving the decaid API:

- **4,131 GETs** over ~3h47m, **peak 95 requests in a single second**
- Server process pinned at **~100–108% CPU** while a skin tab was open
- Each `_getBatches` call is a separate `getBatchesForBean` DB query, so N beans = N SQLite queries per refresh

304s (via `jsonOkConditional`) avoid response body cost, but each request still costs a round trip + DB query + JSON serialization.

### Suggestions

1. **Add a bulk endpoint**, e.g. `GET /api/v1/bean-batches?includeArchived=false` returning all batches (optionally filterable by `beanIds=`), so a skin can replace the N+1 loop with a single request.
2. **Or embed batches in `GET /api/v1/beans`** behind a query flag (e.g. `?includeBatches=true`), which also lets clients skip a separate call entirely.
3. Either way, the per-bean endpoint stays for cases that genuinely need one bean's batches.

Related: filed against WorkFlow-Skin as [Sabotage1/WorkFlow-Skin#2](https://github.com/Sabotage1/WorkFlow-Skin/issues/2) (the ~30-GET burst that makes this visible).

ODevStudio:
## Summary

What changed, and why?

- Add `GET /api/v1/bean-batches` so inventory clients can fetch batches for
  every bean with one request instead of one request and SQLite query per bean.
- Use one Drift query, preserve the existing default archived filter, accept
  `includeArchived=true`, and support the same conditional `ETag` behavior as
  the existing bean and per-bean batch collection routes.
- Exclude batches whose parent bean is archived from the default collection,
  matching the motivating beans-then-batches workflow.
- Keep the per-bean and single-batch routes unchanged. No speculative
  multi-ID filter or embedded-bean response was added.

## Change Type (select all)

- [ ] Bug fix
- [x] Feature
- [ ] Docs
- [ ] Security hardening
- [ ] Chore / infra

- [x] REST API / handlers
- [ ] WebSocket API
- [ ] Scale / weight / flow
- [ ] WebUI skins
- [ ] Plugins / JS runtime
- [ ] UI / Flutter widgets
- [x] Storage / Drift database
- [ ] CI / build / infra
- [x] Docs / specs

## Linked Issues

Fixes #564

> **Remaining validation before merge:** Maintainer review of the additive endpoint and a live API smoke with `curl`. No hardware is required.

## Root Cause (if bug fix)

N/A - this adds a bulk read route for existing bean-batch data.

N/A - feature coverage in `test/webserver/beans_handler_test.dart` and
`test/database/bean_dao_test.dart` locks in cross-bean results, archived
batch and parent-bean filtering, one-query behavior, and conditional GET.

- [ ] N/A - no docs affected

## Security Impact (required)

- New or changed REST endpoints? `Yes`; adds read-only
  `GET /api/v1/bean-batches`.
- New or changed WebSocket topics? `No`.
- BLE/USB surface changed? `No`.
- Risk and mitigation: The route exposes only batch records already available
  through per-bean routes, remains read-only, excludes archived batches and
  parent beans by default, and supports `ETag` validation.

## User-Visible Changes

- API clients can fetch bean batches across all beans with one request instead
  of issuing one request per bean.
- Existing per-bean and single-batch routes and response objects are unchanged.

- [x] `flutter test` - 3,181 passed, 1 skipped

- OS / platform tested: Windows test host; live app did not launch.
- Simulated devices? (`simulate=1`): `No`; CMake stopped the build first.
- Real hardware? (DE1/Bengle/scale): `No`, not applicable.
- What you personally verified and how: Real Shelf handler and Drift DAO paths
  through integration tests.
- Edge cases checked: Archived batches, archived parent beans,
  `includeArchived=true`, cross-bean ordering, `ETag`, and `304 Not Modified`.
- What you did **not** verify: A live curl request against a running app.

### Evidence

- [ ] Log snippets

Detailed evidence:

- Regression-first handler tests failed with `404` before the route existed,
  then passed with cross-bean results, archived filtering, and `304` handling.
- The archived-parent regression exposed the inconsistent default collection
  before the follow-up and passes at the DAO and API-handler levels afterward.
- `flutter test test/webserver/beans_handler_test.dart test/database/bean_dao_test.dart` (36 passed)
- `flutter analyze` (no issues)
- `flutter test` with the package-matched QuickJS DLL on `PATH` and the
  repository's temporary LF normalization for the line-ending-sensitive
  AsyncAPI assertion (3,181 passed, 1 skipped)
- `git diff --cached --check` before commit (clean)
- Attempted `flutter run --no-pub -d windows --dart-define=simulate=1` for a
  live curl smoke. The local build stopped before launch because Flutter
  selected CMake 3.20 while the Firebase SDK requires CMake 3.22 or newer.

## Compatibility & Migration

- Backward compatible? `Yes`; this is an additive route.
- Database migration needed? `No`; the existing `beanBatches` table is queried.
- Exact steps: None.

## Risks & Mitigations

- Risk: A bulk collection could add database or response overhead for large
  inventories.
  - Mitigation: Use one Drift query, keep archived batches and parent beans
    excluded by default, and support conditional requests with `ETag`.

## Contributor Responsibility

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -2166,6 +2166,40 @@ paths:
         "500":
           description: Internal Server Error
 
+  /api/v1/bean-batches:
+    get:
+      summary: List all bean batches
+      description: Returns batches across all beans, optionally including archived ones.
+      tags: [Beans]
+      parameters:
+        - name: includeArchived
+          in: query
+          required: false
+          schema:
+            type: boolean
+            default: false
+          description: Include archived batches
+      responses:
+        "200":
+          description: Array of batches
+          headers:
+            ETag:
+              schema:
+                type: string
+              description: |
+                Strong entity tag derived from the response body. Echo back as
+                `If-None-Match` on subsequent requests for conditional GET.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/BeanBatch"
+        "304":
+          description: Not Modified — client's `If-None-Match` matched the current ETag.
+        "500":
+          description: Internal Server Error
+
   /api/v1/bean-batches/{id}:
     get:
       summary: Get a batch by ID
--- a/doc/Api.md
+++ b/doc/Api.md
@@ -224,6 +224,7 @@ supplied values replace them. Explicit `null` for non-nullable fields returns
 | DELETE | `/api/v1/beans/:id` | Delete bean | |
 | GET | `/api/v1/beans/:id/batches` | List batches for a bean | |
 | POST | `/api/v1/beans/:id/batches` | Create batch | |
+| GET | `/api/v1/bean-batches` | List batches across all beans | |
 | GET | `/api/v1/bean-batches/:id` | Get batch | |
 | PUT | `/api/v1/bean-batches/:id` | Update batch | |
 | DELETE | `/api/v1/bean-batches/:id` | Delete batch | |
--- a/lib/src/services/database/daos/bean_dao.dart
+++ b/lib/src/services/database/daos/bean_dao.dart
@@ -66,6 +66,15 @@ class BeanDao extends DatabaseAccessor<AppDatabase> with _$BeanDaoMixin {
     return (delete(beans)..where((b) => b.id.equals(id))).go();
   }
 
+  Future<List<BeanBatche>> getAllBatches({bool includeArchived = false}) {
+    final query = select(beanBatches);
+    if (!includeArchived) {
+      query.where((b) => b.archived.equals(false));
+    }
+    query.orderBy([(b) => OrderingTerm.desc(b.updatedAt)]);
+    return query.get();
+  }
+
   Future<List<BeanBatche>> getBatchesForBean(
     String beanId, {
     bool includeArchived = false,
--- a/lib/src/services/database/daos/bean_dao.dart
+++ b/lib/src/services/database/daos/bean_dao.dart
@@ -69,7 +69,12 @@ class BeanDao extends DatabaseAccessor<AppDatabase> with _$BeanDaoMixin {
   Future<List<BeanBatche>> getAllBatches({bool includeArchived = false}) {
     final query = select(beanBatches);
     if (!includeArchived) {
-      query.where((b) => b.archived.equals(false));
+      final activeBeans = selectOnly(beans)
+        ..addColumns([beans.id])
+        ..where(beans.archived.equals(false));
+      query.where(
+        (b) => b.archived.equals(false) & b.beanId.isInQuery(activeBeans),
+      );
     }
     query.orderBy([(b) => OrderingTerm.desc(b.updatedAt)]);
     return query.get();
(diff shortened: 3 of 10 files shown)


ODevStudio:
Reviewed PR #646. I found one correctness/behavior issue; otherwise the implementation is clean and CI is green.

[P2] Exclude batches belonging to archived beans by default

getAllBatches() filters only beanBatches.archived. It does not check whether the parent bean itself is archived.

That differs from the N+1 workflow this endpoint is specifically intended to replace. The normal flow is effectively:

GET /beans                 # archived beans omitted
for each returned bean:
    GET /beans/:id/batches

GET /beans defaults to excluding archived beans, so batches belonging to those beans never enter that result set.  Issue #564 explicitly describes this beanList.map(api.listBatches) pattern as the motivating workload.

With this PR, the equivalent sequence can produce inconsistent collections:

bean A
  archived = true

batch B
  beanId = A
  archived = false

GET /api/v1/beans
=> []

GET /api/v1/bean-batches
=> [B]

So a client fetching both default collections receives a batch whose beanId has no corresponding bean. The new test covers an archived batch, but not an archived parent bean.

I would make the bulk DAO a single join against beans and, when includeArchived == false, require both the batch and its parent bean to be non-archived. Alternatively, if returning batches from archived beans is intentional, the endpoint contract should state that explicitly because it is not a drop-in replacement for the motivating N+1 call pattern.

Everything else: routing, storage abstraction, Drift mapping, ETag handling, OpenAPI/docs, and the basic archived-batch tests look consistent with the existing implementation. The GitHub Actions run passed the Linux build smoke plus format/analyze/full Flutter tests. The PR is still correctly marked draft pending the stated live API smoke.

Review disposition: request changes for the archived-parent case; no other blocking findings.

tadelv (approved):

PR #649, opened 2026-08-20
fix(skins): make Dashboard exit clear and reliable

### LINKED ISSUE #427: [UX] Better explanation of dashboard purpose and how to reach it
Apart from the default "streamline", none of the other bundled skins seem to have settings to change the skin. 

~~This means once a user switches to any of the others they are trapped in that skin unless they know how to use the API to switch skins (set new default skin, stop and start web ui server).~~

~~I'm not sure what the best solution for this would be but there should be a somewhat end user friendly way to "escape" a skin even if that skin doesn't provide settings for it~~

/edit: Turns out this is what the "use system back button to go to dashboard" message is about.
Unfortunately as a non android user I had no idea how "system back" works on android (apparently there's no button but it's a swipe from the left on modern tablets) and what the dashboard would be useful for.

ODevStudio:
## Summary

- Explain that Dashboard contains app settings and skin selection.
- Normalize every SkinView entry so Dashboard is directly underneath it, then route explicit exits directly to Dashboard. This preserves the native iOS left-edge back gesture.
- On Windows, disable WebView2 browser accelerator keys and add **Back to Dashboard** to the native Windows system menu.
- Keep the skin at its full viewport size with no overlaid Dashboard control.
- Document the Windows navigation control.

## Linked Issue

Fixes #427

## Verification

- `flutter test --no-pub test/unit/skin_feature/skin_navigation_test.dart` (14 passed, including a real iOS edge-drag gesture and Windows no-overlay coverage)
- `flutter test --no-pub` with the package-matched QuickJS DLL on `PATH` (3,181 passed, 1 skipped)
- Windows CMake/Visual Studio 2022 Debug build completed successfully.
- Computer Use and manual user verification confirmed **Back to Dashboard** appears in the native Windows system menu and opens Dashboard.
- The Windows guide says to use the system menu from the window icon or title bar and does not advertise a Dashboard keyboard shortcut.
- Samsung Galaxy Tab A9+ (Android 15, 1920 x 1200): the skin used the full display; the navigation guide correctly described Dashboard, edge swipes, and button-navigation Back; Android system Back opened Dashboard directly rather than returning to the Skins page.

## Impact

- Windows users get a reliable Dashboard exit while the WebView owns keyboard focus.
- The skin retains its full viewport with no overlaid control covering skin content.
- Android keeps the full skin viewport and relies on native Back/edge navigation. iOS retains Flutter's native interactive left-edge pop gesture.
- Windows WebView2 browser-specific accelerator keys are disabled. Page input shortcuts remain owned by the page/WebView.
- `doc/Skins.md` describes the Windows behavior. No API, storage, BLE/USB, migration, or plugin sandbox changes.

## Contributor Responsibility

### DIFF
--- a/doc/Skins.md
+++ b/doc/Skins.md
@@ -3320,7 +3320,7 @@ This shows "Back to MySkin" in the settings plugin's nav bar. When clicked, it n
 
 **Return to the dashboard:** Skin pages served on port 3000 load the tokenless `/__decent/skin-api.js` from an absolute same-origin URL, which exposes `window.decentApp.exitToDashboard()`. ReaPrime stores the account-proxy token in escaped page metadata that only the same-origin script reads. Token injection accepts loopback and IP addresses currently assigned to the device, including Ethernet and secondary adapters; arbitrary hostnames and stale addresses are rejected. If local interface enumeration is unavailable, the WiFi address cached for the server link is used as a fallback. The script response also uses `Cross-Origin-Resource-Policy: same-origin`. In the embedded webview the callback closes the skin and reveals the Decent dashboard. In an external browser it is a no-op. The script works with `script-src 'self'`; policies that reject all same-origin scripts, such as `script-src 'none'` or nonce-only policies without `'self'`, also reject this API.
 
-The embedded webview also shows a platform-specific navigation guide when a skin opens. On Windows, a persistent Dashboard button is overlaid in the top-right corner because WebView2 owns browser keyboard shortcuts. Disable or restore the guide in **Settings** under **General** with **Skin navigation guide**.
+The embedded webview also shows a platform-specific navigation guide when a skin opens. On Windows, choose **Back to Dashboard** from the system menu, available from the window icon or by right-clicking the title bar. Disable or restore the guide in **Settings** under **General** with **Skin navigation guide**.
 
 **Server control endpoints:**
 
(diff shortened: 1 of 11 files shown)


ODevStudio:
P2] The guide promises “Dashboard,” but Back does not always go there — skin_view.dart:24–32. The new copy says the platform-specific action will “open” Dashboard.  That is only reliable when SkinView was pushed directly from Dashboard. There is a normal route Dashboard → Skins → SkinView: Dashboard exposes the Skins destination, and _goToSkin() then pushes SkinView.   On Windows, for example, the advertised Alt+Backspace handler merely calls Navigator.pop(), so that flow lands back on the skin selector, not Dashboard.  Android/iOS route-back has the same stack-dependent problem. Either make the advertised exit operation explicitly navigate to LauncherView (as the macOS “Back to Dashboard” menu already does), or change the generic copy so it does not promise Dashboard.
[P2] “tap Android Back” is incomplete for button-navigation devices in immersive mode — skin_view.dart:25–26. SkinView switches Android into SystemUiMode.immersiveSticky.  When that mode is honored, Flutter documents that the navigation/status bars are hidden and revealed by an edge swipe. � Android itself supports both gesture navigation and button navigation; gesture Back is an inward swipe from the left or right edge. � So a user on three-button navigation may not have a visible Back control to “tap.” Since this PR specifically addresses users who do not already understand Android navigation, something like “Swipe inward from either side, or reveal the navigation bar and tap Back” would cover that case more effectively.
Flutter API Docs +1
Android Developers +1
The implementation itself is otherwise clean and appropriately scoped. The new test checks the intended Android substrings, but it is only a copy contract; it cannot catch the first issue because it never exercises the navigator stack.  I would add a navigation/widget test for Launcher → Skins → SkinView → Back if the intended contract is genuinely “Back to Dashboard.”

tadelv (approved):

tadelv (approved):

PR #667, opened 2026-08-22
feat(plugins): support enum settings

### LINKED ISSUE #255: Plugin host: add support for `enum` type setting
Upcoming plugins will benefit from an enum setting type, see #254

ODevStudio:
## Summary

What changed, and why?

- Render plugin settings declared as `enum` with a select control.
- Normalize only stale stored enum values before saving, using a valid manifest default or an explicit deletion tombstone.
- Define enum `values` as a JSON array of strings and share one parser between manifest loading, service validation, and the settings UI.
- Reject non-null enum settings outside the manifest's declared values in `PluginLoaderService`, including non-UI callers.

## Linked Issue

Fixes #255

## Verification

- `flutter test --no-pub test/plugin_loader_service_appstore_test.dart test/plugins_settings_view_appstore_test.dart` (42 passed)
- `flutter test --no-pub` (3,220 passed, 1 skipped; the known Windows CRLF-sensitive AsyncAPI schema assertion failed, while CI/Linux uses LF)

## Impact

- Plugin manifests can expose `enum` settings as a fixed-choice selector. Enum manifests use JSON string arrays, and invalid values are rejected at the service boundary. No migration or security impact.

## Contributor Responsibility

### DIFF
--- a/assets/plugins/time-to-ready.reaplugin/manifest.json
+++ b/assets/plugins/time-to-ready.reaplugin/manifest.json
@@ -32,7 +32,7 @@
         },
         "status": {
           "type": "enum",
-          "values": "reached | insufficient_data | not_heating | heating",
+          "values": ["reached", "insufficient_data", "not_heating", "heating"],
           "description": "current plugin heating state"
         }
       },
--- a/doc/Plugins.md
+++ b/doc/Plugins.md
@@ -37,6 +37,12 @@ A Decaid plugin consists of two required files:
       "type": "string",
       "secure": false,
       "description": "Setting description"
+    },
+    "Roast": {
+      "type": "enum",
+      "values": "Light | Medium | Dark",
+      "default": "Medium",
+      "description": "Roast degree"
     }
   },
   "api": [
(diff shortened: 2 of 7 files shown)


tadelv (changes requested):
Two changes requested before merge:

1. `enum` validation currently exists only in `PluginsSettingsView`. `PluginLoaderService._validateSettings()` only checks whether a key exists, so callers outside this UI can still persist arbitrary values for an enum setting. Since this PR introduces `enum` as a plugin setting type, please move/share enum-value parsing at the settings-schema/service layer and reject non-null values outside the declared enum. The UI should consume the same parsed values so the two paths cannot drift.

2. Since this defines the manifest contract, please use a JSON array for `values` (for example `"values": ["Light", "Medium", "Dark"]`) rather than a pipe-delimited string. The array is structurally valid JSON, avoids delimiter/escaping concerns, and is substantially easier to validate consistently.

The rest of the implementation looks good, including stale-value handling and preserving untouched defaults.

tadelv (approved):

PR #672, opened 2026-08-24
fix(de1): recover a missing serial shot-settings frame (gh-660, gh-670)

### LINKED ISSUE #660: Recover missing DE1 shot-settings frames on serial/USB connections
## Summary

Follow-up to #634 and #635.

PR #635 bounds `shotSettings.first` so a missing shot-settings frame can no longer wedge the DE1 device-write queue or cause every later `PUT /api/v1/workflow` to time out behind it. It also preserves write serialization for deferred startup defaults.

The remaining gap is recovery: on serial/USB, if the initial `shotSettings` / `[K]` frame never arrives, Decaid now fails the affected steam/hot-water write promptly instead of hanging, but it still has no active mechanism to obtain the missing frame. Startup defaults may therefore remain unapplied for the life of that connection, and a requested steam duration can still fail to reach the machine.

## Goal

Actively recover from a missing initial DE1 shot-settings frame on serial/USB connections so Decaid can eventually apply startup defaults and later steam/hot-water updates, without compromising device-write ordering.

## Scope

- Determine the protocol-correct way to re-request or re-arm the DE1 `K` / shot-settings endpoint on serial connections.
- Trigger active recovery when the initial shot-settings wait times out instead of only waiting passively for a later frame.
- Use a bounded retry/backoff policy; do not introduce an unbounded retry loop.
- Preserve `De1Controller.runDeviceWrite` serialization and connection-generation/device-identity fencing.
- Distinguish temporary missing-frame recovery from disconnect / endpoint-unavailable failures.
- Apply deferred startup defaults once recovery succeeds.
- Ensure subsequent steam and hot-water settings can use the recovered current `De1ShotSettings`.

## Verification

Add a regression test that reproduces the serial failure mode:

1. Connect a serial DE1 without receiving the initial `K` frame.
2. Let the initial shot-settings wait time out.
3. Verify Decaid performs the active recovery action.
4. Deliver a `K` frame in response.
5. Verify startup defaults, including configured steam duration, are written.
6. Verify normal device writes remain serialized throughout recovery.

Hardware verification on a DE1 over USB/serial is desirable if possible.

## Context

In the #634 field report, the DE1 was connected over USB and there were zero shot-settings frames for the entire affected session. #635 addresses the queue wedge and error propagation; this issue tracks making the underlying serial connection recover automatically.

allofmeng:
## Summary

What changed, and why?

- A DE1 pushes its `K` (shot settings) frame only when the settings change. `UnifiedDe1Transport._bleConnect` seeds `_shotSettingsSubject` with an explicit characteristic read (`unified_de1_transport.dart:128`); `_serialConnect` only issues `<+K>` and waits. On a serial/USB connection where nothing changes, the subject stays empty for the life of that connection.
- Downstream of the empty subject: `De1Controller._initializeData` times out on `_readShotSettings` and defers startup defaults, but the deferral waits passively on `device.shotSettings.first`, which never fires. `_writeSteamSettings` and `_writeHotWaterSettings` need the current `De1ShotSettings` to `copyWith` without clobbering sibling fields, so both fail on every attempt.
- `PUT /api/v1/workflow` carries `Workflow.profile` alongside the steam, hot-water and rinse settings, so the user-visible symptom is "failed to send profiles" (#670).
- After `<B>02`, `UnifiedDe1Transport` now seeds `_shotSettingsSubject` from a local mirror of the 9-byte shot-settings frame (stock firmware defaults initially — BLE firmware resets `K` to the same defaults — refreshed by live `[K]` frames and by every `updateShotSettings` write via `recordLocalShotSettings`). The initial read then succeeds immediately and the existing startup-defaults path (including configured steam duration) and steam/hot-water writes work without any machine cooperation.
- Hardware verification on a real field DE1 over USB (2026-08-24) showed the machine never pushes `[K]`: not after the connect-time `<+K>`, not after `<-K>`/`<+K>` re-arms, not after a full `<K>` write; serial `<E>` MMR reads do answer, so the link is live both ways. The re-arm approach was therefore abandoned in favour of app ownership of the frame.
- The `K` fields have no MMR register counterparts and BLE/serial cannot be active simultaneously, so reading the values from the machine is not an option on serial.
- `De1Controller` is unchanged. Its `_readShotSettings` succeeds from the seeded subject, and later steam/hot-water writes read it through the normal path, so `runDeviceWrite` serialization and connection-generation fencing are untouched.
- `doc/DeviceManagement.md` serial protocol parity rules updated; design doc archived at `doc/plans/archive/serial-shot-settings-recovery/design.md`.

## Linked Issue

Fixes #660 and #670: with the serial subject always seeded, startup defaults and steam/hot-water writes succeed on serial machines, so `PUT /api/v1/workflow` no longer fails because the initial `K` frame never arrived.

## Verification

- Field diagnosis from the #670 log (Android SM-X210, USB serial DE1, build `5bb2d525`, 2026-08-24 09:52-10:22). Eight machine connects, and every one logged `Initial shot settings unavailable; deferring startup defaults` with `TimeoutException after 0:00:02.000000`. Across the same 30 minutes there were zero `mmr write: targetSteamFlow` and zero flush setting writes, and **38 of 40** `PUT /api/v1/workflow` requests returned 503. The two that succeeded changed no steam, hot-water or rinse value, so `updateWorkflowSettings` returned before touching the device.
- **Hardware verification (done).** Real field DE1 on USB serial (`/dev/cu.wchusbserial5B1F0919251`, 115200 8N1), 2026-08-24. Commands were newline-terminated exactly as `SerialTransport.writeCommand` emits them and subscriptions were cleared first. The machine streamed `[N]`/`[M]`/`[Q]` under the app's connect sequence but never pushed `[K]`: not after the connect-time `<+K>` (5s), not after two `<-K>`/`<+K>` re-arms (2x5s), and not after a full `<K>` settings write (10s). Serial `<E>` MMR reads did answer (`[E]0480382864...` for `targetSteamFlow`), confirming the link was live both ways. This refuted the re-arm approach and motivated local frame ownership.
- New `test/unit/models/device/impl/de1/unified_de1/serial_shot_settings_mirror_test.dart` verifies:
  - serial connect seeds the subject from the local firmware-default mirror;
  - no `<-K>` re-arm is issued;
  - a live `[K]` frame refreshes the mirror for the next connect;
  - `recordLocalShotSettings` refreshes the mirror for the next connect;
  - BLE remains on its existing characteristic-read path and issues no serial commands.
- `test/controllers/de1_controller_shotsettings_stall_test.dart` drives a real `UnifiedDe1` over a quiet serial fake that never pushes `[K]`. Startup defaults still run, configured steam duration 16 lands in the `<K>` write, the fan write precedes the steam write (serialization preserved), and no re-arm is emitted.
- The earlier async prime loop, retry configuration, `shotSettingsPrimed` future and teardown-generation fences were removed with the design pivot. The mirror seed is synchronous at serial connect, so there is no in-flight priming lifecycle to fence.
- Serial parity coverage was updated so `shotSettings` is intentionally no longer considered a potentially missing serial notification: it is seeded locally at connect.
- `flutter test test/unit/models/device/impl/de1/unified_de1/ test/controllers/de1_controller_shotsettings_stall_test.dart` -> green.
- Full `flutter test` -> **3234 passed**.
- `flutter analyze` -> **No issues found**.
- PR checks on `bf9cde8` -> **green**.

## Impact

- User-visible: serial/USB connections no longer depend on the DE1 retransmitting `K`. Startup defaults (including configured steam duration) and later steam/hot-water/profile writes have an immediate current `De1ShotSettings` value to work from instead of timing out.
- Serial ownership model: the transport starts with the stock firmware 9-byte `K` defaults, then keeps that mirror current from every local `updateShotSettings` write and any live `[K]` frame the machine does emit. Because Decaid writes the complete `K` frame, sibling fields remain available for later `copyWith` updates.
- Initial-state tradeoff: on a fresh serial transport with no live or locally written frame yet, Decaid treats the stock firmware defaults as authoritative. This is intentional because hardware verification found no serial read/retransmit mechanism for the `K` fields.
- No re-arm/retry loop or serial keepalive is introduced. The existing serial reconnect/liveness model remains unchanged.
- BLE is unaffected: it continues to seed shot settings from the characteristic read during `_bleConnect`.
- Documentation: `doc/DeviceManagement.md` serial protocol parity rules and the archived design record are updated to describe app ownership of the serial shot-settings frame.
- No API/spec, migration, or security impact.

## Contributor Responsibility

---

https://claude.ai/code/session_015J7EV8fcsjBNH15axyAiwk


tadelv (changes requested):
The diagnosis and overall recovery approach look sound, but I found one lifecycle race that should be fixed before merge, plus one regression-coverage gap. I'd also like hardware verification before we treat this as closing #660/#670.

**Blocking: explicit disconnect does not fence the priming task early enough.**

`_cacheGeneration` is only bumped in `_resetCachedState()`, but `disconnect()` first cancels the connection-state listener and then sends the serial unsubscribe sequence before `_resetCachedState()` runs. If the priming timeout expires during that teardown window, `_primeStillCurrent()` can still return true and the background task can send `<-K>` / `<+K>` while disconnect is already in progress.

The current `disconnect during priming abandons it` test doesn't cover this path because it calls `serial.emitDisconnected()`, which lets the still-active connection-state listener immediately reset the cache. Please add coverage that calls `transport.disconnect()` while the prime wait is outstanding and proves no re-arm occurs after teardown begins. I think the generation/prime fence should be invalidated at the start of explicit disconnect; please also consider the equivalent dispose/new-lifecycle edges while touching this.

**Please close the end-to-end #660 regression gap.**

The new tests prove that the transport subject becomes populated after a re-arm, which is useful, but #660's acceptance case asks for the recovered `K` to flow through to deferred startup defaults (including configured steam duration) and for normal device-write serialization to remain intact. #635 already has controller-level late-frame/serialization coverage, so extending that fixture with this real recovery path would give us much stronger confidence that the transport fix actually closes the user-visible failure.

**Hardware verification requested before merge.**

The critical protocol assumption here is that a stock DE1 over USB/serial will actually retransmit its current `K` frame after `<-K>` / `<+K>`. The fake is intentionally programmed to do that, so unit tests cannot validate the assumption itself. Please smoke-test on a real DE1 over USB and confirm the sequence: initial `K` absent -> re-arm occurs -> `K` arrives -> deferred defaults / configured steam duration are successfully written. If the machine does not answer the re-arm, then this PR is a no-op for the field case and #660 needs a different protocol action.

**Non-blocking nits, if you feel like tightening them while here:**

- The BLE test never calls `bleTransport.connect()`, so `await bleTransport.shotSettingsPrimed` only observes the constructor's already-completed future. It would be better to connect and then assert no serial priming path was started.
- The stated `~6s bounded` worst case looks closer to ~7s with the defaults: 2s initial wait + 2s first retry + 1s backoff + 2s second retry.

Everything else I checked looks good: bounded retries, `K` routing to `_shotSettingsSubject`, transport-local scope, and current CI are all sensible.

tadelv:
Review round 2 (baa4e402):

- **Fence at teardown start**: the cache generation is now bumped as the first statement of `connect()`, `disconnect()` and `dispose()`, so the prime is invalidated before the serial unsubscribe sequence runs. New test "an explicit disconnect during the prime wait never re-arms" stalls the teardown mid-unsubscribe (gated fake), lets the prime wait expire inside that window, and proves no `<-K>`/`<+K>` is issued — it fails against the unfenced code.
- **#660 regression gap closed**: new controller-level e2e test drives a real `UnifiedDe1` over a recovering serial fake (answers MMR reads; delivers `K` only on the 2nd re-arm). Asserts: the initial read times out and the deferred path arms before recovery; the recovered frame flows into deferred startup defaults; configured steam duration 16 lands in the shot-settings `<K>` write; the fan write precedes the steam write (serialization intact).
- **BLE nit**: the BLE test now calls `connect()` before asserting no serial priming path started.
- **~6s nit**: PR description now says ~7s bounded (2s + 2s + 1s + 2s).

Hardware verification on a real DE1 over USB is still outstanding — the re-arm assumption (a stock DE1 retransmits `K` after `<-K>`/`<+K>`) remains unproven by unit tests.

tadelv:
Hardware verification done — result is **negative** for the re-arm assumption.

**Setup:** stock DE1 over USB serial at `/dev/cu.wchusbserial5B1F0919251` (WCH CH9102, 115200 8N1 — identified as the DE1 protocol port: it answers the app's connect sequence with `[M]`/`[N]`/`[Q]` frames; the second WCH port on the machine is a firmware debug console). Machine idle, not brewing.

**Protocol probe** (app's exact `_serialConnect` sequence: `<+N><+M><+Q><+K><+E><+I><+R><B>02`, then re-arms):

| step | window | `[K]` frames |
|---|---|---|
| connect-time `<+K>` | 10s | 0 |
| re-arm 1 (`<-K><+K>`) | 5s | 0 |
| re-arm 2 (`<-K><+K>`, +1s backoff) | 5s | 0 |
| fresh session, `<+K>` only | 8s | 0 |
| `<+K>` after `<B>02` | 8s | 0 |

`[M]` shot-sample frames (and `[Q]`, `[N]`) streamed continuously throughout — the link was live both ways. Control requests also got no response: `<+A>` (versions) 8s, `<+J>` (temps) 8s, `<E>` MMR read of 0x00803808 3s — zero frames.

**Conclusion:** on this machine the DE1 does not retransmit its current `K` after `<-K>`/`<+K>`, so the recovery is a no-op for the field case — exactly the risk you flagged. Since serial MMR reads (`<E>`) don't answer either, the MMR-read alternative is also unavailable on this unit. #660 needs a different protocol action (e.g. persist + re-seed the last-known shot settings locally at connect — the app already records every `K` it parses and every `updateShotSettings` write via `recordLocalShotSettings`, so a storage-backed seed would make the deferred-defaults and steam/hot-water writes work without any machine cooperation).

The re-arm loop is kept as-is: bounded, fenced, and provably unable to regress a machine it does not help. Docs and PR description updated to record the verified outcome.

tadelv:
Correction to my earlier hardware report: my first probes wrote commands **without** the trailing `\n` that `SerialTransport.writeCommand` emits. Re-run with correct framing (and unsubscribing first to clear the tablet session's retained subscription state — the machine went fully silent once unsubscribed, proving the earlier `[M]`/`[N]`/`[Q]` streaming was retained state, not a response to my commands):

- app connect sequence: `[N]`/`[M]`/`[Q]` stream, **no `[K]`** (5s window)
- re-arm 1 and 2 (`<-K>`/`<+K>`): **no `[K]`** (2x5s)
- `<E>` MMR reads **do answer** (`[E]0480382864...` = targetSteamFlow 1.0) — the link is live both ways; my earlier "no `<E>`" claim was the newline artifact
- `<+A>`/`<+J>` one-shot subscribes: still no response

The headline conclusion is unchanged and now stands on correct wire format: **this real field DE1 does not retransmit `K` after `<-K>`/`<+K>`**, so the re-arm recovery is a no-op for the field case and #660 needs a different protocol action.

tadelv:
Design pivot after hardware verification (`bf9cde8`): the re-arm recovery is gone, replaced by app ownership of the frame.

**What the hardware showed** (real field DE1, newline-terminated commands as `writeCommand` emits them): the machine never pushes `[K]` — not after the connect-time `<+K>`, not after two `<-K>`/`<+K>` re-arms, not after a full `<K>` settings write (10s). `<E>` MMR reads do answer, so the link is live both ways — the machine simply does not retransmit shot settings.

**New approach:** `UnifiedDe1Transport` keeps a local 9-byte shot-settings mirror (stock firmware defaults initially — matching what BLE firmware resets to), refreshed by live `[K]` frames and by every `updateShotSettings` write via `recordLocalShotSettings`. `_serialConnect` seeds `_shotSettingsSubject` from the mirror right after `<B>02`, so `De1Controller`'s initial read succeeds immediately and startup defaults (including configured steam duration) plus steam/hot-water writes work with zero machine cooperation. MMR-register reads were ruled out (the K fields have no register counterparts) and BLE seeding is impossible (single active connection; BLE firmware resets K to defaults).

The review's teardown fences were tied to the async prime loop; the seed is synchronous at connect, so both the loop and the fences are removed.

Tests: transport mirror tests (seed / live-frame refresh / local-write refresh / BLE untouched) + controller e2e where a quiet serial fake never pushes K — steam duration 16 lands on the wire, no `<-K>` re-arm issued. Full suite 3234 passed, analyze clean.

tadelv (approved):

PR #673, opened 2026-08-24
Fix HDS USB protocol compatibility

### LINKED ISSUE #669: [Bug Report] **Can't tare or connect to scale via USB**
## Description
**Can't tare or connect to scale via USB**

after updating firmware and on, cannot connect to scale via USB. Shows as connected in settings, but weight label is red with spinner and weight never displayed. two entries for machine in settings, but machine info including curves display normally.

## System Info
**System Information:**
- App Version: 0.8.1 (build 2397)
- Commit: 5bb2d525
- Branch: HEAD
- Platform: android
- OS Version: BP4A.251205.006.X520XXS9CZG3

## Attachments
Logs and screenshots: https://gist.github.com/reaprime-bot/d32592efca7f7d3a6c39d175b78c591a

---
_Submitted via in-app feedback on 2026-08-23T11:21:35.968853_

ODevStudio:
## Summary

What changed, and why?

- Send the OpenScale-compatible 10 Hz USB enable command and checksum-valid tare command.
- Keep HDS USB in `connecting` until a valid weight frame arrives, then parse fragmented/coalesced mixed serial input with checksum validation.
- Preserve USB read errors and early transport closure until both the enable write and first valid frame succeed, including when a frame arrives while the write is pending; report only a silent two-second readiness failure as `EndpointUnavailableException` rather than the connection manager's outer timeout.
- Refresh the watchdog only for valid weight frames and verify recovery after the retry threshold is reached.

## Linked Issue

Fixes #669

## Verification

- `flutter analyze` (passes)
- Focused HDS USB and shared BLE protocol tests: 61 tests pass
- VS2022 Windows debug build (passes)
- Physical HDS/OpenScale 3.1.13 or later verification on `COM5` (device advertised `FW: 3.1.13-dev-custom`): USB connected, weight frames arrived at about 10 Hz, tare returned 2.5 g to 0.0 g, and the connection remained healthy without initialization or watchdog errors
- Full `flutter test`: HDS tests pass; the repository suite retains unrelated Windows failures from CRLF-sensitive assertions, missing `quickjs_c_bridge.dll`, and local port collisions

## Impact

- Restores USB weight streaming and tare compatibility with HDS/OpenScale 3.1.13 or later.
- USB scales no longer appear connected before a validated weight stream is available.
- Initialization errors now retain their real cause and the two-second readiness timeout is no longer mislabeled as the connection manager's 30-second timeout.
- No API or migration impact.

## Contributor Responsibility

### DIFF
--- a/doc/DeviceManagement.md
+++ b/doc/DeviceManagement.md
@@ -84,6 +84,7 @@ Discovery services are responsible for scanning and creating device instances. E
   - `lib/src/services/serial/serial_service_android.dart` (Android)
   - `lib/src/services/serial/serial_service.dart` (factory)
 - **Discovery:** Enumerates serial ports, probes for device identification
+- **HDS USB readiness:** `HDSSerial` enables the 10 Hz OpenScale binary stream and remains `connecting` until a checksum-valid weight frame arrives. Its buffered decoder accepts fragmented/coalesced frames mixed with firmware text; only valid weight frames refresh the watchdog.
 
   DE1-family detection uses product names and the normal protocol probe:
   1. Exact `productName == "DE1"` creates `UnifiedDe1`; exact
--- a/lib/src/models/device/impl/decent_scale/protocol.dart
+++ b/lib/src/models/device/impl/decent_scale/protocol.dart
@@ -1,0 +1,7 @@
+import 'dart:typed_data';
+
+Uint8List buildDecentScaleCommand(List<int> commandBytes) {
+  final bytes = <int>[0x03, ...commandBytes];
+  final checksum = bytes.fold(0, (value, byte) => value ^ byte);
+  return Uint8List.fromList([...bytes, checksum]);
+}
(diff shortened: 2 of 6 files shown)


tadelv (changes requested):
There is one initialization race that I think should block merge.

`onConnect()` currently waits concurrently for the enable write and the first valid weight frame. Once a valid frame completes `_initialization`, subsequent stream errors/closure can no longer fail that readiness future. If the valid frame arrives while the enable write is still pending, then the transport errors/closes, and the write later completes, `Future.wait()` can still succeed and `onConnect()` can start the watchdog and emit `connected` on a dead transport.

Smallest follow-up: add a test with a pending enable write, emit a valid frame, then emit a stream error/close before completing the write. Initialization should fail with that transport error and must never emit `connected`. Then keep initialization fail-able by stream error/closure until both the enable write and first valid frame have succeeded.

The framing, tare checksum, buffered weight parser, checksum validation, signed weight decoding, and watchdog behavior otherwise look good.

tadelv (approved):

PR #682, opened 2026-08-25
feat(plugins): document the settings schema and add per-setting labels

allofmeng:
## Summary

What changed, and why?

- `GET /api/v1/plugins` already returns each plugin's manifest `settings` schema verbatim — including the per-setting `description` the plugin author wrote. Verified against a running instance:

  ```
  $ curl -s http://localhost:8080/api/v1/plugins   # .[] | select(.id=="shot-upload.reaplugin") | .settings
  {
    "AutoUpload":      { "type": "boolean", "description": "Upload shots automatically when they finish (opt-in; off by default)", "default": false },
    "LengthThreshold": { "type": "number",  "description": "Only upload shots longer than this many seconds (skips flushes)",      "default": 5 },
    "DrainHistory":    { "type": "boolean", "description": "Also upload shots already in your history, a few at a time while the machine is idle (opt-in; off by default)", "default": false }
  }
  ```

- The spec described that field as nothing more than an opaque `type: object` labelled "Plugin settings schema". A client reading the OpenAPI doc had no way to learn that `settings.<key>.description`, `.type`, `.default`, `.secure` or enum `.values` exist. The data was on the wire and undiscoverable.
- Adds a `PluginSettingSchema` component and points `PluginManifest.settings` at it via `additionalProperties`, with a worked example.
- Corrects `PluginManifest.apiVersion`, documented as `string` but sent as an integer (`"apiVersion": 1`).
- `doc/Plugins.md` now records that the list endpoint carries the schema while `/settings` carries only stored values, and tells authors to write a `label` and a `description` for every setting.

**Adds an optional `label` to each setting** (second commit). Settings render under their storage key today, so a user sees `LengthThreshold` and `BackSyncIntervalSeconds` rather than names written for them. The key is a storage identifier and a poor label, but it was the only string a client had.

- `pluginSettingLabel()` resolves it: trims whitespace, and falls back to the key when the field is absent, blank or not a string. A manifest that omits it renders exactly as before.
- The settings dialog heading uses the resolved label (`plugins_settings_view.dart`, one line).
- All 20 settings across the four bundled plugins in this repository are labelled.
- The four bundled plugins' manifest versions are bumped (patch release each) so `_copyBundledPlugins()` actually replaces on-disk copies in release builds; the plugin-object versions reported in `plugin.js` are bumped in lockstep where present (settings, visualizer). The dye2 and shot-upload plugins are fetched from their own repositories at build time (checksum-pinned, see `scripts/fetch_shot_upload_plugin.sh`); the shot-upload manifest is versioned in [decentespresso/shot-upload](https://github.com/decentespresso/shot-upload), so its settings (`AutoUpload`, `LengthThreshold`) will get labels there in a follow-up — the pinned v0.2.1 still renders its storage keys.
- No parser change needed: `PluginManifest` keeps `settings` as the raw manifest map, so `label` reaches clients through `GET /api/v1/plugins` the same way `description` already does.

```json
"LengthThreshold": {
  "type": "number",
  "label": "Minimum shot length",
  "description": "Only upload shots longer than this many seconds (skips flushes)",
  "default": 5
}
```

**This branch was rebased onto current main**, dropping the original approach (an optional `description` per `api[]` entry). Endpoint descriptions were the wrong gap: the settings schema — the thing a skin needs to render a settings page — already reaches clients and only needed documenting. The rebase also picks up main's extraction of the shot-upload plugin (now fetched from `decentespresso/shot-upload` at build time) and the dye2 move to `packages/dye2-plugin/`. The `label` work touches only `pluginSettingLabel()` in `lib/src/plugins/plugin_manifest.dart` and one line in `plugins_settings_view.dart`.

## Linked Issue

Fixes # N/A — maintainer PR.

## Verification

- **Verified over HTTP against a running instance**, not inferred from source:
  - `GET /api/v1/plugins` returns the full setting schema with descriptions (output above).
  - `GET /api/v1/plugins/shot-upload.reaplugin/settings` returns `{"AutoUpload":true}` — stored values only, no schema. This is what the spec's existing `PluginSettings` component already describes, and it is unchanged here.
  - Cross-checked the documented shape against every plugin the live instance serves: fields in use are `type`, `description`, `default`, `secure`; types in use are `boolean`, `number`, `string`. **No undocumented field or type.** `enum` / `values` are documented too — supported by `_validateSettings` and rendered at `plugins_settings_view.dart:1052`, though no currently-loaded plugin uses them.
- New `test/plugins/plugin_settings_schema_spec_test.dart` (6 tests): `settings` resolves to `PluginSettingSchema`; `apiVersion` is documented as `integer`; every setting field and every setting type used by a bundled plugin manifest is documented. The last two fail if a plugin later adds a field or type the spec omits. Two more cover `label`: the fallback behaviour (absent, blank, non-string, and whitespace-padded input), and that every setting of an in-repo bundled plugin declares one (externally fetched plugins are versioned in their own repositories and excluded via `.gitignore`).
- Negative check: all 4 tests fail against the pre-change spec (`+0 -4`), pass after (`+4`).
- `flutter test test/plugins/` — `+185: All tests passed!`; `flutter analyze` — `No issues found!`
- Full suite skipped locally by request; CI runs it on this push. No `lib/` code changes in this PR, so the blast radius is the spec, one doc file, and one new test.

## Impact

- **API/spec:** `PluginManifest.settings` is now described by a `PluginSettingSchema` component, which also introduces the new optional `label`. No endpoint or request changes; the only payload change is the added `label` key on bundled manifests. The `apiVersion` correction makes a generated client type it as an integer, matching the wire; a client generated from the old spec had it wrong.
- **User-visible:** the plugin settings dialog now shows each setting's `label` instead of its storage key — "Minimum shot length" rather than `LengthThreshold`. A plugin whose manifest has no `label` is unchanged. The wider point is to let skins render a settings page — labels, help text, defaults — from `GET /api/v1/plugins` without reading the plugin's repository.
- **Compatibility:** `label` is optional and additive in both directions. Manifests without it fall back to the key, and an older Decaid ignores the field. Nothing to migrate.
- **Docs:** `doc/Plugins.md` settings field reference.
- **Security:** none. The spec now states explicitly that `secure` values are never returned in plaintext and appear as an `isSet` state object, which was already the behaviour.

## Contributor Responsibility

---

https://claude.ai/code/session_015J7EV8fcsjBNH15axyAiwk

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -6987,10 +6987,12 @@ components:
           example:
             AutoUpload:
               type: boolean
+              label: Upload shots automatically
               description: Upload shots automatically when they finish (opt-in; off by default)
               default: false
             LengthThreshold:
               type: number
+              label: Minimum shot length
               description: Only upload shots longer than this many seconds (skips flushes)
               default: 5
         api:
@@ -7042,6 +7044,14 @@ components:
           description: >
             Value type. `enum` additionally requires `values`, and the API
             rejects a saved value that is not one of them.
+        label:
+          type: string
+          description: >
+            Human-readable name for the setting, written by the plugin author.
+            Optional: when it is absent, empty or whitespace, clients fall back
+            to the setting's key, so a manifest that omits it renders exactly as
+            before. Use it so a settings form can read "Upload shots
+            automatically" instead of the storage key "AutoUpload".
         description:
           type: string
           description: >
@@ -7069,6 +7079,7 @@ components:
           description: Allowed values. Required when `type` is `enum`, otherwise absent.
       example:
         type: boolean
+        label: Upload shots automatically
         description: Upload shots automatically when they finish (opt-in; off by default)
         default: false
 
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -6965,15 +6965,34 @@ components:
         version:
           type: string
         apiVersion:
-          type: string
+          type: integer
+          description: Plugin API contract version the plugin targets.
+          example: 1
         permissions:
           type: array
           items:
             type: string
             enum: [log, api, emit, pluginStorage, events.machine, events.shots, proxy.decent_api, proxy.decent_api.write]
         settings:
-          description: Plugin settings schema
           type: object
+          description: >
+            Declared settings for this plugin, keyed by setting name. This is the
+            manifest schema, not the stored values: it carries the type, the
+            human-readable description and the default for each setting, so a
+            client can render a settings form without reading the plugin's
+            repository. Current values come from
+            GET /api/v1/plugins/{id}/settings instead.
+          additionalProperties:
+            $ref: '#/components/schemas/PluginSettingSchema'
+          example:
+            AutoUpload:
+              type: boolean
+              description: Upload shots automatically when they finish (opt-in; off by default)
+              default: false
+            LengthThreshold:
+              type: number
+              description: Only upload shots longer than this many seconds (skips flushes)
+              default: 5
         api:
           type: array
           description: A list of endpoints this plugin will expose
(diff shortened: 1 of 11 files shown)


tadelv:
thanks @allofmeng ! However, we were discussing the `settings` field, not endpoints.

tadelv (approved):

PR #684, opened 2026-08-25
fix(de1): separate refill-kit detection from override

### LINKED ISSUE #671: [Docs] Document MachineInfo.extra.refillKit (and consider surfacing it live)
## Summary

Decaid reports the refill kit's **configured mode** (`refillKitSetting`: `0` force off, `1` force on, `2` auto) via `GET /api/v1/machine/settings/advanced`, but never reports whether a kit was actually **detected**. Nothing about the refill kit appears on any WebSocket channel at all. Skins therefore cannot tell a plumbed machine in auto mode apart from a machine with no kit.

## Why this matters

On a plumbed machine the auto-refill holds the tank level right around `refillLevel`, so a level-vs-threshold check fires more or less continuously. A skin that warns on low water then shows a permanent or flickering "Out of water" on a machine that is refilling itself perfectly well.

We hit this in Streamline: decentespresso/streamline-js#60 (low-water overlay repeatedly replacing the machine status mid-steam on a DE1XL with an automatic refill kit). We fixed the status-clobbering on our side, and we can suppress the level warning when `refillKitSetting === 1`, but there is no way to do the same for `refillKitSetting === 2` (auto) — which is the default. An auto-mode machine with a kit installed is indistinguishable from one without.

## What exists today

REST — `De1AdvancedSettings` / `De1AdvancedSettingsResponse` (rest_v1.yml):

```yaml
refillKitSetting:
  description: Refill kit behavior. `2` = auto (default), `1` = force on, `0` = force off.
  type: integer
  enum: [0, 1, 2]
```

WebSocket — `ws/v1/machine/waterLevels` (websocket_v1.yml) carries only:

```yaml
WaterLevels:
  properties:
    currentLevel:   # Current water height in the tank (mm)
    refillLevel:    # Water level threshold for refill warning (mm)
```

Grepping `websocket_v1.yml` for "refill" returns those two lines and nothing else. `MachineSnapshot`, `ShotSettings`, `DevicesState` and `DeviceInfo` carry no kit or accessory information either.

## Request

Report the **effective / detected** refill-kit state, and put it somewhere a skin can react to live.

Preferred: add it to the `WaterLevels` frame, next to the values it qualifies —

```yaml
WaterLevels:
  properties:
    currentLevel:
    refillLevel:
    refillKitPresent:      # boolean — a kit is actually detected/active
    refillKitMode:         # optional: 0 | 1 | 2, the configured setting, so a
                           # skin does not need a separate REST read to explain
                           # why refillKitPresent is what it is
```

`MachineSnapshot` would work too if the value belongs with machine state rather than water state. Either way the useful field is the resolved answer ("is this machine refilling itself right now?"), not the configuration — the configuration is already reachable over REST.

If firmware genuinely cannot report detection in auto mode, saying so in the `refillKitSetting` description would still help: it would tell skin authors that `2` is unresolvable and that suppressing low-water warnings requires the user to select "Force On".

## Environment

Reported against Decaid v0.7.17, DE1XL, Linux x86_64.

allofmeng:
## Summary

What changed, and why?

Originally a docs-only PR. @tadelv's review was right that it documented an implementation ambiguity instead of fixing it, so the PR now fixes the model and the docs follow.

- `UnifiedDe1` kept two meanings in one field. `onConnect()` read MMR `refillKitPresent` into `_refillKit` and published bit 0 as `extra.refillKit`, then wrote auto back to the register **without** updating `_refillKit`; `setRefillKitSettings()` later overwrote the same field with the requested override. `getRefillKitSettings()` returned `De1RefillKitSettings.fromInt(_refillKit)`.

  The consequence was a wrong answer, not only unclear docs: straight after connect a machine with a kit reported `refillKitSetting: 1` (force on) and one without reported `0` (force off), while the register actually held `2` (auto, written at `unified_de1.dart:289`).

- Split along the `decentespresso/de1app` line between `refill_kit_detected` and `refill_kit_override`:

  ```dart
  int _refillKitDetected = -1;
  De1RefillKitSettings _refillKitSetting = De1RefillKitSettings.auto;
  ```

  - `_refillKitDetected` takes the connect-time MMR read and feeds `extra.refillKit` via `(_refillKitDetected & 0x01) != 0`. Nothing else writes it.
  - `_refillKitSetting` is set to `auto` right after the connect-time write, and updated by `setRefillKitSettings()`.
  - `getRefillKitSettings()` returns the override directly — no `fromInt` on a raw register value.

- `POST /api/v1/machine/settings/advanced` used `firstWhere` with no `orElse` on `refillKitSetting` (`de1handler.dart:424`) and threw a 500 on anything outside `0`/`1`/`2`. It now answers **400**. Not on the review list, but with `fromInt` falling back to `auto` an unvalidated boundary would silently turn `refillKitSetting: 7` into auto, so the two belong together.

- `De1RefillKitSettings.fromInt` keeps its `orElse: auto`, now reached only for configuration parsing. Detection never goes through it.

- Spec and `doc/Api.md` describe the two concepts separately. All "detection result" language is gone from `refillKitSetting`; `extra.refillKit` now states that writing `refillKitSetting` does not change it.

## Linked Issue

Fixes #671

## Verification

- New `test/unit/models/device/impl/de1/unified_de1/refill_kit_semantics_test.dart` — 8 tests driving `UnifiedDe1` through `FakeBleTransport`, as requested rather than enum conversion alone. `queueOnConnectResponses` already accepts `refillKitPresent`, so no new test harness was needed.

  | Test | Covers |
  |---|---|
  | a detected kit is published as `machineInfo.extra.refillKit` | detection reaches machine info |
  | connect leaves the configuration at auto, not at the detected value | **failed before the fix** (`forceOn`) |
  | writing `forceOff` does not disturb the detected state | review item 3 |
  | writing `forceOn` does not invent a detection | review item 4 |
  | detection masks bit 0 while the configuration stays auto | raw register never parsed as config |
  | the override survives a reconnect of the same device object | actual reconnect behaviour, see below |
  | a freshly discovered device starts from auto again | **failed before the fix** (`forceOn`) |
  | the connect-time write puts auto on the wire | asserts `0x02` on `writeToMMR` |

- Negative check: 2 of the 8 fail against the pre-fix code, both returning `forceOn` where `auto` is correct. The other 6 are regression guards for behaviour the split must not break.
- REST round-trip (review item 5) is already covered by the existing `round-trips through GET after write` in `de1handler_settings_reset_test.dart` — not duplicated. Added `rejects a refillKitSetting outside 0/1/2 with 400` there.
- `refill_kit_settings_test.dart` spec-contract assertions flipped from "documented as a detection result" to override wording, plus a new assertion that `extra.refillKit` is documented as detection.
- `flutter test` — `3341 passed, 6 failed`. The 6 are all in `test/webui_support/webui_token_injection_test.dart` (`getWifiIP` fallbacks), and the identical 6 fail on the base commit with the change stashed. Unrelated to this PR.
- Hardware evidence for the detection path unchanged from the original PR: a DE1XL over USB serial (2026-08-24 09:53) shows `mmr read: refillKitPresent` -> register `1` -> `extra: {refillKit: true, voltage: 220}` -> `mmr write: refillKitPresent`.

## Impact

- **Behaviour (bug fix):** `GET /api/v1/machine/settings/advanced` -> `refillKitSetting` now reports the override. Before the fix it reported the detected register value straight after connect, so a plumbed machine read back `1` (force on) although auto was in effect. A client that treated that GET as the current mode was reading detection.
- **Behaviour:** `GET .../advanced` no longer 500s when the register holds a value outside `0`/`1`/`2` (including `-1`, before the connect-time read lands).
- **API:** `POST .../advanced` with a `refillKitSetting` outside `0`/`1`/`2` now returns **400** instead of **500**. A client sending a valid value sees no change.
- **Spec:** `MachineInfo.extra` gains named `refillKit` and `voltage` properties (payload unchanged — Decaid already sends both), `additionalProperties: true`. Both `refillKitSetting` descriptions rewritten as override; enum values unchanged.
- **User-visible:** none directly. Enables skins to suppress spurious low-water warnings on plumbed machines — the Streamline case in decentespresso/streamline-js#60.
- **Docs:** `doc/Api.md` advanced-settings row rewritten.
- **Security:** none.

## Notes for the reviewer

**One correction to review item 7.** Reconnect does *not* return the configuration to auto. `onConnect()` has `if (_info != null) return;` at `unified_de1.dart:256`, so reconnecting the same device object skips the whole MMR block, including the write of auto — the override survives. Only a device discovered afresh starts from auto. Two tests pin both halves, and the spec says "lasts only until the machine is discovered and connected again" rather than "every connect". The original PR description claimed the latter; it was wrong.

Kept out of scope, per the review: no WebSocket refill-kit state, no persistence of the override across restarts, no new settings storage, no water-level API changes. The masking inconsistency I raised previously is resolved by construction — `fromInt` no longer sees raw register values at all, so there is nothing left to reconcile.

## Contributor Responsibility

---

https://claude.ai/code/session_01UfCnQuQG7i7obdVdkCRG69

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -5541,13 +5541,18 @@ components:
             refillKit:
               type: boolean
               description: >
-                Whether an automatic refill kit is detected on this machine.
+                Whether an automatic refill kit was detected on this machine.
                 Bit 0 of the refill-kit register as read at connect, before
-                Decaid resets that register to auto. Use this, not
-                `refillKitSetting`, to decide whether a machine refills itself:
-                on a plumbed machine in auto mode a low-water check against
+                Decaid writes auto back to that register. This is machine
+                detection, not configuration: writing `refillKitSetting` does
+                not change it, and it is fixed for the lifetime of the
+                connection.
+
+                Use this, not `refillKitSetting`, to decide whether a machine
+                refills itself: on a plumbed machine a low-water check against
                 WaterLevels.refillLevel fires more or less continuously, and
-                this is the only field that tells the two cases apart.
+                this is the only field that tells that case apart from a
+                machine with no kit.
             voltage:
               type: integer
               description: Heater supply voltage in volts, as reported by the machine.
@@ -6134,15 +6139,19 @@ components:
           enum: [120, 230]
         refillKitSetting:
           description: |
-            Refill kit behaviour to write: `2` = auto (default), `1` = force on,
+            Refill kit override: `2` = auto (default), `1` = force on,
             `0` = force off.
 
-            This does not persist. Decaid writes `2` (auto) to the same register
-            on every connect, so a force-on or force-off selection is lost the
-            next time the machine reconnects.
+            Writes the value to the machine's refill-kit register and updates
+            the override Decaid reports back. It does not change
+            `MachineInfo.extra.refillKit`, which is what the machine detected
+            at connect.
 
-            Note that a write and a read do not mean the same thing — see the
-            same field on `De1AdvancedSettingsResponse`.
+            The override is not persisted. Decaid writes `2` (auto) at connect,
+            so a force-on or force-off selection lasts only until the machine is
+            discovered and connected again.
+
+            Any value other than `0`, `1` or `2` is rejected with `400`.
           type: integer
           nullable: true
           enum: [0, 1, 2]
@@ -6170,22 +6179,16 @@ components:
           enum: [120, 230, -1]
         refillKitSetting:
           description: |
-            Refill kit register as last read from the machine. A read is a
-            **detection result**, not the configuration that was written:
-            `1` = a kit is detected, `0` = no kit detected.
+            Refill kit override currently in effect: `2` = auto, `1` = force on,
+            `0` = force off.
 
-            Decaid writes `2` (auto) at connect and the firmware resolves it, so
-            a value read back after connect is the firmware's answer. `2` is
-            only ever returned when a client wrote it during this connection and
-            the register has not been re-read since.
+            This is the configuration, not a detection result. Decaid writes `2`
+            (auto) at connect, so a machine that has not been written to since
+            reports `2` whether or not it has a kit fitted. For detection, read
+            `MachineInfo.extra.refillKit`.
 
-            Building a settings UI that presents `0` / `1` / `2` as three
-            user-chosen modes therefore misrepresents what a read returns.
-            `MachineInfo.extra.refillKit` is the same register's bit 0 and is
-            the clearer field to read for detection.
-
-            A value the app does not recognise (including the register never
-            having been read) is reported as `2`.
+            The value is served from Decaid's cache of the last override
+            written; the register is not re-read.
           type: integer
           enum: [0, 1, 2]
     De1CalibrationRequest:
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -5532,7 +5532,28 @@ components:
           description: Whether GroupHeadController is present and active
         extra:
           type: object
-          description: various extra information.
+          description: >
+            Machine details read from MMR once at connect. The object stays
+            open — a firmware or model may contribute keys not listed here —
+            but these two are always present for a DE1.
+          additionalProperties: true
+          properties:
+            refillKit:
+              type: boolean
+              description: >
+                Whether an automatic refill kit is detected on this machine.
+                Bit 0 of the refill-kit register as read at connect, before
+                Decaid resets that register to auto. Use this, not
+                `refillKitSetting`, to decide whether a machine refills itself:
+                on a plumbed machine in auto mode a low-water check against
+                WaterLevels.refillLevel fires more or less continuously, and
+                this is the only field that tells the two cases apart.
+            voltage:
+              type: integer
+              description: Heater supply voltage in volts, as reported by the machine.
+          example:
+            refillKit: true
+            voltage: 220
 
     MachineCapabilities:
       type: object
@@ -6113,7 +6134,15 @@ components:
           enum: [120, 230]
         refillKitSetting:
           description: |
-            Refill kit behavior. `2` = auto (default), `1` = force on, `0` = force off.
+            Refill kit behaviour to write: `2` = auto (default), `1` = force on,
+            `0` = force off.
+
+            This does not persist. Decaid writes `2` (auto) to the same register
+            on every connect, so a force-on or force-off selection is lost the
+            next time the machine reconnects.
+
+            Note that a write and a read do not mean the same thing — see the
+            same field on `De1AdvancedSettingsResponse`.
           type: integer
           nullable: true
           enum: [0, 1, 2]
@@ -6141,7 +6170,22 @@ components:
           enum: [120, 230, -1]
         refillKitSetting:
           description: |
-            Refill kit behavior. `2` = auto, `1` = force on, `0` = force off.
+            Refill kit register as last read from the machine. A read is a
+            **detection result**, not the configuration that was written:
+            `1` = a kit is detected, `0` = no kit detected.
+
+            Decaid writes `2` (auto) at connect and the firmware resolves it, so
+            a value read back after connect is the firmware's answer. `2` is
+            only ever returned when a client wrote it during this connection and
+            the register has not been re-read since.
+
+            Building a settings UI that presents `0` / `1` / `2` as three
+            user-chosen modes therefore misrepresents what a read returns.
+            `MachineInfo.extra.refillKit` is the same register's bit 0 and is
+            the clearer field to read for detection.
+
+            A value the app does not recognise (including the register never
+            having been read) is reported as `2`.
           type: integer
           enum: [0, 1, 2]
     De1CalibrationRequest:
--- a/doc/Api.md
+++ b/doc/Api.md
@@ -45,7 +45,7 @@ For browser clients on a different origin, `ETag` is exposed via `Access-Control
 | POST | `/api/v1/machine/settings` | Update machine settings (one grouped, serialized device write per request) | |
 | POST | `/api/v1/machine/shotSettings` | Update shot settings (steam temp, hot water, target volume, group temp) | |
 | GET | `/api/v1/machine/settings/advanced` | Advanced heater/phase settings | |
-| POST | `/api/v1/machine/settings/advanced` | Update advanced settings (heater phase flows/timeouts, idle temp, `heaterVoltage`, `refillKitSetting`). `refillKitSetting` does not persist — Decaid writes auto on every connect. Reading it back gives a detection result, not the written mode; `GET /api/v1/machine/info` -> `extra.refillKit` is the clearer detection field | |
+| POST | `/api/v1/machine/settings/advanced` | Update advanced settings (heater phase flows/timeouts, idle temp, `heaterVoltage`, `refillKitSetting`). `refillKitSetting` is an override (`0` force off, `1` force on, `2` auto) and is not persisted — Decaid writes auto at connect, so it lasts only for the current connection. It is not detection: `GET /api/v1/machine/info` -> `extra.refillKit` reports what the machine detected | |
 | DELETE | `/api/v1/machine/settings/reset` | Reset machine settings to defaults (fan, heater idle/phase flows + ph2 timeout, refill kit auto, flow multiplier 1.0, steam purge 0) — one grouped, serialized device write | |
 | GET | `/api/v1/machine/calibration` | Flow estimation calibration (`flowMultiplier`, calFlowEst MMR) | |
 | POST | `/api/v1/machine/calibration` | Update flow estimation calibration | |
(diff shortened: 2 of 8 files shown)


tadelv (changes requested):
There is one blocking semantic issue here: this PR documents around a Decaid implementation ambiguity that the reference `decentespresso/de1app` deliberately avoids.

`de1app` keeps two separate concepts:

- `::de1(refill_kit_detected)`: value actually read from MMR `0x80385C`
- `::settings(refill_kit_override)`: user-selected mode (`2` auto-detect, `0` force off, `1` force on)

On connection, `de1app` explicitly calls `get_refill_kit_present`, stores the machine response in `refill_kit_detected`, and separately applies `refill_kit_override`. Writing an override does not overwrite the cached detected state. Its UI likewise displays detection separately from the override selector.

Decaid currently conflates both meanings in `_refillKit`: `onConnect()` reads the MMR value into `_refillKit` and uses it for `MachineInfo.extra.refillKit`, but `setRefillKitSettings()` later overwrites `_refillKit` with the requested override. `getRefillKitSettings()` then returns that same variable without a machine read. So `_refillKit` changes meaning during a connection: initially detected machine state, later requested configuration.

Because of that, I don't think we should document `refillKitSetting` as a detection result or try to explain the ambiguity as API semantics. Please fix the model in this PR instead by separating detected state from override/configuration, following `de1app` as the reference behavior.

Requested behavior:

1. Keep a distinct cached refill-kit detection value populated from the MMR read on connection.
2. Keep the override/configuration (`auto` / `forceOff` / `forceOn`) separately.
3. `MachineInfo.extra.refillKit` should represent the detected machine state from the connect-time read.
4. `GET /api/v1/machine/settings/advanced` `refillKitSetting` should represent the override/configuration, not detection.
5. `POST .../advanced` should update/write only the override/configuration and must not destroy the cached detected state.
6. Preserve/document the current reconnect behavior explicitly unless we intentionally decide to persist/reapply the override. `de1app` does persist/reapply it, but that is a separate product decision if we don't want to change it here.
7. Update the OpenAPI/docs/tests to reflect those two independent concepts. Add coverage proving that after a detected value is captured, writing a different override does not change `MachineInfo.extra.refillKit`, while GET advanced returns the written override.

The `fromInt` hardening can stay, but once detection and override are separated the fallback semantics should be applied only to the configuration value and should not be used to reinterpret arbitrary detected MMR bit patterns.

This keeps #671 useful without baking the current conflation into the public API contract.

tadelv:
here are the agent instructions that can be used directly:

Update PR #684 so Decaid models refill-kit **detection** and **override/configuration** as two separate concepts, matching the architecture used by `decentespresso/de1app`.

## Problem

Decaid currently uses `_refillKit` for two different meanings:

1. During `UnifiedDe1.onConnect()`, `_refillKit` is populated by reading MMR `refillKitPresent` / `0x80385C`. At this point it represents the machine's detected refill-kit state.
2. `setRefillKitSettings()` later writes an override (`0`, `1`, or `2`) and overwrites `_refillKit` with that requested setting.
3. `getRefillKitSettings()` returns `_refillKit` without reading the machine.

This means the variable changes semantics during the connection.

The reference `decentespresso/de1app` does not do this. It keeps:

- `refill_kit_detected`: machine state read from MMR.
- `refill_kit_override`: user-selected configuration (`2` auto, `0` force off, `1` force on).

Use that separation as the intended model.

## Required implementation

Refactor `UnifiedDe1` so there are separate values for:

- detected refill-kit state;
- selected refill-kit override/configuration.

Naming is up to you, but something equivalent to:

```dart
int _refillKitDetected = -1;
De1RefillKitSettings _refillKitSetting = De1RefillKitSettings.auto;
```

is preferable to one overloaded integer.

### Connection behavior

During `onConnect()`:

1. Read `MMRItem.refillKitPresent`.
2. Store that machine response only in the detection state.
3. Build `MachineInfo.extra.refillKit` from that detection value.
4. Write `2` (`auto`) to the MMR as Decaid currently does.
5. Set/retain the Decaid configuration state as `auto`.

Do not overwrite the detected state when writing auto.

Do **not** add an extra MMR read after the write unless there is a concrete need for one.

### GET advanced settings

`getRefillKitSettings()` must return the **configuration/override**, not the detected machine state.

Consequently:

`GET /api/v1/machine/settings/advanced`

should report:

```json
{
  "refillKitSetting": 0 | 1 | 2
}
```

with the normal configuration semantics:

- `0` = force off
- `1` = force on
- `2` = auto

It should no longer be documented as a detection result.

### POST advanced settings

`setRefillKitSettings()` should:

1. write the selected value to `MMRItem.refillKitPresent`;
2. update only the cached configuration/override;
3. leave the cached detected state unchanged.

For example:

```text
connect:
machine detection = 1
configuration = auto

POST refillKitSetting=0:
machine detection remains 1
configuration becomes forceOff
```

`MachineInfo.extra.refillKit` must therefore remain `true` for that connection.

### MachineInfo.extra.refillKit

Continue exposing:

```json
"extra": {
  "refillKit": true
}
```

but define it as the machine's refill-kit detection captured from the MMR during connection setup.

It must not change merely because an API client writes `forceOn`, `forceOff`, or `auto`.

No live refresh is required in this PR.

## Reconnect behavior

Keep the existing Decaid behavior for now:

- reconnect reads detection from the machine;
- Decaid then writes `auto`;
- configuration becomes `auto`.

Do **not** add persistence/reapplication of the user's override in this PR unless it is already clearly supported elsewhere.

`de1app` persists and reapplies its override, but that is a separate product decision and should not expand this PR unnecessarily.

Document that the Decaid override currently lasts only for the active connection.

## OpenAPI/docs

Correct the PR's documentation so the two concepts are explicit.

`MachineInfo.extra.refillKit`:

- boolean;
- detected refill-kit state captured during connection setup;
- derived from the machine MMR;
- suitable for determining whether a refill kit was detected.

`De1AdvancedSettings.refillKitSetting` and `De1AdvancedSettingsResponse.refillKitSetting`:

- configuration/override;
- `0 = force off`;
- `1 = force on`;
- `2 = auto`;
- current Decaid behavior does not persist it across reconnects.

Remove language claiming that `GET refillKitSetting` is a detection result.

Update `doc/Api.md` consistently.

## Tests

Add regression coverage for the semantic separation.

At minimum test:

1. `0`, `1`, and `2` still map correctly through `De1RefillKitSettings`.
2. On connection, a detected value is retained independently from configuration.
3. Given detected refill kit = `1`, writing `forceOff`:
   - `getRefillKitSettings()` returns `forceOff`;
   - `machineInfo.extra['refillKit']` remains `true`.
4. Given detected refill kit = `0`, writing `forceOn`:
   - `getRefillKitSettings()` returns `forceOn`;
   - detection remains `false`.
5. REST POST followed by GET round-trips the configuration value.
6. Writing an override does not mutate the detected value exposed through machine info.
7. Existing reconnect behavior results in configuration returning to `auto`.

Prefer testing through `UnifiedDe1` / a fake transport where practical rather than only testing enum conversion.

## `fromInt` hardening

The existing `De1RefillKitSettings.fromInt()` hardening may remain, but apply it to configuration parsing only.

Do not use configuration fallback behavior to reinterpret arbitrary raw detection values.

Detection should continue to derive the boolean using the machine register semantics currently used by Decaid:

```dart
(raw & 0x01) != 0
```

unless the firmware/reference implementation gives a stronger reason to change that.

## Scope

Keep this PR focused.

Do not:

- add WebSocket refill-kit state;
- persist refill-kit override across app restarts;
- add new settings storage;
- change firmware behavior;
- redesign water-level APIs.

The goal is simply to stop conflating **what the machine detected** with **what the client asked it to do**, and then make the API/docs accurately describe those two values.

Run:

```text
dart format lib test
flutter analyze
flutter test
```

and update the PR description to explain that the implementation now follows the `de1app` separation between detection and override.

allofmeng:
Fixed in a2d222c7 — you were right, the ambiguity was a bug and not just unclear docs.

The concrete failure: `getRefillKitSettings()` returned `De1RefillKitSettings.fromInt(_refillKit)`, and `onConnect()` never updated `_refillKit` after writing auto at `unified_de1.dart:289`. So straight after connect a machine with a kit reported `refillKitSetting: 1` (force on) and one without reported `0` (force off), while the register held `2`. Two of the new tests fail against the old code for exactly that.

Now split along the `de1app` line:

```dart
int _refillKitDetected = -1;                                   // connect-time MMR read only
De1RefillKitSettings _refillKitSetting = De1RefillKitSettings.auto;  // override
```

`extra.refillKit` stays `(_refillKitDetected & 0x01) != 0`; `_refillKitSetting` is set to auto after the connect-time write and updated only by `setRefillKitSettings()`; the GET serves the override without parsing a raw register.

8 tests through `UnifiedDe1` + `FakeBleTransport` (its `queueOnConnectResponses` already took `refillKitPresent`), covering items 2, 3, 4, 6 and 7. Item 5 was already covered by `round-trips through GET after write` in `de1handler_settings_reset_test.dart`, so I did not duplicate it. `fromInt` keeps `orElse: auto` but is now only reachable for configuration parsing.

Two things worth your call:

**Item 7 — reconnect does not return the configuration to auto.** `onConnect()` has `if (_info != null) return;` at `unified_de1.dart:256`, so reconnecting the same device object skips the MMR block entirely, write of auto included, and the override survives. Only a freshly discovered device resets. Tests pin both halves and the docs now say "lasts only until the machine is discovered and connected again". If you would rather the override actually be reapplied or reset on every transport reconnect, that is a behaviour change I did not want to make unasked.

**One thing beyond your list.** `POST .../advanced` used `firstWhere` with no `orElse` on `refillKitSetting` (`de1handler.dart:424`) and 500'd on anything outside 0/1/2. It now returns 400. With `fromInt` falling back to auto, leaving that boundary unvalidated would have turned `refillKitSetting: 7` into a silent auto. Happy to drop it if you would rather keep the diff to the model split.

`flutter analyze` clean. `flutter test`: 3341 pass, 6 fail — all in `webui_token_injection_test.dart` (`getWifiIP` fallbacks), and the same 6 fail on the base commit with my changes stashed.

tadelv (approved):

PR #685, opened 2026-08-25
feat(plugins): apply plugin settings immediately after save (gh-655)

### LINKED ISSUE #655: Plugin settings saved from Settings UI don't take effect until app restart
## Summary

Changing a plugin setting from the in-app Settings screen (e.g. enabling "Upload shots automatically" on `shot-upload.reaplugin`) persists the new value to disk, but the **already-running plugin instance never picks it up**. The setting shows as enabled everywhere you'd check it, but nothing actually happens until the plugin is reloaded (app restart, or a manual disable/enable).

## Repro

1. Enable `AutoUpload` for `shot-upload.reaplugin` via the app's plugin settings screen (Save).
2. Without restarting the app, pull one or more shots.
3. Observe: no upload happens, and no `[shot-upload]` log line is written at all (not even a "skipped" line — the `shotStored` handler's `if (state.autoUpload)` gate is silently false).

Confirmed on a live 0.8.0 device:

```
GET /api/v1/plugins/shot-upload.reaplugin/settings  ->  {"AutoUpload": true}   (persisted)
GET /api/v1/plugins/shot-upload.reaplugin/status    ->  {"autoUpload": false}  (live plugin state)
```

The log showed the plugin had not reloaded since the prior app boot, well before the setting was changed.

## Root cause

- The in-app Settings screen (`lib/src/settings/plugins_settings_view.dart`, Save button handler) calls `PluginLoaderService.savePluginSettings(pluginId, outgoing)` directly.
- `PluginLoaderService.savePluginSettings` (`lib/src/plugins/plugin_loader_service.dart`) only writes the new settings to storage — it never reloads the plugin or emits a `settingsUpdated` event to the running instance.
- Compare this to `POST /api/v1/plugins/<id>/settings` (`lib/src/services/webserver/plugins_handler.dart:_handlePluginSettingsPost`), which calls `savePluginSettings` **and then `reloadPlugin(id)`** — so changing a setting via the REST API takes effect immediately, but changing it via the app's own Settings UI does not.

## Impact

Any plugin whose behavior depends on a settings toggle (not just `shot-upload.reaplugin`) is affected: the UI implies the change is live, but stale in-memory state silently persists until the next app restart or a manual plugin disable/enable. For `shot-upload.reaplugin` specifically, this means a user can enable `AutoUpload` (or, as of the `DrainHistory` backlog-drain feature, `DrainHistory`) and see nothing happen, with zero error surfaced anywhere.

Related/compounding: plugin `host.log()` output is logged at `Logger.finest` severity (`plugin_manager.dart`, `_handleMessageSafely`), while the app's default log level is `INFO`. So even on a build where the bug above doesn't apply, plugin log lines (upload success/failure/skip) are invisible unless the log level is manually raised to `FINEST` first. Worth considering whether plugin logs should default to a more visible level, or whether the in-app log viewer should offer an easy way to include them.

## Suggested fix

Make `PluginLoaderService.savePluginSettings` (or its caller in `plugins_settings_view.dart`) trigger the same live-reload path the REST endpoint already uses, so both code paths behave consistently.

---
Filed by Claude on behalf of a user who lost a settings toggle to this on 0.8.0; verified the fix has not landed as of the 0.8.1 release (compared `plugins_settings_view.dart` / `plugin_loader_service.dart` between `v0.8.0` and `v0.8.1` — no changes to the save/reload path).

tadelv:
Fixes #655

## Problem

Saving plugin settings from the in-app Settings UI persisted them to disk but left the already-running plugin on its stale in-memory settings. `POST /api/v1/plugins/<id>/settings` worked around this by reloading the plugin in the handler, so REST saves took effect immediately while native UI saves did not — until app restart or a manual unload/load.

## Change

Centralize the apply step in `PluginLoaderService.savePluginSettings`:

- Persist and validate settings exactly as before (unchanged, under the existing per-plugin settings lock).
- Only after the settings lock is fully released, reload the plugin if it is currently loaded, so the new instance receives the persisted settings. The reload happens exactly once and is awaited before the save reports success; reload failures propagate instead of silently claiming the new settings are active.
- An unloaded plugin stays unloaded — saving never implicitly enables or loads a plugin.

The REST handler no longer reloads explicitly; both the native Settings UI and the REST endpoint share the single service-level save-and-apply path.

Locking: the settings write releases the settings lock before the reload enters the load path (`_pluginSettingsForLoad` re-acquires that lock), preserving the existing serialization guarantees for ordinary/secure settings, concurrent patches, and plugin mutation/reload operations. No locks were weakened or bypassed.

## Out of scope (per #655)

- `DrainHistory` retry-after-connection, `shot-upload.reaplugin` backlog-drain, plugin log severity/visibility, and a new `settingsUpdated` event are separate concerns and are not included.

## Tests

- Loaded plugin: saving settings reloads it exactly once and the new instance receives the new value (JS load counter via plugin storage).
- Unloaded plugin: save persists but leaves it unloaded.
- Concurrent saves on a loaded plugin complete without deadlocking.
- REST POST settings applies via the service-level reload (no duplicate reload); handler-contract test pins that the handler does not call `reloadPlugin` itself.

Verification: `flutter analyze` clean; full `flutter test` 3337 pass (remaining failures are pre-existing on `main` in this environment — webui token-injection and a known `mmr_codec` parallel-run flake).

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -2785,7 +2785,9 @@ paths:
       description: >
         Applies a settings patch. Omitted fields preserve the existing value;
         a null value clears the setting. For secure fields, an isSet state
-        object preserves the stored credential; null clears it. The response
+        object preserves the stored credential; null clears it. If the plugin
+        is loaded, it is reloaded so the running instance picks up the patch;
+        if it is unloaded, the patch is only persisted. The response
         is always redacted and never echoes a submitted credential.
       parameters:
         - name: id
--- a/doc/Api.md
+++ b/doc/Api.md
@@ -288,7 +288,7 @@ Settings fields include: `gatewayMode`, `themeMode`, `logLevel`, `weightFlowMult
 |--------|------|-------------|---------|
 | GET | `/api/v1/plugins` | List all plugins (with `loaded`, `autoLoad`, `source`, `pendingUpdate` fields) | `plugins_handler.dart` |
 | GET | `/api/v1/plugins/:id/settings` | Get plugin settings; secure fields return `{isSet}` only | |
-| POST | `/api/v1/plugins/:id/settings` | Patch plugin settings (omitted preserves, `null` clears) and return a redacted result | |
+| POST | `/api/v1/plugins/:id/settings` | Patch plugin settings (omitted preserves, `null` clears), reload the plugin if loaded, and return a redacted result | |
 | POST | `/api/v1/plugins/:id/enable` | Load plugin + enable auto-load | |
 | POST | `/api/v1/plugins/:id/disable` | Unload plugin + disable auto-load | |
 | DELETE | `/api/v1/plugins/:id` | Remove plugin (unload + delete files) | |
--- a/doc/Plugins.md
+++ b/doc/Plugins.md
@@ -586,7 +586,7 @@ Plugins can be managed via REST API:
 | POST | `/api/v1/plugins/update` | Check every GitHub-backed plugin for updates |
 | POST | `/api/v1/plugins/:id/update/approve` | Install an update that requests new permissions |
 | GET | `/api/v1/plugins/:id/settings` | Get plugin settings |
-| POST | `/api/v1/plugins/:id/settings` | Update plugin settings |
+| POST | `/api/v1/plugins/:id/settings` | Update plugin settings; reloads the plugin if it is loaded so the change applies immediately |
 
 `PUT /api/v1/plugins/:id/source` writes `manifest.json` and `plugin.js` for
 that id, creating the plugin if it is not installed yet:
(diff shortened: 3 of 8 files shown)


tadelv:
Addressed the review's locking concern: the loaded check now lives inside the per-plugin mutation lock.

`savePluginSettings` previously decided whether to reload via `isPluginLoaded` outside the mutation lock, then `reloadPlugin` re-checked under the lock and threw `Plugin not loaded` if a concurrent disable/unload won the lock in between — failing the save after settings were already persisted. The decision and the reload are now one serialized mutation operation (`_reloadPluginIfLoaded`), so either ordering (save's reload first, then disable; or disable first, then save skips the reload) is consistent. The public `reloadPlugin` keeps its strict throw-if-unloaded semantics for the manual reload button.

New regression test (`a disable racing a settings save does not fail the save`) holds the mutation lock inside disable's unload via a blocked `onUnload` storage write, runs the save concurrently, and asserts the save completes without error. Verified it fails against the previous code with the exact `Plugin not loaded: ...` exception the review predicted.

PR #693, opened 2026-08-26
fix(de1): bound transport operations

### LINKED ISSUE #687: Hardening: add bounded DE1 transport-operation governor
```dart
enum De1ReplayPolicy { never, sameMachine }

const _maxPendingDeviceWrites = 32;

// ponytail: count-bounded FIFO; add weighted cost only if #688 shows
// operation count does not adequately bound DE1 transport pressure.
```

## Goal

Replace `De1Controller`'s unbounded `Future` write chain with one small bounded DE1 queue.

Keep the governor inside `De1Controller`; it is already the shared choke point used by workflow/settings/profile/calibration handlers. Do not create a generic transport service, interface, scheduler framework, or shared DE1/scale queue.

DE1 only. Leave scale scheduling independent.

## 1. Replace `_deviceWriteQueue` with a bounded explicit FIFO

In `lib/src/controllers/de1_controller.dart`:

- import `dart:collection`;
- remove `_deviceWriteQueue`;
- add a `Queue<_PendingDe1Write<dynamic>>`;
- allow one active operation plus at most **32 pending** operations;
- add `maxPendingDeviceWrites` as a `De1Controller` constructor argument so tests can use tiny limits;
- drain the queue one operation at a time;
- after success or failure, continue draining;
- do not queue beyond the limit.

Keep the private queue entry minimal:

```text
call
completer
target machine identity
replay policy
optional coalescing key
```

Do not add origin IDs, estimated cost, priority numbers, deadlines, or public operation classes yet.

If the queue is full and the new write cannot replace an existing pending value, fail immediately with `De1WriteQueueFullException`.

Add that exception to the existing `lib/src/models/errors.dart`.

## 2. Stop replaying writes onto arbitrary replacement machines

Replace `retryOnReplacement` with:

```dart
De1ReplayPolicy replayPolicy = De1ReplayPolicy.never
```

Capture the intended machine identity when the operation is admitted:

1. usable non-zero DE1 serial number;
2. otherwise `deviceId`.

A queued operation must never silently migrate to a different physical machine.

For `sameMachine` operations:

- if the connection generation changes before/during execution, retry only when the reconnected DE1 has the same captured identity;
- if a different machine appears, fail the operation;
- do not blindly reuse the existing “first non-null replacement” behavior.

For `never`:

- any disconnect/generation change fails the operation;
- no replay.

Use `sameMachine` only for clearly replaceable settings added in the next step.

Leave calibration commands, LED commit/reset, firmware, state changes, and other imperative commands as `never`.

## 3. Add coalescing only for the settings that need it

Add one controller method:

```dart
Future<void> runReplaceableDeviceWrite(
  String key,
  Future<void> Function(De1Interface device) write,
)
```

Rules:

- only **pending** operations may be replaced;
- never replace/cancel the active operation;
- retain at most one pending operation for a key;
- replace the existing queue entry in place so repeated updates do not move that key to the back forever;
- complete the replaced caller with `De1WriteSupersededException`;
- the replacement receives its own future.

Add the superseded exception to `errors.dart`.

Use only these initial keys:

```text
workflow.rinse
workflow.steam
workflow.hotWater
```

Change:

- `updateFlushSettings`
- `updateSteamSettings`
- `updateHotWaterSettings`

to use the corresponding replaceable key.

Change `updateWorkflowSettings` so each changed component is scheduled independently using those same keys, then await the scheduled component futures together.

Do **not** make every DE1 setting coalescible in this PR. Everything else gets bounded FIFO behavior.

## 4. Keep the existing workflow revision logic

Do not invent a second desired-state/revision subsystem.

`WorkflowController` already has revisions and `setWorkflowIfRevision`, and `WorkflowHandler` already retries its merge if another workflow revision wins.

Keep that logic.

Also keep the existing bounded workflow request queue for now. The DE1 governor solves physical pending work; deleting the workflow queue is not required to land #687 safely.

Map:

- `De1WriteQueueFullException` → `503 Service Unavailable`;
- `De1WriteSupersededException` → `409 Conflict`.

HTTP workflow requests are already serialized, so supersession there should be uncommon; the explicit result mainly matters for app/plugin/controller callers.

## 5. Preserve idle/stop as the safety bypass

Add a small controller wrapper:

```text
requestMachineState(state)
```

Behavior:

- `MachineState.idle` → dispatch directly to the current DE1;
- every other state → normal imperative `runDeviceWrite`, `replayPolicy: never`.

Migrate direct normal state requests to this wrapper.

Keep existing stop-intent recording unchanged.

```dart
// ponytail: idle bypasses ordinary settings so queued work cannot starve stop.
```

Do not add a priority heap or reserved-capacity scheduler. Idle already has the simpler safety path.

Test that an ordinary DE1 write can be held open while `idle` still reaches the fake DE1 immediately.

## 6. Put firmware on the same serialized DE1 boundary

Change the firmware start path so the **entire** `updateFirmware()` future runs as one normal, non-replayable DE1 queue operation.

Because the DE1 queue is serialized, this automatically gives firmware exclusive access without inventing an “exclusive operation” abstraction:

- earlier DE1 operations finish first;
- firmware occupies the active slot until completion/cancellation;
- later ordinary writes remain bounded behind it.

Emit the existing `erasing` progress event when the queued firmware closure actually starts, not when it is merely accepted.

Keep `cancelFirmwareUpload()` direct so cancellation is not queued behind the firmware operation it needs to stop.

Do not replay firmware after reconnect.

## 7. Do not add a generic execution timeout

Do not wrap the active operation in `Future.timeout()` and then advance the queue.

Rules:

- caller timeout does not release the active DE1 slot;
- pending-count bounds prevent unlimited accumulation behind a stalled operation;
- idle/stop retains its direct safety path;
- existing transport/operation-specific timeouts remain responsible for operations that have a real bounded wait;
- if implementation inspection finds another genuinely unbounded low-level wait like the old `shotSettings.first`, bound that wait at its source instead of adding a scheduler timeout.

Keep the existing connection-generation check after operation completion.

## 8. Leave raw diagnostic writes alone

`/ws/v1/machine/raw` uses the synchronous/void `sendRawMessage()` API directly.

Do not change that API or invent acknowledgement semantics in #687.

The normal DE1 mutation paths become bounded; raw diagnostic traffic remains an explicit exception.

## 9. Add focused controller tests

Add `test/controllers/de1_controller_governor_test.dart`.

Reuse `TestDe1`, `Completer`, and the existing controller-test patterns; do not build a fake transport framework.

Cover:

### Queue bound

With `maxPendingDeviceWrites: 2`:

1. hold one active write;
2. enqueue two pending imperative writes;
3. the next write throws `De1WriteQueueFullException`;
4. release the active write;
5. both admitted writes execute in FIFO order.

### Replaceable coalescing

Hold one unrelated active operation.

Submit 100 rinse updates.

Verify:

- only one rinse operation remains pending;
- superseded callers receive `De1WriteSupersededException`;
- after releasing the blocker, only the final rinse value reaches the fake DE1.

### Independent keys

While blocked, submit repeated rinse and steam updates.

Verify the final rinse and final steam values both execute; neither key deletes the other.

### Caller timeout does not release serialization

1. start a held write;
2. let the caller apply `.timeout(...)`;
3. enqueue a second operation;
4. verify the second operation has **not** started after the caller times out;
5. release the first operation;
6. verify the second then starts.

### Different-machine replacement

Queue an operation for machine A, replace the connected machine with machine B before it starts, then release the blocker.

Verify the queued operation never executes on B.

### Same-machine reconnect

For a replaceable `sameMachine` operation, reconnect a new DE1 instance exposing the same identity and verify it may reconcile there.

### Imperative command

Verify a normal `runDeviceWrite` operation is not replayed after reconnect.

### Stop

Hold an ordinary write and request `idle`.

Verify the fake receives `idle` before the held ordinary write is released.

### Firmware

Hold an ordinary write and start firmware.

Verify firmware waits. Then hold firmware and verify a later normal write waits until firmware completes/cancels.

No soak test here; #688 owns sustained stress coverage.

## Validation

```sh
flutter test test/controllers/de1_controller_governor_test.dart
flutter test test/controllers/de1_controller_shotsettings_stall_test.dart
flutter test test/controllers/de1_controller_test.dart
flutter analyze
```

Also run the existing firmware/webserver test files touched by the change.

## Done when

- DE1 has one explicit FIFO with one active + at most 32 pending operations.
- Queue capacity cannot grow through repeated non-coalescible writes.
- Rinse/steam/hot-water have at most one pending value per key.
- Final accepted replaceable values reach the machine.
- Superseded callers receive an explicit result.
- Pending operations cannot migrate to another physical machine.
- Imperative commands are not replayed after reconnect.
- A caller timeout cannot cause overlapping DE1 writes.
- Idle/stop is not blocked by the ordinary queue.
- Firmware excludes ordinary DE1 operations for its duration.
- Scale code is untouched.
- No new package, governor service, weighted scheduler, or metrics subsystem is added.

Skipped: weighted costs, priority/fairness framework, generic execution cancellation/fencing, desired-state subsystem, metrics framework, and raw-WebSocket scheduling.

Add it when: #688 shows the 32-operation bound still overloads BLE, direct idle cannot dispatch during a real stalled transport operation, or raw writes measurably starve normal DE1 work.

ODevStudio:
## Summary

What changed, and why?

- Replace the unbounded DE1 write chain with one active operation and at most 32 pending operations.
- Coalesce pending rinse, steam, and hot-water settings while preserving final values and explicit supersession results.
- Make pending firmware updates cancellable and reject a full machine queue before committing an HTTP 200 response.
- Prevent replay onto a different machine, keep idle direct, and serialize firmware without coupling scale operations.

## Linked Issue

Fixes #687

## Verification

- Focused controller, sequencer, firmware, workflow, and no-scale handler tests passed.
- Pending and active firmware cancellation are covered through both DELETE and response-stream cancellation.
- Firmware queue saturation is covered at the HTTP boundary and returns 503 before streaming.
- The no-scale handler fixture now connects through the real controller state required by the governor and asserts the physical espresso request.
- Deterministic boundary tests cover failure recovery, coalescing position at capacity, identity fallback, saturated idle bypass, and queued firmware cancellation.
- Managed firmware eligibility is re-evaluated against the machine connected after catalog/image loading.
- Firmware cancellation covers the active-but-not-started initialization window.
- Automatic skip-step retries after queue rejection; manual rejection is surfaced in the UI.
- Multi-group workflow updates preflight all coalescing slots and reject atomically before any machine write is admitted.
- Queued sleep and scheduled-wake commands revalidate device and state at execution; heartbeats invalidate stale sleep work.
- User-present MMR writes now use the machine governor, and both Bengle bridges replace stale-generation retry timers.
- OpenAPI YAML parses successfully and documents machine-change/write failures for imperative mutations.
- Latest review regression suite: 152 passed (governor, shot sequencer, presence, both Bengle bridges, and workflow handler).
- `flutter analyze` completed with no issues.
- Full Windows `flutter test`: 3169 passed, 1 skipped, 220 failed in existing platform-dependent tests (CRLF-sensitive schema assertions and unavailable native QuickJS/runtime dependencies); the affected regression suite is green.
- Live smoke attempted with `sb-dev`; the helper is POSIX-only on Windows, and its documented direct-`flutter run` fallback is blocked on this host by the existing VS2019/universal_ble linker incompatibility before app startup.

## Impact

- Bounds pending machine transport work and rejects overflow instead of allowing unbounded growth.
- Adds documented 409 supersession and 503 queue-full responses for affected workflow/API paths.
- Firmware canceled while pending no longer starts later.
- Idle remains serviceable under ordinary write saturation; scale transport remains independent.

## Contributor Responsibility

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -358,6 +358,8 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/MachineStateError"
+        "503":
+          description: DE1 write queue is full for non-idle state changes
   /api/v1/machine/profile:
     post:
       summary: Set Machine Profile
@@ -375,7 +377,7 @@ paths:
         "400":
           description: Malformed JSON body
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "500":
           description: No machine was ever connected
   /api/v1/machine/shotSettings:
@@ -395,7 +397,7 @@ paths:
         "400":
           description: Malformed JSON body
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "500":
           description: No machine was ever connected
   /api/v1/machine/capabilities:
@@ -464,7 +466,7 @@ paths:
         "404":
           description: Machine connected but does not support cup warmer (not a Bengle)
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "500":
           description: No machine was ever connected
   /api/v1/machine/cupWarmer/preheat:
@@ -512,7 +514,7 @@ paths:
         "404":
           description: Machine connected but does not support pre-warm
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "500":
           description: No machine was ever connected
   /api/v1/machine/ledStrip:
@@ -564,7 +566,7 @@ paths:
         "404":
           description: Machine connected but does not support LED strip
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "500":
           description: No machine was ever connected
 
@@ -590,7 +592,7 @@ paths:
         "404":
           description: Machine connected but does not support LED strip
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "500":
           description: No machine was ever connected
 
@@ -621,7 +623,7 @@ paths:
         "404":
           description: Machine connected but does not support LED strip
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait, or firmware read failed
+          description: DE1 write queue is full, or firmware read failed
         "500":
           description: No machine was ever connected
   /api/v1/machine/scaleCalibration:
@@ -684,7 +686,7 @@ paths:
               schema:
                 $ref: "#/components/schemas/ScaleCalibrationRejected"
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "500":
           description: No machine was ever connected
   /api/v1/machine/settings:
@@ -715,7 +717,7 @@ paths:
         "400":
           description: Malformed JSON body
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "500":
           description: No machine was ever connected
 
@@ -747,7 +749,7 @@ paths:
         "400":
           description: Malformed JSON body
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "500":
           description: No machine was ever connected
 
@@ -763,7 +765,7 @@ paths:
         "202":
           description: Reset accepted (one grouped serialized device write)
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "500":
           description: No DE1 connected
 
@@ -801,7 +803,7 @@ paths:
         "400":
           description: Malformed JSON body
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "500":
           description: No machine was ever connected
 
@@ -878,7 +880,7 @@ paths:
         "400":
           description: Malformed JSON body, unknown target, or invalid values
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "504":
           description: Machine did not acknowledge the calibration write within the response timeout
         "500":
@@ -900,7 +902,7 @@ paths:
         "400":
           description: Malformed JSON body
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "500":
           description: No machine was ever connected
 
@@ -1338,6 +1340,12 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+        "409":
+          description: A pending DE1 settings write was superseded by a newer value
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
         "413":
           description: Request body exceeds the 1 MiB workflow payload limit
           content:
@@ -1359,9 +1367,9 @@ paths:
         "503":
           description: >
             The request waited more than 30 seconds for its turn and was
-            skipped, or the machine disconnected mid-write and no replacement
-            machine appeared within the bounded wait (10 seconds). In both
-            cases the controller workflow is unchanged.
+            skipped, the DE1 write queue was full, or a replaceable setting
+            could not reconnect to the same machine within the bounded wait.
+            In each case the controller workflow is unchanged.
           content:
             application/json:
               schema:
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -926,7 +926,8 @@ paths:
       description: |
         Send a raw firmware file to the machine. Returns an NDJSON progress
         stream. Rejects an empty body with 400. Returns 409 if an update is
-        already in progress. Returns 503 if no machine is connected.
+        already in progress. Returns 503 if no machine is connected or the
+        DE1 write queue is full.
 
         A successful stream emits `erasing`, zero or more `uploading` events,
         then `done`. A failed stream emits `erasing`, zero or more `uploading`
@@ -956,7 +957,7 @@ paths:
         "409":
           description: Firmware update already in progress
         "503":
-          description: No machine connected
+          description: No machine connected or DE1 write queue full
     delete:
       summary: Cancel an in-progress firmware update
       description: |
@@ -1011,7 +1012,7 @@ paths:
         "422":
           description: Artifact validation, compatibility, or version-policy failure
         "503":
-          description: No machine connected
+          description: No machine connected or DE1 write queue full
 
   # Scale API
   /api/v1/scale/tare:
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -359,7 +359,7 @@ paths:
               schema:
                 $ref: "#/components/schemas/MachineStateError"
         "503":
-          description: DE1 write queue is full for non-idle state changes
+          description: Machine write queue is full for non-idle state changes
   /api/v1/machine/profile:
     post:
       summary: Set Machine Profile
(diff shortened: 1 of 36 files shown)


lib/src/realtime_steam_feature/realtime_steam_feature.dart
```diff
@@ -121,8 +121,10 @@ class _RealtimeSteamFeatureState extends State<RealtimeSteamFeature> {
           .connectedDe1()
           .shotSettings
           .first;
-      await _de1Controller.connectedDe1().updateShotSettings(
-        currentSettings.copyWith(targetSteamDuration: _steamDuration),
+      await _de1Controller.runDeviceWrite(
```
tadelv: Please make this read-modify-write atomic with respect to the governor. `currentSettings` is captured before this request waits for its queue turn, so an earlier queued shot-settings write can run first and then be overwritten here by this stale full `De1ShotSettings` snapshot (except for duration). Move the shot-settings read inside the queued callback (or add a controller helper that does so), and cover the ordering with a regression test.

assets/api/rest_v1.yml
```diff
@@ -375,7 +377,7 @@ paths:
         "400":
           description: Malformed JSON body
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
         "500":
           description: No machine was ever connected
```
tadelv: Please update the `500` contract here and for the other imperative mutation responses changed in this PR. With `De1ReplayPolicy.never`, a disconnect/generation change before or during execution now fails through the generic 500 path; the updated tests explicitly expect 500 for that case. `No machine was ever connected` is therefore no longer the full meaning of this response. Because this spec is authoritative for clients/agents, document machine-change/write failure as well.

tadelv (changes requested):
Two required fixes before this lands:

1. Make the realtime steam `+10s` read-modify-write happen inside the governor turn so a queued write cannot apply a stale full shot-settings snapshot over an earlier admitted update.
2. Update the OpenAPI `500` response descriptions for imperative DE1 mutations to match the new non-replay behavior: disconnect/generation-change failures now also reach the generic 500 path.

Also, per `AGENTS.md` / `.agents/skills/decent-app/verification.md`, please run and record the required `sb-dev` REST smoke (`start` → `status` → exercise affected endpoint(s) → `reload` → re-exercise → inspect logs → `stop`). CI is green for format, analyze, full `flutter test`, and build smoke, but it does not demonstrate that API smoke.

assets/api/rest_v1.yml
```diff
@@ -395,7 +397,7 @@ paths:
         "400":
           description: Malformed JSON body
         "503":
-          description: Machine disconnected mid-write and no replacement appeared within the bounded wait
+          description: DE1 write queue is full
```
tadelv: `Machine` not `DE1`. We're getting new machines

lib/src/realtime_steam_feature/realtime_steam_feature.dart
ODevStudio: Fixed in cd91270b. The controller helper now reads and writes shot settings inside one governor turn, with an ordering regression test in de1_controller_governor_test.dart.

assets/api/rest_v1.yml
ODevStudio: Fixed in 1e59c08b. All queue wording added by this PR now uses Machine rather than DE1, including firmware and workflow descriptions.

assets/api/rest_v1.yml
ODevStudio: Fixed in 1e59c08b. The affected imperative mutation 500 responses now cover no connection, machine replacement during the write, and write failure.

ODevStudio:
Addressed the latest governor hardening review in 54b8cc4d:

- workflow multi-group updates now preflight coalescing capacity atomically;
- queued sleep and scheduled-wake commands revalidate applicability at execution;
- heartbeat/deferred presence MMR writes now go through the machine governor;
- Bengle SAW and steam-stop reconnects replace stale-generation retry timers.

Regression coverage was added for all four paths. The final focused suite passed 152 tests and flutter analyze is clean. Full Windows flutter test reached 3169 passed, 1 skipped, and the same 220 existing platform-dependent failures documented in the PR body.

tadelv (approved):
Re-reviewed current head 54b8cc4. The previous blockers are addressed: the steam +10s path now performs its shot-settings read/modify/write within one governor turn with regression coverage, and the OpenAPI 500 contracts document machine-change/write failures. I also checked the added stale-operation hardening around queued sleep/wake/skip/presence/bridge writes and did not find a new required fix. Current PR CI is green for format, analyze, full flutter test, and Linux build smoke. The sb-dev live smoke could not run on the contributor's Windows host, but the attempted verification and existing environment blockers are documented; I don't consider that a merge blocker here.

PR #694, opened 2026-08-26
fix(api): bound request and websocket admission

### LINKED ISSUE #686: Hardening: add bounded API and streaming admission control
## Goal

Bound accidental pressure from buggy/chatty local skins, plugins, reconnect loops, slow clients, and oversized requests without changing DE1/scale command semantics.

Keep this process-local and dependency-free.

## 1. Add one small admission gate

Add `lib/src/services/webserver/admission_control.dart` as part of the existing webserver library.

Implement one concrete `AdmissionGate` using only Dart collections/counters:

- fixed 1-second accepted-request window;
- global active count;
- per-client active count;
- global rate count;
- per-client rate count;
- maximum tracked client entries;
- opportunistically remove inactive client entries after their rate window expires;
- client key = existing `clientIpFromRequest(request)`;
- no forwarded-header/client-identity framework.

Use permissive LAN defaults as private/source constants:

```text
API:
  global concurrent: 128
  per-client concurrent: 32
  global accepted/sec: 1024
  per-client accepted/sec: 256
  tracked clients: 256

WebSocket:
  global open: 128
  per-client open: 32
  global upgrades/sec: 128
  per-client upgrades/sec: 32
  tracked clients: 256
```

Make the clock a tiny injectable `nowMs` callback so rate-window tests do not sleep.

```dart
// ponytail: fixed process-local limits; make configurable only if normal LAN clients actually hit them.
```

Do not add token-bucket/sliding-window machinery or a package dependency.

## 2. Put API admission in the existing Shelf pipeline

In `lib/src/services/webserver_service.dart`, add one middleware after the existing auth middleware and inside the existing CORS/logging middleware.

Apply it only to `/api/` requests.

Skip:

- `OPTIONS`;
- static/WebUI requests;
- `/ws/` upgrades, which get their own gate below.

Behavior:

- acquire before invoking the API handler;
- always release in `finally`;
- do not queue saturated requests;
- per-client limit → `429 Too Many Requests`;
- global limit → `503 Service Unavailable`;
- include `Retry-After: 1`.

If grep confirms nothing currently emits `X-RateLimit-*`, remove those stale names from the CORS exposed-header list and expose `Retry-After` instead.

Do not add quota/reset headers unless there is an actual stable contract for them.

## 3. Bound WebSocket connections and reconnect churn

Add one small wrapper around the existing `sws.webSocketHandler`.

Use a separate `AdmissionGate` instance so long-lived sockets do not consume normal API request slots.

The wrapper must:

1. acquire before the upgrade;
2. return `429` for the per-client limit or `503` for the global limit;
3. attach release to `channel.sink.done`;
4. release immediately if upgrade setup fails or throws;
5. release exactly once.

Use it for every current WebSocket registration.

Because these files are parts of the same webserver library, reuse the concrete helper directly. Do not add an interface/factory.

Do **not** add a custom per-socket message queue or telemetry coalescer in this issue.

## 4. Bound request bodies only where they are actually read

Do not create a route-policy/body-policy framework.

Keep `WorkflowHandler`'s existing 1 MiB / 30-second bounded body reader and its pending-request queue. The workflow queue currently has semantic serialization responsibilities and belongs with #687 before it can be simplified.

Fix the currently unbounded raw firmware upload in `firmware_handler.dart`:

- reject `Content-Length` above the firmware cap before reading;
- also count streamed bytes so chunked uploads cannot exceed the cap;
- apply an overall body-read timeout;
- return `413` when too large;
- return `408` when body reading times out;
- continue producing the same byte list the existing validator/flasher expects.

Choose the firmware ceiling from an existing protocol/fixture/known firmware-size bound if one exists. If the repository contains no authoritative ceiling, use **16 MiB** as a simple source constant with a **60-second** body-read timeout.

Before finishing, grep the repository for request-body reads (`readAsString`, `request.read`, `toList`) and patch only any other caller that currently buffers an unbounded body.

Do not switch firmware uploads to temp files/streaming storage unless firmware images can legitimately exceed the bounded in-memory ceiling.

## 5. Do not touch device scheduling

This issue must not:

- change `runDeviceWrite` or DE1 write ordering;
- coalesce workflow fields;
- create a DE1/scale global queue;
- move scale operations behind the DE1 governor;
- change reconnect/replay semantics;
- change stop-command priority.

Those belong to #687.

Shared HTTP/WebSocket admission may cover both DE1 and scale endpoints; their physical transport execution remains independent.

## 6. Add focused tests

Reuse the existing webserver test structure.

### `AdmissionGate`

Test directly with small limits and a fake `nowMs`:

- per-client concurrent limit and release;
- global concurrent limit;
- per-client/global rate limit;
- one-second window reset;
- tracked-client map remains bounded and expired inactive entries are removed.

### API middleware

Add one integration test proving:

- saturation returns the expected `429`/`503`;
- `Retry-After` is present;
- releasing work allows the next request through;
- CORS still applies to the rejection.

### WebSocket wrapper

Using an existing WebSocket service test:

- reach the configured connection limit;
- next upgrade is rejected;
- close one connection;
- next upgrade succeeds.

No stress/soak suite here; #688 owns that.

### Firmware body

Test:

- body over the cap → `413`;
- streamed/chunked body crossing the cap → `413`;
- the normal firmware path still reaches existing validation below the cap.

## Validation

Run only the affected tests first, then:

```sh
flutter test test/services/webserver_service_test.dart
flutter test test/services/webserver_service_visualizer_snapshot_test.dart
flutter analyze
```

Run any existing firmware-specific test file touched by the change.

## Done when

- API requests have hard global/per-client concurrent and one-second rate bounds.
- No admission queue can grow.
- Rate-limit client bookkeeping has a hard cap.
- All current WebSockets use the shared connection/upgrade gate.
- Closing/failing a WebSocket releases its slot.
- Raw firmware upload cannot buffer an unlimited request body.
- Existing workflow bounds remain intact.
- No dependency, policy framework, metrics subsystem, or device scheduler is added.

Skipped: custom WebSocket send queues, route cost classes, configurable policy framework, and DE1/scale scheduling changes.

Add it when: bounded connections still show sustained slow-reader heap growth, or bounded concurrency still leaves normal API p95 latency above 200 ms during the #688 soak test.

ODevStudio:
## Summary

What changed, and why?

- Add dependency-free process-local admission gates for REST requests and WebSocket upgrades/connections.
- Bound every buffered API request body through one shared reader with 64 KiB control and 1 MiB ordinary JSON limits.
- Preserve Shelf request charset decoding while enforcing byte and read-time limits.
- Cap buffered raw firmware uploads at 1 MiB; bundled DE1 images are about 464 KiB.
- Close upgraded sockets after synchronous or asynchronous connection-setup failure and release admission only after closure.
- Use a process-lifetime monotonic clock for admission windows.
- Keep the existing workflow and data-transfer streaming limits unchanged.
- Restack on #693 at `54b8cc4` and retain its machine write queue/error contract wording.

This PR is stacked on #693 because API admission is defense in depth over the bounded machine transport boundary.

## Linked Issue

Fixes #686

## Verification

- 36 focused admission and firmware tests passed, including an isolated global accepted-rate window reset regression.
- `flutter analyze --no-pub` completed with no issues.
- Full Windows suite completed with 3198 passing and 1 skipped; 220 existing platform-dependent failures remain, chiefly CRLF assertions and unavailable native QuickJS/runtime dependencies.
- Windows live smoke was attempted with `flutter run --dart-define=simulate=1` but could not launch: Visual Studio 2019 selected CMake 3.20 while Firebase requires 3.22+.
- Android live smoke was also attempted on the connected tablet with ADB forwarding, but Gradle failed before compilation with `java.io.IOException: Unable to establish loopback connection` under both JDK 21 and JDK 17. No `curl`/`websocat` result is claimed.

## Impact

- Saturated local API and WebSocket clients receive `Retry-After: 1` with 429 or 503 rather than consuming unbounded service capacity.
- Buffered request bodies return 413 when oversized and 408 when their read deadline expires.
- Buffered JSON bodies are capped at 64 KiB or 1 MiB by route; raw firmware uploads are capped at 1 MiB.
- Request text still honors the charset declared in `Content-Type`, matching Shelf's prior behavior.
- Failed WebSocket setup closes the upgraded connection without leaking or undercounting admission slots.
- REST and AsyncAPI admission behavior is documented; no dependency or migration is added.

## Contributor Responsibility


test/webserver/firmware_handler_test.dart
```diff
@@ -95,22 +95,23 @@ final class _BlockingFirmwareBundle extends CachingAssetBundle {
   }
 }
 
+late _FixedController _controller;
```
tadelv: `AGENTS.md` has a hard rule against adding new global state without a clear ownership boundary. This fixture used to be scoped inside `main()` and is now mutable library-global only so `_firmwareHandler` can reach it. Please keep the controller test-scoped instead (for example, pass it into `_firmwareHandler` or keep the helper within `main()`).

tadelv (changes requested):
Requesting changes for two AGENTS.md adherence items:

1. This API/spec change still needs the required smoke verification via `scripts/sb-dev.sh` with `curl`/`websocat`. The PR verification records focused/full tests and analyze, but not that smoke test. Please run it and add the evidence/results here before this is considered done.
2. Please address the inline mutable-global-state comment in the firmware handler test.

I did not find another blocking correctness issue in the admission/body-limit implementation; the API/AsyncAPI and `doc/Api.md` updates are present and the WebSocket/body-reader coverage looks consistent with #686.

ODevStudio:
Restacked onto #693 at 54b8cc4 and pushed f92bf311. The OpenAPI conflict retains the machine write queue/error wording while adding the 408/413 and 1 MiB contract. Also corrected AI_API_NOTES.md to 1 MiB, added the isolated global-window reset regression, and moved the firmware controller into main() with explicit helper injection. Focused tests: 36 passed. Analyze: clean. Full Windows suite: 3198 passed, 1 skipped, 220 existing platform-dependent failures. Live smoke was attempted on Windows and Android but both hosts failed before app launch; exact CMake and Gradle/JDK blockers are recorded in the PR body. Fresh CI is running against the restacked head.

tadelv (approved):

PR #695, opened 2026-08-26
test(hardening): cover pressure regressions

### LINKED ISSUE #688: Hardening: add pressure, coalescing, reconnect, and streaming stress coverage
```bash
flutter test test/services/webserver/admission_control_pressure_test.dart
flutter test test/controllers/de1_controller_governor_pressure_test.dart
flutter test test/services/webserver/pressure_integration_test.dart
flutter analyze
```

## Goal

Prove the concrete bounds introduced by #686 and #687 hold under runaway/chatty local clients, reconnects, stalled operations, and repeated setting updates.

This issue should add tests and one simulator scenario. Do not add another scheduler, limiter, metrics framework, benchmark dependency, or general-purpose fake transport framework.

Reuse existing `fakeAsync`, `TestDe1`, `TestScale`, controllable discovery helpers, simulated devices, and the repository's existing deterministic-test / simulator-scenario split.

## 1. Add API pressure tests

Add:

```text
test/services/webserver/admission_control_pressure_test.dart
```

Test the concrete admission gate from #686 directly with **small injected limits** so tests are fast and exact.

### One noisy client

Configure:

```text
global concurrent: 4
per-client concurrent: 2
global rate/sec: 8
per-client rate/sec: 4
tracked clients: 8
```

Fire 100 requests without sleeping.

Assert:

- active requests never exceed 4 globally;
- one client never exceeds 2;
- excess requests receive the documented `429`/`503`;
- no admission wait queue exists;
- after held requests complete, active count returns to zero.

Do not test arbitrary throughput numbers against wall-clock timing.

### Two clients

Hold requests from client A at its per-client limit.

Send a request from client B.

Assert B can still enter while global capacity remains.

This is enough for the fairness requirement; do not build weighted fair queuing tests because #686 does not add a fair scheduler.

### High-cardinality identities

Submit requests from 100 unique client keys against `maxTrackedClients: 8`.

Assert:

```dart
expect(gate.trackedClientCount, lessThanOrEqualTo(8));
```

Advance the injected/fake clock beyond expiry, touch the gate again, and verify expired entries are removed.

### Rate-window reset

Use the injected clock from #686.

Do not use real `Future.delayed`.

Fill the rate budget, advance exactly past the one-second window, then verify admission resumes.

## 2. Add one API integration pressure test

Add:

```text
test/services/webserver/pressure_integration_test.dart
```

Use the existing webserver test setup rather than constructing a second Shelf stack.

Send a short burst through the real middleware:

```text
GET /api/v1/...
PUT /api/v1/workflow
```

Use injected low limits.

Assert:

- admitted requests complete;
- overflow returns the expected status;
- `Retry-After` is present;
- CORS headers remain present on rejected API responses;
- a rejection does not make the next valid request fail after capacity is released.

Do not attempt to simulate thousands of real TCP connections inside `flutter test`.

## 3. Add DE1 governor pressure tests

Add:

```text
test/controllers/de1_controller_governor_pressure_test.dart
```

Reuse `TestDe1` plus `Completer<void>` gates.

### 1000 replaceable writes, one key

Configure:

```text
maxPendingDeviceWrites: 4
```

Hold one unrelated DE1 operation active.

Submit **1000 rinse updates**.

Assert before releasing the active operation:

```text
active operations: 1
pending rinse entries: 1
total pending operations <= 4
```

Then release the blocker.

Assert:

- only the final rinse value is physically written;
- physical rinse writes from the 1000-call burst are exactly **1** after the blocker;
- all replaced futures complete with `De1WriteSupersededException`;
- the final caller completes successfully;
- no operation future remains pending.

Do not assert fuzzy wording such as “substantially fewer writes.”

### Multiple keys

While blocked, alternate 100 updates each to:

- rinse;
- steam;
- hot water.

Assert there is at most one pending entry for each key.

After release:

- final rinse value applied;
- final steam value applied;
- final hot-water value applied;
- no field disappears because another key was coalesced.

## 4. Regression-test workflow partial patches

Extend the existing workflow handler test file rather than creating another workflow-specific suite if it already has suitable fixtures.

Submit interleaved partial patches such as:

```json
{"steamSettings":{"duration":12}}
{"rinseData":{"flow":3.0}}
{"steamSettings":{"duration":16}}
{"hotWaterData":{"volume":120}}
```

Assert the final `WorkflowController.currentWorkflow` contains:

```text
steam duration = 16
rinse flow = 3.0
hot-water volume = 120
```

Also assert the final DE1 writes match those values.

Keep the current workflow revision/`setWorkflowIfRevision` implementation; #688 only verifies it.

## 5. Prove caller timeout does not release DE1 serialization

Test:

1. start an ordinary DE1 write and hold it with a `Completer`;
2. apply `.timeout(...)` to the caller's returned future;
3. let that caller observe `TimeoutException`;
4. enqueue another normal DE1 operation;
5. verify the second physical operation has **not** started;
6. release the original operation;
7. verify the second then executes.

The key assertion is physical execution order, not merely which future completed.

## 6. Test reconnect identity semantics

Keep these as separate tests.

### Same machine

Use machine A, disconnect, then reconnect a new `TestDe1` instance with the same machine identity.

Queue a replaceable setting permitted to use `sameMachine`.

Assert:

- stale device instance receives no later write;
- latest desired setting applies to the reconnected same machine;
- caller completes once.

### Different machine

Queue work targeting machine A.

Disconnect and connect machine B with a different serial/device identity.

Assert:

- pending work for A does not execute on B;
- imperative work is not replayed;
- the operation completes with the declared disconnected/replacement failure.

### Stale completion

Start a held write on A.

Disconnect A and connect a replacement.

Allow A's old future to complete afterward.

Assert:

- its result cannot publish applied state for the new generation;
- it does not trigger another write on the replacement;
- a later valid operation on the replacement executes normally.

### Flapping

Do a small deterministic sequence:

```text
A connect
A disconnect
A reconnect
A disconnect
A reconnect
```

with a few queued operations.

Assert all futures terminate and pending count returns to zero.

Do not build a randomized/property-testing harness.

## 7. Test stop/idle under saturation

Fill the normal DE1 queue:

```text
1 active held write
max pending normal writes
```

Then request:

```dart
MachineState.idle
```

Assert the fake DE1 receives `idle` **before** the active ordinary write is released.

Use an event/call-order assertion rather than wall-clock milliseconds:

```text
calls == ['ordinary-start', 'idle']
```

That is a stronger deterministic assertion than claiming a 50 ms latency on CI.

Do not create a stalled-transport reset test unless #687 actually introduces a transport-reset mechanism.

## 8. Test firmware exclusivity

Reuse existing firmware test fixtures.

Hold an ordinary DE1 operation and start firmware.

Assert firmware has not started.

Release ordinary work and start firmware.

While firmware is held:

- submit a normal setting write;
- assert it does not execute;
- cancel/complete firmware;
- assert the setting then executes.

Also verify firmware is never replayed after a disconnect.

No throughput benchmark is needed.

## 9. Test DE1 and scale independence

Add one small regression test using `TestDe1` and `TestScale`.

Do not invent a common device governor.

### Stalled scale

Hold a scale operation if the existing scale abstraction provides a controllable async operation.

Execute a DE1 write.

Assert the DE1 write completes without releasing the scale operation.

### Stalled DE1

Hold a DE1 write.

Execute one existing scale operation, such as tare, that can be tested safely.

Assert the scale operation completes.

If the current `TestScale` cannot conveniently block an operation, add the minimum override to the test-local subclass rather than modifying production scale scheduling.

## 10. Exercise WebSocket limits, not hypothetical buffering

Extend the WebSocket test file created/touched by #686.

With small injected limits:

```text
global connections: 3
per-client connections: 2
```

Assert:

1. client A opens two sockets;
2. A's third is rejected;
3. client B opens one socket;
4. another global connection is rejected;
5. close one socket;
6. a new socket succeeds;
7. connection count returns to zero after all sockets close.

Also fill the handshake-rate limit using the fake clock and verify it resets.

Do **not** add slow-reader/latest-frame buffer tests unless #686 actually implements a custom bounded outbound buffer.

```dart
// ponytail: connection bounds only; add outbound buffering if soak evidence
// shows slow readers can grow process memory.
```

## 11. Add one simulator pressure scenario

Add:

```text
.agents/skills/decent-app/scenarios/api-pressure-hardening.md
```

Use the repo's existing simulate-mode workflow rather than adding a benchmark executable.

Scenario:

1. start Decaid with `scripts/sb-dev.sh` in simulate mode;
2. keep the machine snapshot WebSocket open;
3. send a finite burst of workflow mutations, alternating steam/hot-water/rinse values;
4. periodically call the machine-state/settings GET endpoint;
5. request idle during the burst;
6. verify:
   - API remains responsive;
   - WebSocket continues producing telemetry;
   - some excess requests may be rejected according to #686;
   - final workflow value equals the last admitted update;
   - simulated machine remains connected;
   - idle still works.

Use a **finite reproducible burst**, for example:

```bash
for i in $(seq 1 500); do
  # PUT alternating workflow settings
done
```

Do not create an endless soak script or CI job.

Document an optional manual longer run for hardware/simulator validation, but do not make normal CI wait minutes.

## 12. Keep observability assertions proportional

If #686/#687 expose counters such as:

```text
active requests
tracked clients
open WebSockets
pending DE1 operations
coalesced operations
```

assert those counters return to zero/expected values after each test.

If they do not expose them, add only `@visibleForTesting` getters on the concrete classes needed to prove the bounds.

Do not add Prometheus/OpenTelemetry/log aggregation solely for #688.

For logs, add at most one regression asserting a 1000-update coalescing burst does **not** emit one warning/error per superseded request.

Normal debug/fine logging need not be counted.

## 13. Explicit cleanup assertion

Every pressure test must finish with the relevant resources empty:

```dart
expect(gate.activeCount, 0);
expect(gate.openWebSocketCount, 0);
expect(controller.pendingDeviceWriteCount, 0);
```

For lists of returned futures:

```dart
await Future.wait(
  futures.map((f) => f.catchError((_) {})),
);
```

No test may finish while deliberately leaving held `Completer`s, sockets, subscriptions, or queued operations unresolved.

## Validation

Run the new focused suites first:

```bash
flutter test test/services/webserver/admission_control_pressure_test.dart
flutter test test/services/webserver/pressure_integration_test.dart
flutter test test/controllers/de1_controller_governor_pressure_test.dart
```

Then the child-issue regressions:

```bash
flutter test test/controllers/de1_controller_shotsettings_stall_test.dart
flutter test test/services/webserver
flutter analyze
```

Finally:

```bash
flutter test
```

Run the simulator scenario once manually before marking #688 complete.

## Done when

- A 1000-update single-key burst retains at most one pending replaceable value.
- Queue/request/WebSocket/client-map limits are asserted with exact integers.
- Final rinse/steam/hot-water values are correct after coalescing.
- All superseded/rejected operations terminate explicitly.
- Caller timeout cannot release an active physical DE1 operation.
- Same-machine reconnect and different-machine replacement are separate passing tests.
- A stale old-generation completion cannot affect the current DE1.
- Idle bypasses saturated ordinary DE1 work.
- Firmware excludes ordinary DE1 writes while active.
- Stalled DE1 and scale paths do not block each other.
- WebSocket connection/handshake limits are quantitatively tested.
- Every test finishes with zero orphaned requests, sockets, or DE1 queue entries.
- One finite simulate-mode pressure recipe reproduces the #678 class of chatty-client behavior without depending on Streamline.js.

Skipped: benchmark framework, randomized fuzzing, production metrics backend, custom WebSocket buffering tests, and long-running CI soak jobs.

Add it when: the finite simulator run exposes heap growth, unfair scheduling, transport starvation, or a failure that deterministic tests cannot reproduce.

ODevStudio:
## Summary

What changed, and why?

- Add deterministic pressure coverage for exact API, WebSocket, client-map, and DE1 queue bounds.
- Cover 1000-write coalescing, timeout serialization, reconnect identity, stale completions, idle, firmware exclusivity, and DE1/scale independence.
- Exercise API pressure through the production middleware composition and real info/workflow route registration.
- Exercise overlapping workflow mutations while the first physical write is blocked and the workflow revision changes.
- Add a finite simulator pressure scenario for the chatty-client behavior reported in #678.
- Bound every scenario network call, require the WebSocket process to remain alive, and require telemetry to advance during the burst.
- Retry idle and final recovery only for documented admission responses; fail immediately on other statuses.
- Prove process-global and per-client API/WebSocket rate windows reset independently.
- Assert pressure tests drain admission and DE1 resources before completion.

This draft is stacked on #686 and #687 because it verifies the concrete contracts introduced by both implementation changes.

## Linked Issue

Fixes #688

## Verification

- Focused governor, MockDe1 pressure, admission, bounded-body, and production middleware run: 64 tests passed.
- All Bash blocks in the pressure scenario pass `bash -n` using Git Bash.
- `flutter analyze --no-pub` completed with no issues.
- Current USB-tablet simulator run: 500/500 requests accounted for, all 20 control probes remained responsive, telemetry grew from 28 to 222 snapshots while the socket stayed alive, final workflow values matched, idle succeeded, and MockDe1 remained connected.
- Full Windows suite reached 3199 passing and 1 skipped; 220 unrelated checkout/environment failures remain, dominated by the unavailable `quickjs_c_bridge.dll` and CRLF-sensitive fixtures.

## Impact

- Adds a testable entry point for the existing production middleware composition while preserving auth, body-error, admission, logging, and CORS ordering.
- No dependency or migration is added.

## Contributor Responsibility

### DIFF
--- a/.agents/skills/decent-app/SKILL.md
+++ b/.agents/skills/decent-app/SKILL.md
@@ -49,6 +49,7 @@ Pick the scenario that matches the task, run it verbatim, and finish before call
 | Account-proxy write forwarding + write-scope gate | `scenarios/account-proxy-write.md` |
 | Account-proxy native consent gate | `scenarios/account-proxy-consent.md` |
 | Plugin Decent-account proxy bridge (host.decentProxy) | `scenarios/plugin-decent-proxy.md` |
+| API pressure hardening | `scenarios/api-pressure-hardening.md` |
 
 ## Authoritative sources
 
--- a/.agents/skills/decent-app/scenarios/api-pressure-hardening.md
+++ b/.agents/skills/decent-app/scenarios/api-pressure-hardening.md
@@ -80,6 +80,7 @@ test "$WS_BASELINE" -gt 0
   done
   wait
 )
+kill -0 "$WS_PID"
 test "$(wc -l < /tmp/decaid-pressure-snapshot.jsonl)" -gt "$WS_BASELINE"
 ```
 
--- a/.agents/skills/decent-app/scenarios/api-pressure-hardening.md
+++ b/.agents/skills/decent-app/scenarios/api-pressure-hardening.md
@@ -14,7 +14,7 @@ WS=ws://localhost:8080
 Confirm the simulated machine is connected:
 
 ```bash
-curl -sf "$BASE/api/v1/devices" \
+curl -sf --max-time 2 "$BASE/api/v1/devices" \
   | jq -e '.[] | select(.name == "MockDe1" and .state == "connected")'
 ```
 
@@ -46,7 +46,7 @@ test "$WS_BASELINE" -gt 0
     esac
 
     (
-      code=$(curl -s -o /dev/null -w '%{http_code}' \
+      code=$(curl -s --max-time 35 -o /dev/null -w '%{http_code}' \
         -X PUT "$BASE/api/v1/workflow" \
         -H 'content-type: application/json' \
         -d "{\"$field\":{\"$key\":$value}}")
@@ -105,7 +105,7 @@ admission rejection, then prove all three fields reached the workflow.
 ```bash
 final_ok=false
 for attempt in $(seq 1 20); do
-  if curl -sf -X PUT "$BASE/api/v1/workflow" \
+  if curl -sf --max-time 35 -X PUT "$BASE/api/v1/workflow" \
     -H 'content-type: application/json' \
     -d '{"steamSettings":{"duration":47},"hotWaterData":{"volume":121},"rinseData":{"flow":3.25}}' \
     >/dev/null; then
@@ -116,12 +116,12 @@ for attempt in $(seq 1 20); do
 done
 test "$final_ok" = true
 
-curl -sf "$BASE/api/v1/workflow" | jq -e '
+curl -sf --max-time 2 "$BASE/api/v1/workflow" | jq -e '
   .steamSettings.duration == 47 and
   .hotWaterData.volume == 121 and
   .rinseData.flow == 3.25'
-curl -sf "$BASE/api/v1/machine/state" | jq -e '.state.state == "idle"'
-curl -sf "$BASE/api/v1/devices" \
+curl -sf --max-time 2 "$BASE/api/v1/machine/state" | jq -e '.state.state == "idle"'
+curl -sf --max-time 2 "$BASE/api/v1/devices" \
   | jq -e '.[] | select(.name == "MockDe1" and .state == "connected")'
 ```
 
--- a/.agents/skills/decent-app/scenarios/api-pressure-hardening.md
+++ b/.agents/skills/decent-app/scenarios/api-pressure-hardening.md
@@ -25,10 +25,17 @@ at most 64 concurrent clients. The idle request is made during the burst and
 retried only when generic API admission rejects it.
 
 ```bash
-rm -f /tmp/decaid-pressure-status /tmp/decaid-pressure-snapshot.jsonl
+rm -f /tmp/decaid-pressure-status /tmp/decaid-pressure-probes \
+  /tmp/decaid-pressure-snapshot.jsonl
 websocat "$WS/ws/v1/machine/snapshot" \
   > /tmp/decaid-pressure-snapshot.jsonl &
 WS_PID=$!
+for attempt in $(seq 1 20); do
+  test -s /tmp/decaid-pressure-snapshot.jsonl && break
+  sleep 0.1
+done
+WS_BASELINE=$(wc -l < /tmp/decaid-pressure-snapshot.jsonl)
+test "$WS_BASELINE" -gt 0
 
 (
   for i in $(seq 1 500); do
@@ -46,7 +53,10 @@ WS_PID=$!
       printf '%s %s %s %s\n' "$i" "$field" "$value" "$code" \
         >> /tmp/decaid-pressure-status
       if ((i % 25 == 0)); then
-        curl -sf "$BASE/api/v1/machine/state" >/dev/null || true
+        probe_code=$(curl -s --max-time 2 -o /dev/null -w '%{http_code}' \
+          "$BASE/api/v1/machine/state")
+        printf '%s %s\n' "$i" "$probe_code" \
+          >> /tmp/decaid-pressure-probes
       fi
     ) &
 
@@ -57,17 +67,20 @@ WS_PID=$!
     if ((i == 100)); then
       idle_ok=false
       for attempt in $(seq 1 20); do
-        if curl -sf -X PUT "$BASE/api/v1/machine/state/idle" >/dev/null; then
-          idle_ok=true
-          break
-        fi
-        sleep 1
+        code=$(curl -s --max-time 2 -o /dev/null -w '%{http_code}' \
+          -X PUT "$BASE/api/v1/machine/state/idle")
+        case "$code" in
+          200) idle_ok=true; break ;;
+          429|503) sleep 1 ;;
+          *) printf 'unexpected idle status: %s\n' "$code" >&2; exit 1 ;;
+        esac
       done
       test "$idle_ok" = true
     fi
   done
   wait
 )
+test "$(wc -l < /tmp/decaid-pressure-snapshot.jsonl)" -gt "$WS_BASELINE"
 ```
 
 The run is finite. Successful and rejected requests must account for all 500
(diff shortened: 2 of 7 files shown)


ODevStudio:
Restacked the 11 pressure-regression commits onto current #694 at f92bf311 and force-pushed 65cf664c. git range-diff confirms all 11 patches are content-identical to the previously reviewed series. The restacked composition passes 57 focused pressure/workflow tests, dart formatting is unchanged, and flutter analyze is clean. GitHub now reports #693, #694, and #695 mergeable; fresh #695 CI is running.

tadelv (approved):

PR #698, opened 2026-08-26
fix(account): verify stored Decent credentials and invalidate stale auth state

### LINKED ISSUE #696: fix(account): verify stored Decent credentials and invalidate stale auth state
## Problem

Decaid currently treats the presence of a stored Decent account email as equivalent to being authenticated.

`DecentAccountService.isLoggedIn()` is currently:

```dart
Future<bool> isLoggedIn() async => await _store.read(key: 'email') != null;
```

This means stale or invalid stored credentials can leave Decaid showing **Logged In** indefinitely even though authenticated Decent API calls return `401`.

We observed this while debugging `shot-upload.reaplugin`: the plugin successfully reaches `DecentProxyService`, but `/support/api/shot_upload` repeatedly returns `401`. A local reproduction also confirms that deliberately invalid credentials can coexist with UI state that still reports the account as logged in.

The REST status endpoint has the same semantics because `/api/v1/account/decent` simply returns `isLoggedIn()`.

## Relevant code

- `lib/src/services/account/decent_account_service.dart`
  - `login()` validates via `/support/api/login_test`
  - `isLoggedIn()` only checks whether an email exists in the credential store
  - stored password is the encrypted `cryptpw` returned by `login_test`
- `lib/src/services/account/decent_proxy_service.dart`
  - reads stored email + encrypted password
  - sends them as Basic Auth to Decent API endpoints
  - currently returns upstream `401` without updating account state
- `lib/src/account/account_page.dart`
  - renders **Logged In** directly from `isLoggedIn()`
- `lib/src/services/webserver/account_handler.dart`
  - `/api/v1/account/decent` exposes the same boolean
- `test/services/account/decent_account_service_test.dart`
  - currently tests that previously stored credentials imply `isLoggedIn() == true`
  - does not cover stale/invalid stored credentials

## Desired behavior

Persisted credentials should mean **account linked / credentials present**, not automatically **authenticated**.

Decaid should have a reliable way to determine whether stored credentials are still valid and should stop reporting an authenticated/logged-in state after a definitive upstream authentication failure.

### Required behavior

1. Introduce clear semantics between:
   - credentials/account are present locally; and
   - credentials have been accepted by the Decent backend.

2. Add a credential-validation path using an inexpensive authenticated Decent endpoint.
   - Prefer reusing an endpoint whose success unambiguously proves that the stored `email + cryptpw` are valid.
   - Do not expose raw credentials outside `DecentAccountService` / `DecentProxyService`.

3. Handle definitive authentication failure consistently.
   - An upstream `401` from an authenticated Decent request must mark the stored authentication as invalid.
   - UI and REST account status must no longer continue to report **Logged In** after that condition.
   - Decide centrally whether invalidation means deleting credentials immediately or retaining them in an explicit invalid/unverified state. Prefer the simplest model that gives correct behavior without duplicating auth state across services.

4. Preserve good credentials on a failed replacement login.
   - If a user is already linked with working credentials and submits a different bad email/password in a new login attempt, that failed attempt should not destroy the existing valid credentials unless the user explicitly unlinks the account.

5. Keep `DecentProxyService` and the account UI/REST state consistent.
   - A plugin/API request receiving a real authentication failure should not leave the account page claiming the account is authenticated.

## Suggested implementation direction

Keep the source of truth in the account service / credential layer rather than adding plugin-specific handling.

A reasonable shape would be:

- rename or replace the current local-presence check with something explicit such as `hasLinkedAccount()` / `hasStoredCredentials()`;
- add a validation method for the stored credentials;
- make the UI and `/api/v1/account/decent` consume auth-aware state rather than the current email-presence check;
- give `DecentProxyService` a narrow callback/service hook to report an upstream `401` so account state is invalidated centrally;
- avoid validating credentials on every widget rebuild or every local status poll if a cached/known-valid state can be maintained safely;
- preserve current security properties: plugins never receive credentials, and proxy logging must never include Authorization headers, plaintext passwords, or stored cryptpw values.

Exact naming/state representation is intentionally left to the implementation as long as callers cannot confuse `credentials exist` with `credentials are valid`.

## Regression tests

Add tests covering at least:

- no stored credentials -> not authenticated;
- successful `login_test` -> credentials persisted and account authenticated;
- stored valid credentials -> validation succeeds and account reports authenticated;
- stored stale/invalid credentials -> validation fails and account does not report authenticated;
- authenticated proxy request receives `401` -> account auth state is invalidated;
- after proxy `401`, `AccountPage` no longer renders **Logged In**;
- after proxy `401`, `/api/v1/account/decent` reports `loggedIn: false` (or equivalent new auth-status contract if the API is intentionally expanded);
- failed replacement login does not erase previously valid stored credentials;
- logout/unlink still clears credentials and auth state;
- no credentials or auth material are exposed through REST/plugin APIs or logs.

Update the existing `isLoggedIn` tests that currently encode `email exists == logged in` as expected behavior.

## Acceptance criteria

- [ ] Decaid no longer treats a stored email alone as proof of authentication.
- [ ] Stored Decent credentials can be actively validated against the backend.
- [ ] A definitive upstream `401` invalidates the authenticated state centrally.
- [ ] Flutter account UI and REST account status agree on the resulting state.
- [ ] Shot upload and other Decent-proxy consumers naturally benefit from the same fix; no shot-upload-specific credential workaround is introduced.
- [ ] Failed attempts to replace credentials do not unnecessarily destroy a previously valid account link.
- [ ] Regression tests cover stale credentials, proxy `401`, UI/REST consistency, successful login, failed replacement login, and logout.
- [ ] No sensitive credentials are logged or exposed to plugins/REST clients.

## Non-goals

- Changing the Decent backend authentication scheme.
- Moving credentials into plugin settings.
- Adding plugin-specific auth state.
- Broadly redesigning account/token UI beyond what is necessary to represent valid vs invalid/unlinked state correctly.

tadelv:
## Summary

Decaid previously treated the presence of a stored Decent account email as equivalent to being authenticated. `isLoggedIn()` only checked `store.read('email') != null`, so stale or invalid credentials left the UI and `/api/v1/account/decent` reporting **Logged In** indefinitely while authenticated API calls returned 401 (observed with `shot-upload.reaplugin`).

This PR separates *linked* from *authenticated*:

- `hasLinkedAccount()` — credentials present locally (email + encrypted password stored)
- `isLoggedIn()` — auth-aware: validates stored `email + cryptpw` against `login_test` once per session (memoized, so UI/REST polls never re-validate per request), then returns the cached in-memory state
- `verifyStoredCredentials()` — validation path; 200 + non-`0` = valid, 401 or `0` = definitively invalid, 5xx/network = indeterminate (state untouched)
- `reportAuthenticationFailure()` — `DecentProxyService` reports upstream 401s via a narrow `onAuthFailure` hook; auth state is invalidated centrally. Credentials are kept on 401 so a transient backend failure does not destroy the link; the next session re-validates
- Fixed `login()` accepting the backend's rejection body `"0\n"` (verified live: `login_test` returns `200` with body `0\n` for wrong credentials — the untrimmed `body != '0'` comparison treated it as success, storing wrong credentials and showing **Logged In**)

Failed replacement logins never touch stored credentials, so a bad re-login cannot erase a working link.

## Linked Issue

Fixes #696

## Verification

- Full suite: 3392 tests passed, 1 skipped; `flutter analyze` clean
- New/updated regression tests: no stored credentials, successful login persists + authenticates, stored valid credentials validate, stored stale credentials rejected, proxy 401 invalidates auth state, AccountPage no longer renders **Logged In** after a proxy 401, `/api/v1/account/decent` reports `loggedIn: false` after invalidation, failed replacement login preserves valid stored credentials, logout clears, one validation per session, real backend `"0\n"` rejection body; indeterminate (network/5xx) validation retries on the next check while definitive rejection stays sticky
- Manually verified against the live backend: wrong credentials via Basic auth and via the documented `email+cryptpw` param form both return `200 "0\n"`

## Impact

- REST contract unchanged in shape (`{loggedIn}`), semantics now auth-aware — `doc/Api.md` and `assets/api/rest_v1.yml` updated
- No credentials or auth material exposed to plugins/REST clients; proxy logging unchanged
- Rationale documented in `doc/AI_STORAGE_NOTES.md`

## Contributor Responsibility

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -5133,9 +5133,14 @@ paths:
 
   /api/v1/account/decent:
     get:
-      summary: Decent account link status
+      summary: Decent account auth status
       description: >
-        Reports whether a Decent Espresso account is currently linked.
+        Reports whether the linked Decent Espresso account is currently
+        authenticated. `loggedIn` is true only when the stored credentials
+        have been accepted by the Decent backend (validated via
+        `login_test` on first status read per session); a definitive upstream
+        401 invalidates the state until the next successful login or
+        re-validation. Stored credentials being present is not sufficient.
         Linking and unlinking are performed only through the app's native UI,
         not over the network, so no login/logout routes are exposed here. The
         linked email is deliberately omitted — it is PII and this endpoint is
@@ -5143,7 +5148,7 @@ paths:
       tags: [Account]
       responses:
         "200":
-          description: Current link status
+          description: Current auth status
           content:
             application/json:
               schema:
@@ -5152,7 +5157,7 @@ paths:
                 properties:
                   loggedIn:
                     type: boolean
-                    description: True if a Decent account is linked
+                    description: True if the linked Decent account is authenticated
 
   /api/v1/account/proxy/support/api/{endpoint}:
     get:
--- a/doc/AI_STORAGE_NOTES.md
+++ b/doc/AI_STORAGE_NOTES.md
@@ -51,6 +51,10 @@ Gotchas:
 
 Keep these stores independent. A settings reset must not clear account credentials unless explicitly requested.
 
+## Account Auth State (gh#696)
+
+`DecentAccountService` stores `email` + encrypted password (`cryptpw`, returned by `login_test`) in the secure store. Stored credentials mean **linked**, not **authenticated**: `hasLinkedAccount()` is the presence check; `isLoggedIn()` is auth-aware and returns a cached in-memory state, validated against `/support/api/login_test` (memoized while in flight; only definitive results are sticky — an indeterminate network/5xx outcome sets a 30s cooldown (`retryInterval`) so widget rebuilds and status polls cannot hammer `login_test`, and the next check after the cooldown retries). `login_test` accepts `pw` or `cryptpw` and returns `0` (body `"0\n"`, status 200 — always trim before comparing) on rejection, otherwise the cryptpw. A definitive rejection (401 or `"0"`) marks the account not authenticated; 5xx/network errors are indeterminate and leave the cached state untouched. `DecentProxyService` reports upstream 401s via `onAuthFailure` -> `reportAuthenticationFailure()`. Credentials are kept on 401 — a transient backend failure must not destroy the link; the next session re-validates. Auth state carries a generation counter: login success, logout, and reported auth failures bump it, and an in-flight validation only writes its result if its generation is still current, so a stale validation of old credentials cannot clobber a newer successful login (or resurrect auth after a 401). Failed `login()` attempts never touch stored credentials (a bad replacement login cannot erase a working link).
+
 Plugin settings marked `secure` in the manifest are stored through
 `CredentialStore` under a plugin-scoped key. Ordinary plugin settings remain in
 SharedPreferences. Reads exposed to UI and REST clients return only `{isSet}`
--- a/doc/Api.md
+++ b/doc/Api.md
@@ -448,7 +448,7 @@ archive is also bounded by the 2 GiB import request limit.
 
 | Method | Path | Description | Handler |
 |--------|------|-------------|---------|
-| GET | `/api/v1/account/decent` | Decent account link status: `{loggedIn}` | `account_handler.dart` |
+| GET | `/api/v1/account/decent` | Decent account auth status: `{loggedIn}` (true only when stored credentials are accepted by the backend) | `account_handler.dart` |
 | GET | `/api/v1/account/proxy/<path>` | Auth-enriching proxy to `decentespresso.com/<path>` | `account_proxy_handler.dart` |
 | POST | `/api/v1/account/proxy/<path>` | Auth-enriching write proxy (relays body) | `account_proxy_handler.dart` |
 | PUT | `/api/v1/account/proxy/<path>` | Auth-enriching write proxy (relays body) | `account_proxy_handler.dart` |
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -445,6 +445,7 @@ void main(List<String> args) async {
       credentialStore: credentialStore,
       requireConsent: gate.requireConsent,
       baseUrl: decentBaseUrl,
+      onAuthFailure: () => decentAccountService!.reportAuthenticationFailure(),
     );
     accountTokensController = AccountTokensController(
       tokenService: proxyTokenService,
--- a/lib/src/services/account/decent_account_service.dart
+++ b/lib/src/services/account/decent_account_service.dart
@@ -1,4 +1,6 @@
 import 'dart:convert';
+
+import 'package:clock/clock.dart';
 import 'package:http/http.dart' as http;
 
 List<String> parseSerialNumbers(String body) {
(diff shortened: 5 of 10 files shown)


tadelv:
Addressed — retry is now throttled by a service-level cooldown, decoupled from widget rebuilds.

**What changed** (`isLoggedIn` in `decent_account_service.dart`):

- Indeterminate (network/5xx) results keep the account unauthenticated but start a 30s cooldown (`retryInterval`, injectable). Checks inside the window return the cached result with no network call; after expiry the next check re-validates.
- The cooldown uses `clock.now()` (package:clock, already a dependency) so it follows fake time in tests.
- Definitive results (200-valid, 401, `"0"`) remain sticky via the cached auth state; the in-flight memo still coalesces concurrent calls.

**About the specific loop you predicted:** instrumenting the widget test showed the `FutureBuilder`-in-`build()` loop cannot occur on its own. When the future completes, `FutureBuilder` calls its own `setState` and rebuilds only its own subtree — `AccountPage.build()` (and therefore `isLoggedIn()`) is not re-invoked. Empirically: with a 500-returning client and the memo cleared immediately (no cooldown), the page settles after exactly **1** network request. The `future:` argument is only re-evaluated when the page rebuilds for another reason (ancestor setState, navigation, etc.).

**The real hazard** is those other rebuild sources, and the cooldown guards them. New widget regression test `parent rebuilds during an outage do not hammer the backend`:

- Wraps the page in a host that rebuilds it every second (new widget instance per rebuild, like real app rebuilds) against a 500 client
- 10 seconds of 1 Hz rebuilds → exactly **1** request (throttled)
- Pumping past the 30s cooldown → exactly **1** retry (requests == 2)
- Negative check: with the cooldown disabled the same test fails with **11** requests

Service tests updated/added: `throttles retries after an indeterminate failure` (2 immediate calls, 1 request) and `retries validation after an indeterminate network failure` (with `retryInterval: Duration.zero`). Full suite: 3396 passed, 1 skipped; `flutter analyze` clean.

tadelv:
Addressed — the stale-write race is fixed with an auth-state generation counter.

**The guard** (`decent_account_service.dart`):

- `_authGeneration` starts at 0; `login()` success, `logout()`, and `reportAuthenticationFailure()` each bump it.
- `verifyStoredCredentials()` captures the generation when it starts and writes `_authenticated` only through `_setAuthenticated(generation, value)`, which no-ops when the generation has moved on. A stale validation returns the *current* state instead of its own result, so callers holding the in-flight future also see the correct state.
- `logout()` now also clears `_authenticated` (previously redundant, now load-bearing for the stale-return path).

**Regression tests** (`concurrent auth updates`, both use a `Completer` to pin the ordering):

1. *Stale validation does not clobber a newer successful login* — validation of old credentials hangs; `login(B)` succeeds; the stale validation then completes with a rejection. Assertions: login authenticated, `isLoggedIn()` still true, store holds B. Without the guard this fails (`Expected: true, Actual: false`).
2. *In-flight validation does not resurrect auth after an upstream failure* — validation hangs; `reportAuthenticationFailure()`; validation then completes 200. Assertions: validation returns false, `isLoggedIn()` false. Without the guard this fails (`Expected: false, Actual: true`).

Both negative checks were verified by temporarily removing the guard. Full suite: 3398 passed, 1 skipped; `flutter analyze` clean.

Issue #301, opened 2026-06-08
Account proxy: harden — pin :8080 proxy CORS to skin origin


tadelv:
## Goal
Defense-in-depth: restrict `Access-Control-Allow-Origin` on the `/api/v1/account/proxy/*` path to the known skin origin(s) instead of the global `*`, so a malicious origin can't even read proxy responses if it somehow obtains a token.

## Where it fits
Hardening on top of the Decent account auth-proxy — see **#296**. Branch `feature/account-proxy`, design in `doc/plans/account-proxy-design.md` ("CORS posture").

## Context
Token secrecy already rests on the `:3000` CORS lockdown (shipped in #296 — the skin token can't be scraped cross-origin). This issue adds a second layer on the `:8080` side. The global API CORS must stay permissive (existing skins depend on `*` for the rest of the API), so this is a **path-scoped** override only.

## Reuse these
- Pipeline + `corsHeaders` setup in `lib/src/services/webserver_service.dart` (the proxy middleware is already path-scoped there).
- `WebUIService` knows the LAN IP via `network_info_plus` (`lib/src/webui_support/webui_service.dart`) — use it to build the allowlist (`http://<app-ip>:3000` + `localhost`/`127.0.0.1` variants).

## Scope
- For `/api/v1/account/proxy/*` only: set `Access-Control-Allow-Origin` to the specific skin origin(s) (echo if in allowlist), not `*`. Handle the preflight (`OPTIONS`) accordingly.
- Leave all other API paths on the existing permissive CORS.

## Open question
Exact allowlist entries — just `<app-ip>:3000` + loopback, or also mDNS/`*.local` hostnames? (carried over from the design doc).

## Acceptance criteria
- [ ] A request to the proxy path from an allowed skin origin gets a matching ACAO and works.
- [ ] A request from an unlisted origin does not receive a permissive ACAO for the proxy path.
- [ ] Non-proxy API paths unchanged.

## Tests
- Unit: CORS header on proxy path for allowed vs disallowed origin.

## Deps
Builds on #296. Low priority — token secrecy is already covered by the `:3000` lockdown.

PR #700, opened 2026-08-26
fix(webui): accept LAN hostnames that resolve to this device

### LINKED ISSUE #699: WebUI entry on port 3000 404s when browsed by LAN hostname
Since 0.8.2 serves each skin generation on a random port, port 3000 is the stable entry that 307-redirects to the live generation. That redirect (and the skin-api injection) only answer when the request's \`Host\` string-matches \`localhost\` or a literal device interface IP — \`_isLocalHost\` in \`lib/src/webui_support/webui_service.dart\` does a string \`contains\` against \`NetworkInterface.list\` addresses.

Browsing by a LAN name (router DNS, mDNS, a hosts file entry) therefore 404s at the entry:

```
$ curl -sD - http://decent:3000/          # name resolves to the tablet
HTTP/1.1 404 Not Found

$ curl -sD - http://10.0.0.76:3000/       # same device by IP
HTTP/1.1 307 Temporary Redirect
location: http://10.0.0.76:41059/
```

Everyone who bookmarked their tablet by name (the pre-0.8.2 \`http://<name>:3000\` worked directly) lost access with the upgrade. And because the same check gates skin-api injection, a page reached under a hostname would also be served without the injected script/token, so skins that rely on it (e.g. Beanie's gateway discovery) degrade even past the entry.

Expected: a Host that resolves — via the device's own resolver — to one of the device's interface addresses should be treated as local, same as the literal IP. Names resolving elsewhere should stay rejected.

Repro: decaid 0.8.2 on an M50 Mini tablet, Android 14; any client whose DNS maps a name to the tablet.

giladger:
## Summary

What changed, and why?

- Since the 0.8.2 per-generation skin ports, the port-3000 entry redirect reuses the same strict host check as token injection, so browsing the WebUI by a LAN name 404s before anything is served. Pre-0.8.2 the entry served those requests (as static files, without injection).
- Reachability and trust are now separate decisions. The entry redirect and the skin-api script tag accept a hostname that resolves to an address currently assigned to the device; the account-proxy token stays restricted to loopback and literal device addresses.
- The token restriction is deliberate, not an oversight in the relaxation: `doc/AI_API_NOTES.md` records that hostnames are rejected to preserve the DNS-rebinding boundary, and a rebound name resolves to a device address by construction. The first revision of this PR missed that and would have handed the token to a rebound origin; it now gets only what the pre-0.8.2 entry already served it, plus a tokenless `window.decentApp` that no-ops outside the embedded webview.
- Resolutions are cached as addresses rather than as a local/not-local verdict, under a short TTL, and re-checked against the live interface list on every request — so a lookup that recovers, an address that moves, or an interface that disappears reclassifies without a restart (addressing @ODevStudio's review). The cache stays bounded so spoofed Host headers cannot grow it or force a lookup per request.
- Explanatory comments removed per `AGENTS.md`; rationale moved to `doc/AI_API_NOTES.md`, skin-facing behavior to `doc/Skins.md` (addressing @tadelv's review).

## Linked Issue

Fixes #699

## Verification

- `flutter test test/webui_support/webui_token_injection_test.dart` — 24/24. New cases: a resolving hostname gets the script but no token; a literal device address still gets the token; a hostname that resolves elsewhere is rejected; the entry redirect follows a resolving hostname to the current generation; a resolver that starts failing and then recovers reclassifies once the entry expires; a cached resolution is rejected after the device's interface address changes.
- Full `flutter test` — 3390 pass, 31 fail. All 31 failures are shot-upload plugin tests that fail identically on a clean `main` without this branch (missing `assets/plugins/shot-upload.reaplugin/` locally); none touch `webui_support`.
- `flutter analyze` — clean apart from the same pre-existing missing-asset-directory warning. `dart format lib test` — no changes.
- Hardware: reproduced the original failure on decaid 0.8.2 (M50 Mini, Android 14) — `curl http://decent:3000/` → 404 while `curl http://10.0.0.76:3000/` → 307 to the live generation.

## Impact

- User-visible: browsing the WebUI by a LAN name works again under the new entry-redirect model.
- Security: the DNS-rebinding boundary is unchanged — the account-proxy token still requires loopback or a literal device address, and `doc/AI_API_NOTES.md` now states why in terms of rebinding rather than of hostnames generally. A skin that needs the token must be opened by loopback or device IP; one that only needs `window.decentApp` also works by name. Account-proxy CORS pinning (#301) is untouched and still excludes hostname origins.
- Docs: `doc/Skins.md` (skin-facing behavior) and `doc/AI_API_NOTES.md` (rationale) updated.

## Contributor Responsibility


ODevStudio:
Nice fix overall. The hostname resolution approach looks appropriate for the intended trusted-LAN environment, and the added tests cover the main local/non-local cases well.

One thing I’d consider changing is the caching in _resolvedHosts. Both successful and failed resolutions are cached for the lifetime of the served generation. That means a transient DNS/mDNS lookup failure can cause a valid hostname to remain rejected even after name resolution recovers, and conversely a hostname whose address changes keeps its previous classification.

I’d prefer either not caching the boolean result, or using a short-lived cache of resolved addresses and checking those against the current interface addresses on each request. A regression test where the resolver result changes between requests would be useful here.

Otherwise this looks good to me.

Separately, CI is currently blocked by the contribution-policy check for #699 rather than by a code/test failure.

tadelv (changes requested):
Two remaining items to address:

1. Please update `doc/Skins.md`. This PR changes skin-facing behavior: LAN hostnames can now pass the WebUI entry redirect and receive the skin-api injection. Both `AGENTS.md` and `CONTRIBUTING.md` require `doc/Skins.md` to be updated when skin behavior changes.

2. Please remove the new explanatory/rationale comments in `lib/src/webui_support/webui_service.dart` (the cache rationale and `_resolvesToLocalAddress` history). `AGENTS.md` explicitly asks new/substantially rewritten code not to add explanatory comments, and to put rationale/debugging history in the matching docs instead.

giladger:
Thanks both. Reworked, with one correction worth calling out.

`doc/AI_API_NOTES.md` records the hostname rejection as the DNS-rebinding boundary, and the first revision of this PR would have handed the account-proxy token to a rebound origin: a rebound name resolves to a device address by construction, so resolving the Host and trusting the result defeats exactly the control that note describes.

Reachability and trust are separate decisions now. The port-3000 entry redirect and the skin-api script tag accept a hostname that resolves to an address currently assigned to the device; the token stays restricted to loopback and literal device addresses. A rebound origin gets what the pre-0.8.2 entry already served it, plus a tokenless `window.decentApp` that no-ops outside the embedded webview.

On the caching: resolutions are cached as addresses rather than as a local/not-local verdict, under a short TTL, and re-checked against the live interface list on every request, so a lookup that recovers or an address that moves reclassifies without waiting for a restart. The cache stays bounded so spoofed Host headers cannot grow it or force a lookup per request. Both regression cases are covered - a resolver that fails and then recovers, and a cached resolution invalidated by the device's interface address changing.

Comments dropped per `AGENTS.md`; rationale moved to `doc/AI_API_NOTES.md` and the skin-facing behavior to `doc/Skins.md`.

Verification: `test/webui_support/webui_token_injection_test.dart` 24/24; full `flutter test` 3390 passing, with 31 shot-upload plugin failures that reproduce identically on a clean `main` here (missing `assets/plugins/shot-upload.reaplugin/` locally) and touch nothing in `webui_support`; `flutter analyze` clean apart from the matching pre-existing warning.

tadelv (approved):

PR #709, opened 2026-08-27
feat(plugins): show Decent account status

ODevStudio:
## Summary

- Shows credential-validated Decent account status in native settings for plugins requesting `proxy.decent_api` or `proxy.decent_api.write`, including plugins with no manifest settings.
- Distinguishes authenticated, unauthenticated, and indeterminate validation results so network failures do not appear as rejected credentials.
- Offers an Account settings action only for rejected or missing credentials; opening it closes the plugin dialog without saving pending settings.

## Linked Issue

N/A - requested through the attached product task.

## Verification

- `flutter test test/services/account/decent_account_service_test.dart test/plugins_settings_view_appstore_test.dart` - 66 passed.
- `flutter analyze` - no issues.
- Full `flutter test` exercised 3,425 passing tests and 1 skipped test. The unrelated `connect_result_schema_test.dart` assertion still fails because it expects LF while this Windows checkout reads the AsyncAPI file as CRLF.
- `dart format` passed for the changed Dart files.
- `git diff --check` passed.
- GitHub contribution policy, change detection, analyze/test, and Linux build-smoke checks passed.

## Impact

- User-visible: Decent-account proxy plugins show Logged In, Not Logged In, or Account status unavailable in their settings. Account-only proxy plugins receive a Close-only dialog.
- Security: exposes only the validation state; no email or credentials are shown.
- Compatibility, migration, API/spec, external plugin release, and documentation impact: none.

## Contributor Responsibility


tadelv (changes requested):
One edge case to fix before merge: preserve a previously known-invalid account state when a fresh validation is indeterminate.

If the proxy has already received a 401, `reportAuthenticationFailure()` sets `_authenticated = false`, and `DecentProxyService` will continue blocking requests as known-invalid. But `verifyStoredCredentialsStatus()` currently returns `indeterminate` unconditionally on a network error or non-definitive server response. That makes the plugin settings UI show “Account status unavailable” and hides the Account settings remediation action even though the credentials are still known-invalid.

Please have the indeterminate paths preserve an existing `_authenticated == false` as `DecentAccountStatus.unauthenticated`; otherwise return `indeterminate` as they do now. This keeps transient failures from turning unknown/valid state into logged-out while retaining a definitive prior rejection.

Please add a regression test covering: stored credentials → `reportAuthenticationFailure()` → `login_test` throws or returns 500 → `verifyStoredCredentialsStatus()` remains `unauthenticated`. A widget test confirming the Account settings action remains available in that state would also be useful.

Everything else in the PR looks good.

ODevStudio:
Thank you for catching this edge case, @tadelv. Fixed in 750cdad3: both indeterminate validation paths now preserve a previously definitive `unauthenticated` state, while previously valid or unknown state still reports `indeterminate`. I added regression coverage for stored credentials -> `reportAuthenticationFailure()` -> network failure; the existing widget test confirms the Account settings action remains available for an unauthenticated result. Verified with the focused service/widget tests, `flutter analyze`, and the full suite (3,526 passed, 1 skipped).

tadelv (approved):

PR #718, opened 2026-08-28
feat(profiles): bundle Adaptive v3 profile linked to v2 parent

tadelv:
## Summary

What changed, and why?

- Add **Adaptive v3** as `assets/defaultProfiles/best_practice_v3.json`, converted from de1app-nightly's `best_practice.tcl` (commit `1a61f9b4` "Adaptive v3"). Deltas vs v2: Pressurize exit pressure 8.8 -> 7.7 bar, extraction `max_flow_or_pressure` limiter 9.5 -> 8.6.
- Keep **Adaptive v2** (`best_practice.json`) bundled alongside, per the Damian LRv2/LRv3 precedent.
- Bundle provenance: v3 JSON carries `"parent_id": "profile:b73ce54592d667d00b31"` — the immutable content-hash id of the Adaptive v2 we ship. After seeding, `ProfileController` validates that the referenced parent record exists and applies the fixed `parentId`, so `GET /api/v1/profiles/{id}/lineage` shows `v2 -> v3` and lineage stays stable even if the v2 file's content changes in a future release.

## Linked Issue

N/A

## Verification

- `flutter test test/profiles/` — 12/12 pass, including the P1 test asserting `v3.parentId == v2.id`, lineage `[v2, v3]`, and that the bundled `parent_id` equals the hash computed from the bundled v2 file (guards against future v2 content drift).
- `flutter analyze` — clean (one pre-existing unrelated asset warning).
- `mock_replay_de1` / `workflow_device_sync` suites pass.
- Full suite: 31 failures, all in `shot_upload_*`/`bundled_plugin_permissions_test.dart`, reproduced identically on a clean tree (pre-existing, unrelated to profiles).
- Converter round-trip: `tools/ingest_profiles.py` output for v3 verified against bundled JSON.

## Impact

- User-visible: new "Adaptive v3" default profile; v2 remains. Version lineage visible via profile lineage endpoint/UI.
- No API/spec change (lineage endpoint already exists).
- `doc/Profiles.md` updated to document the `parent_id` key.

## Contributor Responsibility


tadelv (commented):
One improvement I'd like before merging: please make bundled profile lineage reference an immutable parent profile ID directly rather than resolving the parent dynamically from a filename.

For example, use something like `"parent_id": "profile:…"` in the bundled JSON instead of `"parent": "best_practice.json"`.

The current filename-based approach resolves to whatever content hash that filename has on each startup. If `best_practice.json` changes in a future release, v3's `parentId` would be rewritten to the new v2 hash, so its historical provenance changes retroactively even though the old v2 record is intentionally retained as a stale/hidden default.

Supplying the exact parent ID makes the relationship mean "this profile was derived from this exact profile version" and keeps lineage stable across future bundled-profile updates. After all defaults are seeded, initialization can validate that the referenced parent exists and apply that fixed `parentId`.

tadelv (commented):
Addressed. The bundled JSON now carries an immutable parent id instead of a filename:

- `best_practice_v3.json` sets `"parent_id": "profile:b73ce54592d667d00b31"` (the content hash of the Adaptive v2 we shipped).
- Seeding resolves the child record by filename but applies the fixed `parent_id` directly, after validating that the parent record exists. If `best_practice.json` content ever changes, the old v2 record (retained as a hidden/stale default by `_retireStaleDefaults`) is still present in storage, so v3's lineage stays pinned to the exact version it was derived from.
- P1 test now asserts the bundled `parent_id` equals the hash computed from the bundled v2 file, so a future v2 content change fails CI instead of silently rewiring lineage.

All tests pass (12/12 in test/profiles/) and `flutter analyze` is clean apart from the pre-existing unrelated asset warning.

PR #723, opened 2026-08-28
fix(attach): USB machine intent wins over preferred BLE during automatic connection (#659)

### LINKED ISSUE #659: Bug: USB machine intent can lose to preferred BLE machine during automatic connection
## Summary

The Android USB-machine override added for #473 can still lose to a stored preferred BLE machine during automatic connection.

A physical USB attachment is explicit user intent: once Decaid receives that attach intent, a passive preferred-BLE scan should not be able to win while the USB path is settling/probing.

## Current behavior

`AttachReconnectCoordinator` waits for the USB settle delay before attempting the probe and re-checks whether a machine is already connected. During that window, the existing automatic connection path can find and connect the saved preferred BLE machine. Once that happens, the pending USB attach attempt is skipped because `ConnectionManager` sees `_machineConnected == true`.

Possible sequence:

```text
USB DE1 attached
  -> attach event received
  -> USB settle timer starts
  -> automatic scan finds stored preferred BLE machine
  -> BLE machine connects
  -> settle timer expires
  -> USB attach attempt is skipped because a machine is already connected
```

There is a related startup gap: if the USB DE1 is already connected before Decaid starts, `SerialServiceAndroid.initialize()` enumerates it but does not emit an attach event. The normal scan/selection path can therefore still prefer a saved BLE machine.

## Expected behavior

When Decaid is not yet connected to a machine and Android reports a USB attachment, that USB intent should be latched immediately and take precedence over passive automatic preferred-machine selection.

The preferred BLE policy should resume only if the attached USB device is unsupported or the serial machine cannot be successfully identified/connected.

A USB DE1 already present at application startup should receive equivalent priority over a saved BLE preference.

This should not change behavior for an already-established machine connection where the cable is attached later; the bug is in automatic machine selection before connection has settled.

## Suggested direction

- Record/latch pending USB machine intent as soon as the attach event is received, before the settle delay.
- Supersede or pause in-flight automatic/adapter-recovery machine selection immediately so a preferred BLE machine cannot win during the USB settle window.
- After the settle delay, probe the attached USB device normally.
- On successful DE1 detection/connection, adopt it and update `preferredMachineId` to the USB ID.
- On unsupported device or probe/connect failure, clear the pending USB intent and resume the existing preferred-machine policy.
- Handle a supported USB DE1 already present at startup before normal automatic preferred-BLE selection.

## Acceptance criteria

- [ ] USB attach while an automatic preferred-BLE scan is in progress cannot result in the BLE machine winning during the USB settle delay.
- [ ] USB attach just before the preferred BLE machine finishes connecting still gives the USB probe/adoption path priority.
- [ ] Successful USB adoption updates `preferredMachineId` only after the serial DE1 is positively identified and connected.
- [ ] Unsupported USB devices leave the existing preference/connection policy unchanged.
- [ ] USB probe/connect failure falls back to the existing preferred-machine recovery policy.
- [ ] Starting Decaid with a supported USB DE1 already attached and a saved BLE preference connects/adopts the USB DE1 first.
- [ ] Existing explicit discovery, scale-only, BLE-only, simulated-device, and non-Android behavior remains unchanged.

## Regression tests

Add coverage for:

1. USB attach during in-flight automatic preferred-BLE scan.
2. USB attach during the race where the BLE connection completes inside the settle window.
3. App startup with USB DE1 already attached plus a saved BLE preference.

## Related

- #473 — original USB-over-stale-BLE preference behavior.
- Archived design: `doc/plans/archive/android-usb-attach-machine-adoption/2026-07-18-design.md`.

tadelv:
## Summary

Fixes the race where a passive automatic preferred-BLE scan can beat an explicit Android USB attachment during the attach settle window (#659). USB intent is now latched the moment the attach event arrives, automatic machine selection is paused while latched, and a machine connected by a superseded automatic attempt is released before the queued probe runs. A USB DE1 already attached at startup now gets the same priority via a startup enumeration hint.

Three changes:

- `AttachReconnectCoordinator` gains `onLatched` (fired when the settle timer arms) and `onSettled` (fired when settle expiry skips the attempt).
- `ConnectionManager` owns the latch: `_maybeScheduleMachineReconnect` defers, `connectMachine` refuses automatic/adapter-recovery connects, `_executeConnect` defers automatic machine work, and a single completion helper clears the latch and resumes whatever it interrupted (replayed automatic connect or re-armed recovery). Explicit direct connects, explicit scans, and scale-only work are never superseded.
- Startup: `DeviceController` subscribes to attach streams before service `initialize()`, and `SerialServiceAndroid.initialize()` subscribes to the platform USB stream before enumeration and emits one generic metadata-free startup hint when devices are found.

## Linked Issue

Fixes #659

## Verification

- New regression tests (all green):
  - Coordinator: `onLatched` fires once per burst, not when a machine is connected; `onSettled` on the skip path.
  - `DeviceController`: attach emitted from inside a service `initialize()` reaches the controller.
  - Serial: one generic hint for a non-empty enumeration, none for empty, platform listener established before enumeration completes, and the generic hint finds the supported machine regardless of DE1/keyboard enumeration order.
  - `ConnectionManager` (probe-capable group): attach during automatic scan supersedes the scan; attach during an in-flight tracked BLE connect releases BLE and adopts USB (including settle expiry inside the transient connected-but-not-yet-released window); attach during remembered quick-connect releases BLE; startup hint before `connect()` prevents quick-connect; unsupported probe clears the latch and resumes the interrupted automatic policy; probe failure after releasing BLE re-arms recovery; supersession marker is consumed by a failed connect so a later connect is not released; a direct explicit connect is not superseded while latched.
- `dart format` clean, `flutter analyze` clean (1 pre-existing pubspec asset warning).
- Full `flutter test`: 0 failures outside the pre-existing `shot_upload` plugin suite (those fail on `main` too — the bundled `assets/plugins/shot-upload.reaplugin/` directory is absent in this checkout).

## Impact

- USB attach behavior on Android: a supported attached DE1 now reliably wins over a saved BLE preference during the settle window and at startup. Unsupported devices and probe failures fall back to the existing preferred-machine policy.
- No REST/WebSocket/DB/skin/plugin API changes. Explicit discovery, scale-only, BLE-only, simulated-device, and non-Android behavior unchanged.
- Docs updated: `doc/DeviceManagement.md`, `doc/AI_BLE_NOTES.md`; design plan archived at `doc/plans/archive/659-usb-latch-vs-preferred-ble/2026-08-28-design.md`.

## Contributor Responsibility

### DIFF
--- a/lib/src/controllers/connection_manager.dart
+++ b/lib/src/controllers/connection_manager.dart
@@ -422,7 +422,16 @@ class ConnectionManager {
     try {
       final probe = _attachProbe;
       if (probe == null) return false;
-      final result = await probe.connectAttachedMachine(event);
+      final AttachProbeResult result;
+      try {
+        result = await probe.connectAttachedMachine(event);
+      } catch (e, st) {
+        // An exceptional probe must not leave the latch set forever: treat
+        // it as unavailable so the fallback path completes the lifecycle
+        // and recovery/replay scheduling stays ungated.
+        _log.warning('Attach probe raised an exception', e, st);
+        return false;
+      }
       switch (result) {
         case AttachProbeConnected(machine: final machine):
           _log.info(
@@ -944,6 +953,18 @@ class ConnectionManager {
           }
         }
         _publishStatus(currentStatus.copyWith(phase: ConnectionPhase.ready));
+        return;
+      }
+      // Quick-connect missed. If USB intent latched while it was pending,
+      // defer the automatic scan to the latch lifecycle instead of
+      // starting a fresh scan that the queued attach probe would wait
+      // behind (the latch's stopScan already ran before this scan existed).
+      if (_usbAttachLatched) {
+        _resumeAutomaticAfterUsbAttach = true;
+        _log.fine(
+          'USB attach latched during quick-connect; deferring automatic '
+          'scan',
+        );
         return;
       }
     }
--- a/lib/src/controllers/connection_manager.dart
+++ b/lib/src/controllers/connection_manager.dart
@@ -338,6 +338,7 @@ class ConnectionManager {
   }
 
   Future<void> _completeUsbAttachLifecycle() async {
+    if (_pendingAttachAttempt) return;
     _usbAttachLatched = false;
     _automaticMachineAttemptSuperseded = false;
     if (_machineConnected) {
--- a/lib/src/controllers/connection_manager.dart
+++ b/lib/src/controllers/connection_manager.dart
@@ -1488,8 +1488,14 @@ class ConnectionManager {
     return connectMachine(resolved);
   }
 
-  Future<ConnectionResult> connectMachine(De1Interface machine) async {
-    if (_usbAttachLatched && _activeAutomaticMachineAttempt) {
+  Future<ConnectionResult> connectMachine(
+    De1Interface machine, {
+    bool automatic = false,
+  }) async {
+    // Only the passive automatic attempt may be superseded by USB attach
+    // intent. Explicit direct connects (REST/WS) are never superseded, even
+    // when they overlap an in-flight automatic scan.
+    if (automatic && _usbAttachLatched && _activeAutomaticMachineAttempt) {
       _log.fine(
         'connectMachine: superseded by USB attach intent, skipping '
         '${machine.deviceId}',
@@ -1524,7 +1530,7 @@ class ConnectionManager {
 
     try {
       await de1Controller.connectToDe1(machine).timeout(_connectTimeout);
-      if (_automaticMachineAttemptSuperseded) {
+      if (automatic && _automaticMachineAttemptSuperseded) {
         // This connect was in flight when USB intent latched; it may still
         // have completed its transport connect, but it must not persist its
         // own preference. The release path disconnects it afterwards. Flush
@@ -1748,7 +1754,7 @@ class ConnectionManager {
     ScanReportBuilder scanReport,
   ) async {
     scanReport.markAttempted(machine.deviceId);
-    final result = await connectMachine(machine);
+    final result = await connectMachine(machine, automatic: true);
     scanReport.recordResult(machine.deviceId, result);
     return _releaseSupersededAutomaticMachine();
   }
--- a/lib/src/controllers/connection_manager.dart
+++ b/lib/src/controllers/connection_manager.dart
@@ -312,8 +312,15 @@ class ConnectionManager {
         _machineRecoveryActive ||
         _machineReconnect != null ||
         _activeAutomaticMachineAttempt;
-    _machineReconnect?.cancel();
-    _machineReconnect = null;
+    if (_machineReconnect != null) {
+      _machineReconnect?.cancel();
+      _machineReconnect = null;
+      if (_machineReconnectFailures > 0) {
+        // A pending timer already counted its backoff tier; no attempt ran,
+        // so undo the tier so the rescheduled attempt keeps the same delay.
+        _machineReconnectFailures--;
+      }
+    }
     if (_activeAutomaticMachineAttempt) {
       _automaticMachineAttemptSuperseded = true;
       _explicitScanGeneration++;
@@ -332,20 +339,22 @@ class ConnectionManager {
       _resumeAutomaticAfterUsbAttach = false;
       return;
     }
-    if (!_resumeAutomaticAfterUsbAttach) return;
-    _resumeAutomaticAfterUsbAttach = false;
     if (_machineRecoveryActive) {
+      // Recovery may have been armed during the probe (e.g. a failed attach
+      // machine) while scheduling was gated by the latch, or suspended at
+      // latch time; either way the timer must be (re)armed now that the
+      // latch gate is gone.
+      _resumeAutomaticAfterUsbAttach = false;
       _maybeScheduleMachineReconnect();
       return;
     }
-    if (settingsController.preferredMachineId != null &&
-        settingsController.preferredMachineId!.isNotEmpty) {
-      _log.info(
-        'USB attach resolved without adoption; resuming automatic '
-        'machine selection',
-      );
-      await _attemptAutomaticConnect();
-    }
+    if (!_resumeAutomaticAfterUsbAttach) return;
+    _resumeAutomaticAfterUsbAttach = false;
+    _log.info(
+      'USB attach resolved without adoption; resuming automatic '
+      'machine selection',
+    );
+    await _attemptAutomaticConnect();
   }
 
   Future<bool> _releaseSupersededAutomaticMachine() async {
@@ -384,10 +393,13 @@ class ConnectionManager {
     );
     if (!handled && settingsController.preferredMachineId != null) {
       _resumeAutomaticAfterUsbAttach = false;
-      await _completeUsbAttachLifecycle();
+      unawaited(_completeUsbAttachLifecycle());
       return _attemptAutomaticConnect();
     }
-    await _completeUsbAttachLifecycle();
+    // Do not await the resume replay here: the coordinator releases its
+    // in-flight guard when this attempt returns, so a second USB attach
+    // during the resumed scan must be able to latch again.
+    unawaited(_completeUsbAttachLifecycle());
     return true;
   }
 
@@ -1260,6 +1272,9 @@ class ConnectionManager {
   @visibleForTesting
   bool get machineRecoveryActive => _machineRecoveryActive;
 
+  @visibleForTesting
+  int get machineReconnectFailures => _machineReconnectFailures;
+
   bool _shouldRetryMachine() {
     return _machineRecoveryActive &&
         !_machineConnected &&
@@ -1509,6 +1524,22 @@ class ConnectionManager {
 
     try {
       await de1Controller.connectToDe1(machine).timeout(_connectTimeout);
+      if (_automaticMachineAttemptSuperseded) {
+        // This connect was in flight when USB intent latched; it may still
+        // have completed its transport connect, but it must not persist its
+        // own preference. The release path disconnects it afterwards. Flush
+        // the registration so the release sees the machine as connected
+        // (the de1 subject delivers asynchronously).
+        _log.fine(
+          'connectMachine: connected but superseded by USB attach intent, '
+          'not persisting preference for ${machine.deviceId}',
+        );
+        await de1Controller.de1.firstWhere(
+          (connected) => connected == machine,
+          orElse: () => machine,
+        );
+        return const ConnectionResult.conflict();
+      }
       await settingsController.setPreferredMachineId(machine.deviceId);
       selectionSession?.scanReport.recordResult(
         machine.deviceId,
--- a/lib/src/controllers/connection_manager.dart
+++ b/lib/src/controllers/connection_manager.dart
@@ -308,7 +308,12 @@ class ConnectionManager {
 
   void _onUsbAttachLatched() {
     _usbAttachLatched = true;
+    // Preserve an already-recorded replay obligation: a second attach may
+    // latch while the first attach's probe is in flight, when the ambient
+    // automatic state is false, and must not clear the interrupted
+    // automatic attempt's right to resume.
     _resumeAutomaticAfterUsbAttach =
+        _resumeAutomaticAfterUsbAttach ||
         _machineRecoveryActive ||
         _machineReconnect != null ||
         _activeAutomaticMachineAttempt;
@@ -438,7 +443,15 @@ class ConnectionManager {
             'Attach probe: machine ${machine.name} (${machine.deviceId}) '
             'connected, adopting',
           );
-          await _adoptAttachedMachine(machine);
+          try {
+            await _adoptAttachedMachine(machine);
+          } catch (e, st) {
+            // A failed adoption (e.g. preference persistence) must not
+            // leave the latch set forever; fall back like an unavailable
+            // probe so the lifecycle completes.
+            _log.warning('Adopting the attached machine failed', e, st);
+            return false;
+          }
           return true;
         case AttachProbeUnsupported():
           _log.fine(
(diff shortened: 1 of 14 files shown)


lib/src/controllers/connection_manager.dart
```diff
@@ -291,10 +304,68 @@ class ConnectionManager {
...
+      _resumeAutomaticAfterUsbAttach = false;
+      return;
+    }
+    if (!_resumeAutomaticAfterUsbAttach) return;
+    _resumeAutomaticAfterUsbAttach = false;
+    if (_machineRecoveryActive) {
+      _maybeScheduleMachineReconnect();
+      return;
+    }
+    if (settingsController.preferredMachineId != null &&
```
ODevStudio: One edge here: resumption is conditioned on `preferredMachineId`, but `_resumeAutomaticAfterUsbAttach` can also be true for a normal automatic scan with **no** preference. In that state, an unrelated/unsupported USB attach stops the scan and we never replay it, so no-preference auto-discovery behavior changes. Could we replay the interrupted automatic attempt regardless of whether a preference is set, and add a no-preference + unsupported-attach test?

lib/src/controllers/connection_manager.dart
```diff
@@ -291,10 +304,68 @@ class ConnectionManager {
...
+        settingsController.preferredMachineId!.isNotEmpty) {
+      _log.info(
+        'USB attach resolved without adoption; resuming automatic '
+        'machine selection',
+      );
+      await _attemptAutomaticConnect();
```
ODevStudio: I think awaiting `_attemptAutomaticConnect()` here keeps the coordinator's `attempt()` in `_inFlight` for the entire fallback scan. Since `_onAttach` drops events while `_inFlight`, a second (real DE1) attach during this resumed scan can be lost and BLE can win again. Could we end the USB lifecycle/coordinator attempt before replaying automatic policy, or have the coordinator retain an attach received while in flight? A two-attach regression test would pin this down.

lib/src/controllers/connection_manager.dart
```diff
@@ -291,10 +304,68 @@ class ConnectionManager {
...
   DeviceAttachedEvent? _pendingAttachEvent;
+  bool _attachProbeInFlight = false;
+
+  void _onUsbAttachLatched() {
+    _usbAttachLatched = true;
+    _resumeAutomaticAfterUsbAttach =
+        _machineRecoveryActive ||
+        _machineReconnect != null ||
+        _activeAutomaticMachineAttempt;
+    _machineReconnect?.cancel();
```
ODevStudio: Small but consequential recovery-state issue: scheduling `_machineReconnect` already increments `_machineReconnectFailures`; cancelling the timer here doesn't undo that. If an unsupported USB attach arrives before the timer fires, rescheduling after the latch clears advances the backoff even though no reconnect attempt happened (e.g. 5s → 10s). Can we preserve the current backoff tier when suspending recovery, or move the failure increment to when the timer actually fires?

lib/src/controllers/connection_manager.dart
```diff
@@ -1373,6 +1474,13 @@ class ConnectionManager {
   }
 
   Future<ConnectionResult> connectMachine(De1Interface machine) async {
+    if (_usbAttachLatched && _activeAutomaticMachineAttempt) {
```
ODevStudio: This entry check handles connects that start after the latch, but not one already awaiting `connectToDe1()` when the latch arrives. That in-flight connect still reaches `setPreferredMachineId()` before `_connectMachineTracked()` releases it, so a superseded BLE attempt can mutate durable preference state. Could we re-check the supersession/generation immediately after the transport connect and before persisting the preference? A no-preference regression test would expose this.

ODevStudio (changes requested):
Thanks for digging into this race — the direction makes sense, and the regression coverage around the original BLE-vs-USB settle-window problem is strong. In particular, latching before settle, handling the late BLE completion, and fixing startup subscription ordering all look like the right pieces.

I found a few state-machine edges that I think we should close before merging. The common theme is that the USB latch now spans coordinator state, automatic-connect state, recovery timers, and persisted preference state, and a few exit paths don't quite preserve the policy that was interrupted:

- an unsupported attach can cancel a no-preference automatic scan without replaying it;
- replaying automatic selection inside the coordinator attempt keeps the coordinator `inFlight`, so a second attach during that fallback can be dropped;
- temporarily cancelling a recovery timer consumes a backoff tier even though no recovery attempt actually ran;
- a BLE connect that was already in flight when USB intent arrived can still persist itself as `preferredMachineId` before we release it.

I've left inline comments on those paths with concrete failure sequences and suggested regression tests.

One adjacent concern (not necessarily introduced by this PR): a direct explicit `connectMachine()` is intentionally allowed while an attach probe is in flight. A successful serial probe has already opened the USB transport before it returns, so an explicit BLE connect can overlap it; depending on completion order, `De1Controller` can replace its current machine reference without disconnecting the other physical transport. The new explicit-connect test only covers an unsupported probe, so it doesn't exercise that successful/successful overlap. I wouldn't necessarily fold a larger serialization cleanup into #659, but I'd either cover the successful overlap here or document it as a follow-up.

Once the four latch/recovery cases below are addressed, I think the core fix will be in much better shape.

lib/src/controllers/connection_manager.dart
```diff
@@ -1373,6 +1489,13 @@ class ConnectionManager {
...
```
ODevStudio: I think the supersession check is still using the *ambient manager state* rather than the ownership of this specific `connectMachine()` call. A concrete case: an automatic scan is active (`_isConnecting == true`, intent `automatic`), REST/WS calls `connectMachine(explicitBle)`, then USB attaches. `_onUsbAttachLatched()` marks `_automaticMachineAttemptSuperseded` because the surrounding scan is automatic, even though the in-flight `connectMachine()` is the explicit REST call. If USB arrives slightly earlier, this entry guard can reject the explicit call outright; if it arrives while `connectToDe1()` is awaited, the post-connect guard returns `conflict` and skips persisting the explicit selection. That contradicts the PR’s “explicit direct connects are never superseded” guarantee. The current direct-connect test starts while `_attachProbeInFlight` is true, so `_activeAutomaticMachineAttempt` is false and it misses this overlap. Could we carry explicit/automatic ownership into `connectMachine` (or separate the tracked automatic path from the public direct path) and add a test: automatic scan active -> direct machine connect starts -> USB latch -> direct connect is not classified as superseded?

ODevStudio (changes requested):
Thanks for closing out the four latch/recovery issues from my previous pass. I re-checked the current head (`817f818`): no-preference replay, coordinator release before replay, recovery backoff preservation, and the superseded preference-write guard all now behave as intended, and the latest PR checks are green. I’ve resolved those old threads.

I found one remaining correctness issue around the PR’s guarantee that explicit direct connects are not superseded. `connectMachine()` is a public/direct entry point, but its USB-supersession checks infer ownership from the manager’s *ambient* automatic state rather than from the caller that initiated this particular connect. If a REST/WebSocket/direct connect overlaps an automatic scan, `_activeAutomaticMachineAttempt` is still true because the scan owns `_isConnecting`/`currentStatus.intent`. A USB latch can therefore either reject that explicit call at entry or set `_automaticMachineAttemptSuperseded` while the explicit transport connect is in flight; the post-connect guard then returns `conflict` and skips persisting the explicit selection even though that connect was not the passive automatic attempt we intended to supersede.

The existing explicit-connect regression does not cover this because it starts the direct connect while the attach probe is already in flight, and `_attachProbeInFlight` deliberately makes `_activeAutomaticMachineAttempt` false. I left an inline note with a concrete overlap sequence and suggested test. Once caller/attempt ownership is explicit here, I think the #659 fix is in good shape. The documented successful-probe/direct-connect transport overlap remains a separate follow-up as agreed.

lib/src/controllers/connection_manager.dart
```diff
@@ -312,8 +392,14 @@ class ConnectionManager {
       attachEvent: event,
     );
     if (!handled && settingsController.preferredMachineId != null) {
+      _resumeAutomaticAfterUsbAttach = false;
+      unawaited(_completeUsbAttachLifecycle());
```
ODevStudio, line 411: There is still an exceptional exit that leaves the USB latch stuck. `_completeUsbAttachLifecycle()` only runs after `_runConnect()` returns normally, but `_executeAttachProbe()` allows `probe.connectAttachedMachine(event)` to throw. `SerialServiceAndroid.connectAttachedMachine()` currently does exactly that if `_listDevices()` throws. The coordinator catches the exception and calls `recover()`, but `_maybeScheduleMachineReconnect()` immediately returns while `_usbAttachLatched` is still true, so automatic recovery can remain permanently paused until another attach happens to clear the latch. Could we guarantee lifecycle completion on the exceptional probe path (e.g. via `finally`/a single completion owner) and add a regression where `connectAttachedMachine()` throws?

lib/src/controllers/connection_manager.dart
```diff
@@ -821,6 +932,7 @@ class ConnectionManager {
       );
       final qcMachine = await _tryQuickConnectMachine();
       if (qcMachine != null) {
+        if (await _releaseSupersededAutomaticMachine()) return;
```
ODevStudio, line 1003: I think the latch needs another check after this await. Sequence: remembered quick-connect starts -> USB attaches/latches -> settle expires while quick-connect is still pending, so the queued attach calls `stopScan()` before any scan exists -> quick-connect later fails/returns `null` -> this method falls through and starts a new automatic BLE scan. There is no later stop signal for that new scan, so the queued USB probe waits behind the full scan (15s in the production BLE service). Could we re-check `_usbAttachLatched` after quick-connect returns null / before `runScan()`, and add a regression that gates quick-connect past settle, returns null, and asserts no scan starts before the queued probe runs?

ODevStudio (changes requested):
Thanks for fixing the explicit-vs-automatic ownership issue from the previous pass. I re-checked the current head (`c3980f1`): the new `automatic` flag correctly keeps direct REST/WS/selection connects out of the USB supersession path, and the current PR checks are green.

I found two remaining lifecycle/timing edges around the latch:

- if the attach probe throws rather than returning an `AttachProbeResult`, the USB lifecycle never completes, leaving `_usbAttachLatched` set; the coordinator's recovery callback then cannot arm machine recovery because reconnect scheduling is gated by that same latch;
- if USB latches while remembered quick-connect is in flight and that quick-connect later returns `null`, `_connectImpl()` falls through into a fresh automatic BLE scan without re-checking the latch. The attach was already settled and its `stopScan()` happened before this scan existed, so the USB probe can be delayed behind the full scan.

I've left inline comments on both paths with concrete failure sequences and suggested regression tests. Once those exceptional/post-await paths also honor the latch lifecycle, I think the #659 fix is in good shape. The previously documented successful-probe/direct-connect transport overlap remains the separate follow-up as agreed.

ODevStudio (changes requested):
Thanks for closing the throwing-probe and post-quick-connect paths in `048cdeb`. I re-reviewed this head. CI and the existing focused tests are green, but two remaining state-machine edges still reproduce:

1. **[P1] Preserve automatic replay across repeated USB latches.** [`_onUsbAttachLatched()`](https://github.com/decentespresso/decaid/blob/048cdeb46be3eb0540379994f9891136f522d85a/lib/src/controllers/connection_manager.dart#L309-L314) replaces `_resumeAutomaticAfterUsbAttach` with the current-state expression. If automatic selection was interrupted, the first queued probe starts, and a second attach latches while that probe is running, `_activeAutomaticMachineAttempt` is false and the second latch clears the already-recorded replay obligation. When both probes are unsupported, automatic selection never resumes. The focused reproduction expects two scans but gets one.

2. **[P2] Complete the USB lifecycle when post-probe adoption throws.** The exception guard in [`_executeAttachProbe()`](https://github.com/decentespresso/decaid/blob/048cdeb46be3eb0540379994f9891136f522d85a/lib/src/controllers/connection_manager.dart#L425-L442) covers only `connectAttachedMachine()`. If `_adoptAttachedMachine()` subsequently throws, for example while persisting `preferredMachineId`, `_attemptAttachReconnect()` never reaches `_completeUsbAttachLifecycle()`. The coordinator catches the exception, but recovery and later automatic connects remain gated by `_usbAttachLatched`. After disconnecting the adopted machine, the focused reproduction expects one scan but gets zero.

Both failures can leave the app disconnected with automatic connection or recovery suppressed, so I am requesting changes.

lib/src/controllers/connection_manager.dart
```diff
@@ -291,10 +304,82 @@ class ConnectionManager {
...
+      deviceScanner.stopScan();
+    }
+  }
+
+  void _onUsbAttachSettled() {
+    unawaited(_completeUsbAttachLifecycle());
+  }
+
+  Future<void> _completeUsbAttachLifecycle() async {
+    _usbAttachLatched = false;
```
ODevStudio, line 350: [P1] Keep the latch until every queued attach is probed

When a second attach is queued during the first probe, this completion clears `_usbAttachLatched` and starts automatic replay before `_runConnect` drains `_pendingAttachAttempt`. That replay can connect the BLE candidate, after which the pending attach is skipped by the `_machineConnected` check. I reproduced this with probe results `[unsupported, connected USB]`: current head invokes only the first probe and persists BLE. Return here while `_pendingAttachAttempt` is set so the queued attach drains before replay.

ODevStudio (changes requested):
Thanks for closing the repeated-latch replay and adoption-failure findings in `fcdb781`. I rechecked this exact head: those fixes work and CI is green. One P1 ordering race remains when another attach is queued while the first probe is running. The focused reproduction fails on this head with one probe instead of two; the local one-line guard and regression test pass the 94-test PR slice, static analysis, and the full suite (3547 passed, 1 skipped).

ODevStudio (approved):
Pushed the queued-attach ordering fix in 7cb7b3ac and rechecked the resulting diff. The regression now runs both USB probes before any automatic BLE replay; the focused PR slice, static analysis, and full suite pass locally. No remaining P0-P2 findings in the PR scope.

PR #726, opened 2026-08-28
feat(identity): resolve legacy DE1 serial/model from linked account

### LINKED ISSUE #674: [Account] resolve legacy DE1 serial/model from support API
Certain early DE1-family machines (notably v1.3 and earlier) can report `0` for `SerialN` and/or `v13Model`. Decaid currently trusts those MMR values directly, which leaves the machine with serial `0` / model `Unknown` and breaks account/model-dependent behavior.

We already have the account plumbing needed for this:

- `DecentAccountService` authenticates against `https://decentespresso.com/support/api/`
- `/support/api/sn?onlyespressomachines=1` is already implemented in `fetchSerialNumbers()`
- **SKU metadata is only returned when the request includes `withskus=1`**, so identity resolution must fetch `/support/api/sn?onlyespressomachines=1&withskus=1`
- with `withskus=1`, the real backend response contains both serial and SKU, e.g.:
  - `1337 DE-BE1BENGLE220V_15A_3000W_B0-01101`
  - `1338 DE-DE1PRO220V7-00533`
- `parseSerialNumbers()` currently discards the SKU metadata
- de1app uses the same support API for the user's registered machines and, when the DE1 does not provide a serial, asks the user to choose/type one of the serials owned by that account

Scope: **legacy DE1-family machines only**. Do not turn this into generic Bengle identity/account resolution.

### 1. Cache registered machine records for the linked account

Replace/extend the current serial-only parsing with a richer account machine record, at minimum:

- serial
- raw SKU
- parsed DE1 model when recognized

Keep `fetchSerialNumbers()` compatibility if useful, but derive it from the richer representation rather than throwing SKU data away.

On successful login:

1. fetch `/support/api/sn?onlyespressomachines=1&withskus=1`
2. parse and persist the last successful registered-machine list
3. keep it available across restarts/offline operation

On app startup with a linked account, load the cached list immediately and refresh it after stored credentials validate. A transient network/server failure must keep the last good cache. Clear account-machine cache and any account-scoped identity mappings on explicit logout/account replacement.

Do not use a definitively rejected account session as fresh authority until the user authenticates again; the persisted cache may remain stored for recovery.

### 2. Normalize DE1 model mapping

Before using API SKU data as authoritative model information, fix/centralize the DE1 model mapping so firmware values and API SKU parsing agree.

The legacy DE1 firmware model values used by de1app are:

- `0` unknown/unset
- `1` DE1
- `2` DE1+
- `3` DE1PRO
- `4` DE1XL
- `5` DE1CAFE
- `6` DE1XXL
- `7` DE1XXXL
- Bengle values remain the existing `>= 128` range and are outside this issue's resolution flow

Current `DecentMachineModel` does not represent all of these and its 5/6 mapping is not aligned with the above. Fix that as part of this work and use one canonical conversion for MMR values and API SKU -> model.

Parse known DE1-family SKU model tokens explicitly. Unknown SKU formats must remain unknown rather than being guessed.

### 3. Resolve effective identity after connection

`UnifiedDe1` should continue to read the raw machine values from MMR. Do **not** write inferred serial/model values back to the machine.

Add an application/controller-layer identity resolution step after the raw machine info has been read. Keep account/UI concerns out of the BLE/transport layer.

Resolution applies only to `DeviceImplementation.unifiedDe1`, not `Bengle`.

When a registered support-API machine record is confidently matched, **the support API is authoritative** for the effective serial/model exposed by Decaid. This includes overriding a nonzero but incorrectly configured machine model: server-side account data is easier to correct than a value burned/configured in machine firmware.

Important: API data only wins **after a specific account machine has been confidently resolved**. Never overwrite a connected machine with an arbitrary account record just because the user owns several machines.

### 4. Matching rules

Use this order:

#### A. Machine reports a nonzero serial

If the serial exactly matches a registered machine record:

- resolve directly to that record
- use the API-derived model when the SKU model is recognized, even when the machine reports a different nonzero `v13Model`
- keep other raw machine information (firmware, GHC, voltage, etc.) unchanged

If the reported serial does not match an account record, keep existing serial-ownership/mismatch behavior. Do not guess another registered machine.

#### B. Machine reports serial `0`

Try, in order:

1. **Existing persisted transport identity mapping**
   - key it by account + `transportType` + Decaid `deviceId`
   - treat `deviceId` as an opaque platform identifier, not specifically a MAC address
   - Decaid already persists remembered-device IDs that can be MAC-like or UUID-like depending on platform
   - only accept the mapping if its serial still exists in the current cached account-machine list

2. **Exactly one registered legacy DE1-family candidate**
   - resolve automatically

3. **Multiple candidates, hardware hints uniquely narrow the set**
   - use a nonzero `v13Model` only as a matching hint
   - optionally inspect legacy `MMRItem.model` (`0x00800004`) as a secondary hint, but do not depend on it
   - current de1app identity code uses `v13Model` (`0x0080000C`); no useful reference implementation reliance on the older `Model` item was found
   - only auto-select when the hints narrow the account list to exactly one candidate
   - if a hint produces no sensible candidate, ignore the hint rather than forcing a bad match

4. **Still ambiguous**
   - show a native Flutter popup/dialog listing the plausible registered machines (serial + friendly parsed model, with SKU available for disambiguation)
   - let the user choose the connected machine
   - persist the `account + transportType + deviceId -> serial` association for future connects

If the persisted BLE/device identifier later changes or cannot be re-used on that platform, simply fall back to the normal automatic/manual matching flow. The feature must not depend on a literal BLE MAC being available.

### 5. Missing account

When a legacy DE1 connects with `serial == 0` and/or `v13Model == 0` and no usable linked account/cache is available, actively prompt the user to link their Decent account so Decaid can identify the machine.

Do not block normal machine operation if the user declines/cancels. Keep the raw machine identity until resolution succeeds.

### 6. Effective machine info

Once resolved, Decaid should expose the resolved values through normal `MachineInfo` consumers:

- resolved serial
- resolved model
- existing firmware/GHC/voltage/refill-kit/etc. unchanged

Prefer keeping raw MMR identity available internally for diagnostics if useful, but public/application behavior should use the resolved effective identity.

Do not write `SerialN`, `v13Model`, or legacy `Model` back to the DE1.

## Suggested implementation shape

Follow `AGENTS.md`: this is a non-trivial multi-subsystem feature, so create a plan under `doc/plans/` and implement test-first.

Likely pieces:

1. Introduce a `RegisteredDecentMachine` (or equivalent) model and parser in the account service area.
2. Extend `DecentAccountService` to fetch/cache registered machine records using `/support/api/sn?onlyespressomachines=1&withskus=1` while retaining serial verification behavior.
3. Add canonical DE1 model/SKU conversion and fix the existing `DecentMachineModel` numeric mapping.
4. Add a small identity resolver service with no UI dependency. Inputs should include raw machine identity, account machine records, transport/device identity, and any persisted association; output should distinguish resolved / ambiguous / unavailable.
5. Integrate resolution at the controller/application layer after `UnifiedDe1.onConnect()` has read raw MMR values and before downstream code relies on effective identity where practical. Do not inject account services into `UniversalBleTransport` or other BLE code.
6. Add the native login prompt / ambiguous-machine selection dialog at the existing UI/state-management layer (there is already navigator access in `De1StateManager`).
7. Persist account-scoped device-id -> serial associations using existing storage conventions; invalidate associations whose serial no longer appears in a refreshed account list.
8. Ensure the existing `_checkSerialOwnership()` flow does not email support because of a temporary raw `0` identity before resolution has had a chance to run.
9. Update `doc/DeviceManagement.md` with the post-connect identity-resolution behavior, per repo guidance.

## Tests / acceptance criteria

At minimum cover:

- registered-machine fetch uses `onlyespressomachines=1&withskus=1`
- parsing real `serial SKU` support API responses without losing SKU metadata
- SKU -> model mapping for supported DE1-family variants
- corrected MMR model mapping (`1..7`) and existing Bengle `>=128` behavior
- login fetches and persists the machine list
- startup uses the last good cache and refreshes it when possible
- transient refresh failure preserves usable cached data
- logout/account replacement clears account-scoped cache/mappings
- raw nonzero serial matching an API record resolves directly
- API model overrides a conflicting nonzero machine `v13Model` after serial match
- serial `0` + one DE1 candidate resolves automatically
- serial `0` + multiple candidates + a unique model hint resolves automatically
- contradictory/unhelpful model hints do not force a match
- serial `0` + ambiguous candidates produces a manual-selection request/dialog
- manual selection persists against `transportType + deviceId` and is reused on reconnect
- persisted mapping is ignored/removed if that serial is no longer registered to the account
- device identifier changing simply falls back to matching rather than breaking connection
- Bengle account records are not selected by the legacy DE1 resolver
- unknown SKU model does not get guessed; serial can still resolve while raw model is retained if needed
- no linked account + missing identity prompts login but does not prevent use when dismissed
- no confident API match leaves raw machine identity untouched
- resolved values are visible through normal `MachineInfo` consumers
- existing serial-ownership mismatch reporting still works for a real nonzero serial that is not on the account

Run the relevant focused tests, `dart format lib test`, `flutter analyze`, and full `flutter test` before considering the work complete.

## References

- Support API: https://decentespresso.com/support/api/
- Current account client: `lib/src/services/account/decent_account_service.dart`
- Current DE1 MMR/model definitions: `lib/src/models/device/impl/de1/de1.models.dart`
- Current raw machine-info read: `lib/src/models/device/impl/de1/unified_de1/unified_de1.dart`
- Existing ownership check/UI access: `lib/src/controllers/de1_state_manager.dart`
- Remembered device IDs / platform-opaque BLE IDs: `lib/src/models/device/remembered_device.dart`
- de1app account serial flow: `decentespresso/de1app`, `de1plus/utils.tcl` + `de1plus/skins/default/de1_skin_settings.tcl`
- de1app machine model read: `decentespresso/de1app`, `de1plus/de1_comms.tcl` (`0x80000C` / `v13Model`)

tadelv:
## Summary

Legacy DE1-family machines (v1.3 and earlier) can report `0` for `SerialN` and/or `v13Model`, leaving the machine with serial `0` / model `Unknown` and breaking account/model-dependent behavior. This PR resolves the effective identity from the linked Decent account's registered-machine list (`/support/api/sn?onlyespressomachines=1&withskus=1`) without ever writing inferred values back to the machine.

What changed:

- **Canonical model mapping** — `DecentMachineModel` now covers firmware values 0..7 (0 Unknown, 1 DE1, 2 DE1+, 3 DE1Pro, 4 DE1XL, 5 DE1Cafe, 6 DE1XXL, 7 DE1XXXL) plus Bengle >= 128, fixing the previous 5/6 misalignment. One conversion is used for both raw MMR values and API SKU parsing; unknown SKU formats stay unknown (never guessed). Rationale recorded in `doc/AI_BLE_NOTES.md`.
- **Registered machine records** — new `RegisteredDecentMachine` (serial, raw SKU, recognized model) and `parseRegisteredMachines`; `parseSerialNumbers`/`fetchSerialNumbers` kept as serial-only compatibility projections. Explicit DE1-family SKU tokens parsed anchored, longest-first; `DE1C...`/`DE1X...` partials stay unknown.
- **Account cache + mappings** — `DecentAccountService` loads the linked account's cached machines and identity mappings from the credential store on startup (before `runApp`), then validates stored credentials and refreshes in the background via one shared memoized refresh (`accountReady`). Transient failures keep the last good cache; a later successful validation after a transient failure refreshes the list. Caches and mappings are bound to the normalized current account; logout/account replacement clears scoped state; a definitively rejected session retains persisted data but never exposes it.
- **Identity resolution** — pure `LegacyDe1IdentityResolver` (resolved / ambiguous / unavailable) with the issue's matching order: exact nonzero serial match (API model overrides a conflicting raw `v13Model`), persisted account+transport+deviceId mapping (deviceId treated as opaque), single legacy DE1 candidate, unique raw-model hint, then a native manual-selection dialog (serial + friendly model + SKU). Bengle and unknown-SKU records are never serial-0 candidates; Bengle is excluded from the legacy resolver entirely.
- **Controller integration** — `De1StateManager` resolves for `DeviceImplementation.unifiedDe1` only, awaiting account readiness, guarding every identity apply with a connection-generation/identity check so a replaced/disconnected machine is never mutated, and running before `_checkSerialOwnership()` so a temporary raw `0` never triggers a support email. Missing/rejected account shows a non-blocking link prompt (copy covers serial+model); cancel leaves the machine fully usable. Manual choices persist and are reused on reconnect; stale mappings are pruned and a changed device ID falls back to normal matching.
- **Effective identity** — `UnifiedDe1.applyEffectiveIdentity()` overrides only serial/model on `MachineInfo` (firmware/GHC/voltage/refill-kit unchanged); raw identity stays available via `rawMachineInfo`/`rawModelValue` for diagnostics. No MMR writes (regression test asserts zero writes).

## Linked Issue

Fixes #674

## Verification

- Focused suites: account service, registered-machine/resolver, `De1StateManager` legacy identity, and effective-identity override — all pass (177 focused + 3,591 full suite).
- Full `flutter test`: 3,591 passed, 1 skipped (pre-existing `REGEN_SIM_ASSETS` regeneration skip).
- `flutter analyze`: no issues. `dart format lib test`: clean (0 files changed).
- All 23 issue acceptance criteria covered by tests, including: `withskus=1` fetch, startup cache + refresh, transient failure preserving cache, logout/account-replacement clearing, exact-match API model override, serial-0 single/multi/hint/ambiguous paths, manual selection persistence, stale-mapping removal, device-ID change fallback, Bengle exclusion, unknown-SKU non-guessing, no-account prompt with dismissal, no-confident-match keeping raw identity, effective values visible through `MachineInfo`, and unchanged real-nonzero mismatch emailing.
- Two independent review loops (spec + standards) were run by a second agent; all findings fixed with regression tests in follow-up commits.

## Impact

- **User-visible**: legacy DE1 machines connected to a linked Decent account now show their registered serial/model instead of `0`/`Unknown`; ambiguous or unregistered serial-0 machines may see a one-time machine-selection or link-account dialog. Non-legacy machines and Bengle are unaffected.
- **Behavior change**: firmware model values 5/6 now map to DE1Cafe/DE1XXL (was DE1XXL/DE1XXXL) and 7 to DE1XXXL (was Unknown) — aligned with de1app and documented in `doc/AI_BLE_NOTES.md` and `doc/DeviceManagement.md`.
- **Storage**: new secure-store keys `registered_machines` and `identity_mappings` (account-scoped JSON), owned by `DecentAccountService`. No DB schema change, no migration.
- **API/spec**: none — no endpoint changes.
- **Docs**: `doc/DeviceManagement.md` documents the post-connect identity flow and exclusions; design rationale archived under `doc/plans/archive/legacy-de1-account-identity/`.

## Contributor Responsibility


lib/src/controllers/de1_state_manager.dart
```diff
@@ -171,23 +181,225 @@ class De1StateManager with WidgetsBindingObserver {
...
     if (machine != null) {
       _logger.info('DE1 connected, starting to listen for state changes');
       _snapshotSubscription = machine.currentSnapshot.listen(_handleSnapshot);
 
-      final sn = machine.machineInfo.serialNumber;
-      if (sn != '0' &&
-          sn.isNotEmpty &&
-          !['mock-bengle', 'mock-de1'].contains(sn)) {
-        unawaited(_checkSerialOwnership(sn));
+      if (machine is UnifiedDe1 && _accountService != null) {
```
tadelv: This type check is not enough to exclude Bengle. `UnifiedDe1.onConnect()` explicitly allows raw model values `>= 128` to continue in degraded DE1-compatible mode, so a Bengle with serial `0` reaches `_resolveLegacyIdentity()`. With one legacy DE1 account candidate, the resolver will auto-select that DE1 and overwrite the Bengle's effective identity; with multiple candidates it can show the legacy selection UI. #674 explicitly scopes this flow away from Bengle. Please guard on `rawModelValue`/`isBengleModelValue()` before entering identity resolution and add a regression test with a real `UnifiedDe1` reporting a Bengle model value (the current “Bengle” controller test only exercises `MockDe1`).

lib/src/services/account/registered_decent_machine.dart
```diff
@@ -0,0 +1,110 @@
...
+    if (_matchesAtTokenBoundary(modelPart, token)) return model;
+  }
+  return null;
+}
+
+/// True when [modelPart] starts with [token] and the token sits at a
+/// boundary. The bare DE1 token must not swallow unknown longer variants
+/// (DE1C..., DE1X...); those stay unknown instead of being guessed.
+bool _matchesAtTokenBoundary(String modelPart, String token) {
+  if (token != 'DE1') return true;
```
tadelv: This makes the parser prefix-based rather than boundary-based for every token except bare `DE1`. For example, an unknown SKU beginning `DE-DE1PRO<letters...>` is accepted as `DE1Pro`, and similarly for `DE1CAFE`, `DE1XL`, etc. That violates this function's contract and #674's requirement that unknown SKU formats remain unknown rather than guessed. Please enforce the token boundary consistently for all recognized tokens (and add a regression test for a longer unknown variant sharing a known prefix).

tadelv (commented):
Blocking findings from the audit (GitHub does not allow the PR author to formally request changes on their own PR):

1. Legacy identity resolution currently also runs for Bengle hardware represented by `UnifiedDe1` in degraded DE1-compatible mode. A Bengle reporting serial `0` can therefore be auto-resolved to the user's sole legacy DE1 account record (or enter the legacy prompt flow), despite #674 explicitly excluding Bengle. Please guard on the raw model (`>= 128`) before this flow and add a regression test using an actual `UnifiedDe1` with a Bengle model value.
2. SKU parsing does not actually enforce a token boundary for any recognized token except bare `DE1`, so unknown variants sharing prefixes can be guessed as a known model. This conflicts with #674's requirement that unknown SKU formats remain unknown.
3. The Contributor Responsibility checkbox in the PR template is still unchecked. `CONTRIBUTING.md`/the PR template treat this as a required gate.
4. `AGENTS.md` says archived plans should retain design rationale, while step-by-step implementation plans should be deleted. The new archived issue plan is primarily milestone/task-list implementation instructions; please remove it or reduce it to durable design rationale/constraints.
5. `AGENTS.md` says not to add explanatory comments to new/substantially rewritten code. This PR adds many explanatory doc comments in the new account/resolver/identity code. Please trim those from code and keep rationale in the domain docs already updated by this PR.

CI is green; these findings are from code/repository-guideline review rather than failing checks.

tadelv (commented):
Blocking follow-up finding: the Bengle exclusion now returns early inside `_resolveLegacyIdentity`, but `_handleDe1Change` routes every `UnifiedDe1` with an account service into that method and therefore never falls through to `_maybeCheckSerialOwnership`. A Bengle surfaced through `UnifiedDe1` (the degraded-compatible path this fix explicitly accounts for) now skips the pre-existing serial ownership/mismatch flow entirely. That is a behavior change despite the PR stating Bengle is unaffected.

Please gate the legacy-resolution branch before routing (e.g. only call `_resolveLegacyIdentity` for non-Bengle raw model values, otherwise preserve `_maybeCheckSerialOwnership(machine.machineInfo.serialNumber)`), and add a regression test asserting a Bengle `UnifiedDe1` with a real nonzero serial still exercises the ownership path while never entering identity resolution/prompts.

The previous Bengle mis-resolution and SKU token-boundary findings are otherwise fixed, the contributor checkbox is now complete, and the archived plan/comments cleanup is substantially addressed. CI is green on `bd203234`.

lib/src/services/account/legacy_de1_identity_resolver.dart
```diff
@@ -0,0 +1,70 @@
...
+    if (rawSerial.isNotEmpty && rawSerial != '0') {
+      final match = registeredMachines
+          .where((m) => m.serial == rawSerial && !m.isBengle)
+          .firstOrNull;
+      if (match != null) return ResolvedLegacyDe1Identity(match);
+      return const UnavailableLegacyDe1Identity();
+    }
+
+    final mappedSerial = mappedMachine?.serial;
+    if (mappedMachine != null &&
```
tadelv, line 42: Blocking: a persisted mapping can still make an unknown-SKU record a serial-0 resolution candidate. `lookupMapping()` returns the current registered-machine record for the mapped serial, and this guard rejects only Bengle. If that record now has `recognizedModel == null` (for example because the backend SKU became unknown), the mapping is accepted and the resolver applies its serial anyway. That conflicts with this PR’s stated contract/design that unknown-SKU records are never serial-0 candidates because they may represent out-of-scope hardware. Please require `mappedMachine.isLegacyDe1Candidate` here (rather than only `!isBengle`) and add a regression test for a persisted mapping whose current record has an unknown SKU.

lib/src/services/account/registered_decent_machine.dart
```diff
@@ -0,0 +1,98 @@
...
+  final modelPart = upper.substring(3);
+  if (modelPart.isEmpty) return null;
+  for (final (token, model) in _skuModelTokens) {
+    if (!modelPart.startsWith(token)) continue;
+    if (_matchesAtTokenBoundary(modelPart, token)) return model;
+  }
+  return null;
+}
+
+/// True when [modelPart] starts with [token] and the token sits at a
```
tadelv: One repo-guideline cleanup remains from the previous review: this is a new explanatory code comment, while `AGENTS.md` says not to add explanatory comments to new/substantially rewritten code and to keep rationale in the matching docs. The helper name plus tests already communicate the behavior, so please remove this comment (or move any needed rationale to the existing identity/BLE docs).

tadelv (commented):
Two remaining findings from this pass. GitHub does not allow the PR author to formally request changes on their own PR, so I’m posting these as review comments. The first is blocking correctness; the second is a repo-guideline cleanup. Current PR checks are green.

lib/src/controllers/de1_state_manager.dart
```diff
@@ -171,23 +182,226 @@ class De1StateManager with WidgetsBindingObserver {
...
+  }
+
+  Future<void> _applyResolvedIdentity(
+    UnifiedDe1 machine,
+    RegisteredDecentMachine record, {
+    required bool persistMapping,
+  }) async {
+    if (!_stillCurrent(machine)) return;
+    final model = record.recognizedModel?.name ?? machine.rawMachineInfo.model;
+    machine.applyEffectiveIdentity(serial: record.serial, model: model);
```
ODevStudio, line 363: I think there’s a cross-subsystem identity issue here. `De1Controller`’s write governor captures `_machineIdentity(device)` when a write is queued; for this legacy case that key starts as `device:<id>` while the serial is `0`, then this call changes it to `serial:<resolved>`. A pending write that starts after resolution can therefore fail `_resolveDeviceFor()` with “Machine changed before device write” even though it is the same object and same connection. This is plausible while `accountReady` is waiting on network and UI/plugins queue multiple writes. Could we keep the governor’s identity stable for the lifetime of the connection (or explicitly re-key pending operations), and add a regression that queues writes before resolution, applies `0 -> resolved serial`, and verifies they still drain?

lib/src/controllers/de1_state_manager.dart
```diff
@@ -171,23 +182,226 @@ class De1StateManager with WidgetsBindingObserver {
...
+        );
+      },
+    );
+    if (accepted != true) return;
+    if (!_stillCurrent(machine)) return;
+    final navigator = _navigatorKey.currentState;
+    if (navigator == null) return;
+    await navigator.pushNamed(AccountPage.routeName);
+    if (!_stillCurrent(machine)) return;
+    await _resolveLegacyIdentity(machine, allowPrompts: false);
```
ODevStudio: There’s an ambiguous-account hole in the post-link flow here. If the newly linked account has multiple plausible legacy DE1s, the resolver returns `AmbiguousLegacyDe1Identity`, but `allowPrompts: false` makes that branch stop without showing the selector. The machine is also still in `_identityPromptedMachines` from the link dialog, so a later prompt-enabled retry cannot show the selection dialog either. The user can end up stuck on raw identity until disconnect/reconnect. Could we clear/separate the link-prompt state and run the normal prompt-enabled resolution after `AccountPage` returns? A regression with “no account -> link account containing two candidates -> select one” would cover it.

lib/src/services/account/decent_account_service.dart
```diff
@@ -45,6 +75,109 @@ class DecentAccountService {
...
 
+  Future<void> initialize() async {
+    _hasLinkedAccount = await hasLinkedAccount();
+    await _loadCachedRegisteredMachines();
+    await _loadCachedMappings();
+    _cacheLoaded = true;
+    unawaited(_ensureMachinesFresh());
+  }
+
+  Future<void> get accountReady => _refreshFuture ?? Future.value();
```
ODevStudio: `accountReady` only waits for a refresh that is already running. If credential validation succeeds but the subsequent `/sn` request fails transiently, `_authenticated` stays `true`, `_machinesRefreshed` stays `false`, and `_refreshFuture` is cleared. From then on `isLoggedIn()` returns the cached `true` before reaching the validation hook that calls `_ensureMachinesFresh()`, while `accountReady` just completes immediately — so the registered-machine list never retries after the network recovers. For a serial-0 machine with no cache, identity can remain unresolved for the rest of the session. Could this path initiate/retry freshness when `_machinesRefreshed` is false (with the existing memoization/backoff semantics), plus a test where `login_test` succeeds, first `/sn` fails, and a later attempt succeeds without re-login?

lib/src/services/account/decent_account_service.dart
```diff
@@ -45,6 +75,109 @@ class DecentAccountService {
...
+      ];
+    } catch (e) {
+      _log.warning('Failed to load cached identity mappings: $e');
+    }
+  }
+
+  Future<void> _persistRegisteredMachines(
+    List<RegisteredDecentMachine> machines,
+  ) async {
+    final email = await _store.read(key: 'email');
```
ODevStudio: I think there’s a small but important account-replacement race here. `refreshRegisteredMachines()` captures and validates the old account/generation, but after those checks this helper re-reads the *current* email. If a successful replacement login lands between the validation and this read/write, the old account’s machine list can be persisted tagged with the new email. If the new account’s own `/sn` refresh then fails, that stale list can be loaded on the next restart as authoritative cache for the new account. The existing in-flight replacement test gates before the generation check, so it doesn’t exercise this post-check window. Could we bind persistence to the captured account/generation and serialize/invalidate account-scoped writes around login/logout, with a gate after validation but before persistence?

ODevStudio (changes requested):
Thanks for the thorough follow-up passes on this one. I reviewed the current head (`9978043`) rather than re-litigating the earlier self-review findings: the Bengle routing/ownership behavior, SKU token boundaries, unknown-SKU persisted mapping case, contributor gate, and docs/comment cleanup all look addressed now. CI is also fully green.

I found four remaining correctness/lifecycle issues that I think are worth closing before merge. They mostly show up where the new effective identity/account lifecycle meets existing subsystems, rather than in the resolver's happy path:

- applying the effective serial changes the identity key used by `De1Controller`'s queued-write governor mid-connection, which can make already-queued writes look like they target a different machine;
- after following the new link-account prompt, an account with multiple plausible legacy DE1s cannot reach the required selection dialog during that connection;
- a transient registered-machine (`/sn`) refresh failure after successful credential validation has no later retry path, so the cache can remain stale/empty for the rest of the session even after the network recovers;
- there is a post-validation account-replacement race where an old account refresh can be persisted under the newly linked account if the account changes between the refresh checks and the cache write.

I've left inline notes with concrete sequences and suggested regression coverage. The implementation is otherwise in good shape and the amount of targeted coverage here is strong; these are the kind of cross-subsystem edges that are easy to miss even with a large focused suite.

lib/src/services/account/decent_account_service.dart
```diff
@@ -53,28 +192,142 @@ class DecentAccountService {
...
+    final account = _normalizeEmail(email);
+    final machines = await fetchRegisteredMachines();
+    if (generation != _authGeneration) return;
+    final currentEmail = await _store.read(key: 'email');
+    if (currentEmail == null || _normalizeEmail(currentEmail) != account) {
+      return;
+    }
+    if (generation != _authGeneration) return;
+    _machines = machines;
+    _cacheLoaded = true;
```
ODevStudio: This closes the mis-attribution part of the race, but I think the stale write can still win *after* replacement. Concrete ordering: old refresh passes both generation/account checks and blocks inside `_persistRegisteredMachines`; `login(new)` increments the generation, clears state, refreshes new account and successfully writes `{account:new, machines:new}`; then the old write resumes and overwrites the same `registered_machines` key with `{account:old, machines:old}`. On restart the blob is ignored for the new account, so the new account has lost its last-good offline cache. The following `_pruneMappings(machines, account)` is also using the shared current `_mappings`, so an old refresh can prune mappings created for the new account in this window. The new gated test currently asserts the stale old blob wins, which codifies the remaining race rather than preventing it. Could we serialize account-scoped writes or add a generation/account check at the actual commit point (including mapping persistence), and change the regression to assert the new account's cache/mappings remain intact after the old gate is released?

ODevStudio (changes requested):
Thanks — I re-reviewed `26ad75a` against the four issues from my previous pass. Three are cleanly addressed now: the write governor keeps a stable per-connection identity, post-link ambiguity can reach the selector, and `accountReady` retries a failed machine-list refresh. The added regressions match those failure modes, and the latest PR checks are green.

There is one remaining race in the account-replacement fix. Binding the persisted blob to the captured account prevents old data from being *mislabelled* as the new account, but it does not prevent an already-started old-account write from landing after the replacement account has written its fresh cache. Because `registered_machines`/`identity_mappings` are single-slot keys, the stale write can overwrite the new account's good persisted cache (and the old refresh's subsequent prune can operate on the new account's in-memory mapping list).

I left one inline comment with the concrete ordering. I think the write itself needs to be invalidated/serialized, not only tagged with the captured account. Once that is closed, the earlier lifecycle findings look resolved from my side.

ODevStudio (changes requested):
Thanks — I re-reviewed the current head (`215fd78`) after the account-write serialization fix. The previous account-replacement race now looks properly closed, and the latest CI run is green.

I found two remaining lifecycle issues:

- **Blocking:** explicit logout can still race with an in-flight registered-machine refresh or mapping save. `logout()` clears the account-scoped state before incrementing `_authGeneration` and removing credentials. An already-started operation can therefore wait behind `_clearAccountScopedState()` on `_accountWriteLock`, then acquire the lock with the still-current generation and persist `registered_machines` / `identity_mappings` again after logout's clear. Logout then completes without deleting those keys a second time. This violates the issue requirement that explicit logout clear the account cache and mappings. I think logout needs to invalidate the generation before waiting on the scoped-state clear (and ideally serialize the credential/state transition under the same lock), with a gated regression where refresh/save overlaps logout and the scoped keys are still absent afterward.

- **Post-link cancel loop:** after the user chooses “Link account”, `_maybePromptLinkAccount()` unconditionally removes the prompt guard and retries identity resolution when `AccountPage` returns. If the user simply presses Cancel on the account page without linking anything, that retry immediately sees the same missing-account condition and shows the “Link your Decent account” dialog again. The issue says declining/cancelling must leave the machine usable, so backing out of the account flow should not immediately re-prompt. Please only retry/reset the prompt state when account state actually became usable, and add a regression for “Link account → Cancel AccountPage → no second prompt”.

The rest of the previously raised resolver, Bengle, governor, refresh-retry, and account-replacement cases look addressed on this head.

ODevStudio (changes requested):
Thanks — I re-reviewed the current head (`0a8fbfe`). The two issues from my previous pass are fixed: logout now invalidates the auth generation before serialized cleanup, and backing out of the account page no longer immediately re-opens the link prompt. The added regressions match those failure modes.

I found three remaining P2 issues:

1. **Account replacement briefly exposes the previous account's in-memory cache.** On a successful replacement login, `login()` writes the new credentials, increments `_authGeneration`, then sets `_authenticated = true` / `_hasLinkedAccount = true` before awaiting `_clearAccountScopedState()`. Until that locked clear runs, `_machines`, `_mappings`, and `_machinesRefreshed` still belong to the previous account. If the old account had previously refreshed successfully and was later marked invalid, `accountReady` can return immediately and `usableRegisteredMachines` becomes visible again while it still contains the old account's machines. A serial-0 machine resolving identity during that window can apply the previous account's serial/model for the rest of the connection. Please hide or invalidate the old scoped state before publishing the new account as usable, with a gated replacement-login regression.

2. **Preconnected machines can permanently miss the required identity prompt.** `De1StateManager._initialize()` subscribes to the replaying DE1 stream before the navigator is necessarily mounted. If a connected legacy machine is replayed then, both prompt paths return when `currentContext` is null. The delayed `_checkNavigationContext()` only updates a readiness flag and does not retry identity resolution; on desktop there may be no later resume event to trigger the retry. Reproduced with a preconnected machine reporting serial/model `0`: mounting the navigator afterward never showed the link prompt. Please retry pending identity resolution when navigation becomes ready and cover this startup ordering.

3. **Identity dialogs survive disconnect.** `_handleDe1Change(null)` clears `_identityPromptedMachines` but does not dismiss an active machine-selection or link-account dialog. The stale dialog remains visible after disconnect, and reconnecting can open another prompt while the old one is still present; selecting from the old dialog then does nothing because its machine is no longer current. Please tie the active identity dialog to its connection and dismiss it on disconnect, with a regression that disconnects while the selection dialog is open.

ODevStudio (approved):
Re-reviewed current head 4ebdc6c2. All previously requested P0-P2 fixes are present, and the final realistic P0-P2 audit is clean. Local verification: 3,670 tests passed (1 skipped), 180 focused tests passed, and flutter analyze --no-pub reported no issues.

PR #735, opened 2026-08-30
fix(connection): early-connect simulated scale in demo mode

### LINKED ISSUE #734: Bug: demo mode scan doesn't stop early after simulated devices connect
Tapping **Try Demo Mode** connects MockDe1 + MockScale almost instantly, but the BLE scan indicator keeps spinning for the full scan duration instead of stopping early.

## Root cause

`enableSimulatedDevicesForSession()` sets session-only preferred IDs (`MockDe1`/MockScale). `SimulatedDeviceService` emits both devices in one stream update. `EarlyConnectWatcher` starts the machine early-connect, then the scale path hits `_shouldDeferEarlyScaleConnect()` before the machine appears on the connected-machine stream. The unresolved-preferred-machine guard (added in `fb949f9b` for the Bengle external-scale race) defers the scale, the watcher treats the deferred call as complete, and `_checkEarlyStop()` never sees both preferred devices connected. The scan runs to full completion before the post-scan scale phase connects MockScale.

## Evidence

A disposable regression probe against current `main` reproduced: both preferred devices in one emission → MockDe1 connected during scan → no scale connect and no `stopScan()` before scan completion → MockScale connected only post-scan.

## Known issue provenance

PR #222 (`feat: Try Demo Mode`) explicitly lists this under **Known issue**: https://github.com/decentespresso/decaid/pull/222

Plan: doc/plans/2026-08-30-demo-mode-scan-early-stop.md

tadelv:
## Summary

Try Demo Mode connects MockDe1 + MockScale almost instantly, but the BLE scan indicator kept spinning for the full scan duration instead of stopping early.

Root cause: `SimulatedDeviceService` emits both devices in one scan update. `EarlyConnectWatcher` starts the machine early-connect, then the scale path hits `_shouldDeferEarlyScaleConnect()` before the preferred machine appears on the connected-machine stream. The unresolved-preferred-machine guard (added in `fb949f9b` for the Bengle external-scale race) defers the scale, so `_checkEarlyStop()` never sees both preferred devices connected and `stopScan()` never fires — MockScale connects only in the post-scan scale phase.

Fix:

- `_shouldDeferEarlyScaleConnect(scale)` now exempts simulated scales (`SimulatedDevice`) from the unresolved-preferred-machine fallback. The Bengle gate still runs first and is unchanged, so the external-scale-before-Bengle protection from `fb949f9b` is preserved for physical scales.
- `_checkEarlyStop()` is guarded by `_earlyStopFired` so two simulated pendings completing back-to-back cannot double-call `stopScan()`.

## Linked Issue

Fixes #734

## Verification

- New regression test `simulated preferred devices emitted together early-connect and stop the scan` — fails on `main` (scale deferred, `stopScan` uncalled), passes with the fix.
- `early-stop` group: 10 pass.
- `test/integration/connection_manager_bengle_scale_test.dart`: 7 pass — external scale visible before preferred Bengle still loses the slot to the virtual scale.
- `flutter analyze`: no issues.
- Full `flutter test`: 3700 pass, 1 pre-existing skip.

## Impact

None. Restores the documented Try Demo Mode behavior (PR #222 listed this as a known issue); no API/spec/documentation changes.

### DIFF
--- a/lib/src/controllers/connection_manager.dart
+++ b/lib/src/controllers/connection_manager.dart
@@ -30,6 +30,7 @@ import 'package:reaprime/src/models/device/device_scanner.dart';
 import 'package:reaprime/src/models/device/transport/data_transport.dart';
 import 'package:reaprime/src/models/device/scale.dart';
 import 'package:reaprime/src/models/device/scan_filter.dart';
+import 'package:reaprime/src/models/device/simulated_device.dart';
 import 'package:reaprime/src/models/device/usb_attach_probe.dart';
 import 'package:reaprime/src/models/scan_report.dart';
 import 'package:reaprime/src/settings/scale_power_mode.dart';
@@ -1187,6 +1188,7 @@ class ConnectionManager {
 
   void _checkEarlyStop(bool earlyStopEnabled) {
     if (!earlyStopEnabled) return;
+    if (_earlyStopFired) return;
     final preferredMachineId = settingsController.preferredMachineId;
     final preferredScaleId = settingsController.preferredScaleId;
     if (preferredMachineId != null && preferredScaleId != null) {
@@ -1880,7 +1882,7 @@ class ConnectionManager {
       );
       return;
     }
-    if (_shouldDeferEarlyScaleConnect()) {
+    if (_shouldDeferEarlyScaleConnect(scale)) {
       _log.fine(
         'Deferring external scale early-connect until machine resolves',
       );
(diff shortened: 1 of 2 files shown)


ODevStudio (approved):
No blocking findings. The fix matches the reported #734 race: EarlyConnectWatcher can launch machine and scale connects from the same device emission, and the new simulated-scale exemption prevents MockScale from being discarded solely because the MockDe1 connection has not propagated to the machine stream yet.

The _earlyStopFired idempotency guard also looks correct. It is reset at the start of every full connection attempt, so it prevents the two concurrent completion callbacks from double-calling stopScan() without suppressing early-stop in subsequent scans.

I also checked the Bengle interaction. The explicit Bengle gate still runs before the SimulatedDevice exemption, and current scale handoff behavior replaces a previously connected scale when the Bengle virtual scale is attached, so I don't see a final-state regression there.

The new regression test exercises the important same-emission case and asserts both the early connections and exactly one stopScan() invocation.

PR #740, opened 2026-08-30
fix: publish connectingMachine during USB attach probe (#738)

### LINKED ISSUE #738: improve direct usb connect flow
When direct connection is available over usb, we're stuck with a blank screen with only 'Retry' button. 

This is a follow-up task for #723

tadelv:
## Summary

Fixes decentespresso/decaid#738: when Decaid detects a direct USB machine and is actively probing/connecting it, the discovery UI now shows the normal machine connection progress state ("Connecting to your machine..." + progress indicator) instead of appearing empty/stuck with a stale `Retry` action.

## Root cause

`ConnectionManager._executeAttachProbe()` only published `ConnectionPhase.connectingMachine` after `UsbAttachProbe.connectAttachedMachine()` returned an `AttachProbeConnected` result. The potentially slow USB open/handshake therefore ran with no matching public phase, and `StatusPublisher` never got a chance to clear an earlier `machineConnectFailed` / `machineDisconnected` state — leaving the stale `Retry` visible.

## Changes

- `lib/src/controllers/connection_manager.dart`:
  - Publish `ConnectionPhase.connectingMachine` (clearing `pendingAmbiguity`, without guessing the transport) as soon as the probe is committed, before awaiting `connectAttachedMachine()`. `StatusPublisher`'s existing non-sticky error clearing now applies, so a stale `Retry` disappears while the connection is legitimately in progress.
  - Guarded `finally` cleanup: if the probe published `connectingMachine` but did not hand off to an adopted machine (unsupported / failed / unavailable / exception / adoption failure), the transient phase is reset to `idle` unless a newer status already replaced it.
- `test/controllers/connection_manager_usb_attach_test.dart`: regression test holding the probe open on the `probeGate` seam, asserting `connectingMachine` with a cleared error mid-probe, then the unchanged successful outcome.
- `test/device_discovery_view_test.dart`: locks the UI contract — `connectingMachine` shows "Connecting to your machine..." and no `Retry`.

USB-vs-BLE priority/latch behavior from #723 is unchanged; no transport, `UsbAttachProbe` API, or recovery policy changes.

## Verification

- `dart format lib test` — clean
- `flutter test test/controllers/connection_manager_usb_attach_test.dart` — 43/43 pass (new regression test fails on unchanged `main` at the in-flight phase assertion)
- `flutter test test/device_discovery_view_test.dart` — 7/7 pass
- `flutter analyze` — 1 pre-existing warning (missing `assets/plugins/shot-upload.reaplugin/` dir, unrelated)
- `flutter test` — 3689 pass; 31 failures all reproduce identically on unchanged `main`: `PathNotFoundException: assets/plugins/shot-upload.reaplugin/manifest.json` (missing asset in this checkout, verified via worktree on `origin/main`)

### DIFF
--- a/lib/src/controllers/connection_manager.dart
+++ b/lib/src/controllers/connection_manager.dart
@@ -435,9 +435,22 @@ class ConnectionManager {
     if (_machineConnected) return true;
     _isConnecting = true;
     _attachProbeInFlight = true;
+    var adopted = false;
     try {
       final probe = _attachProbe;
       if (probe == null) return false;
+      // The probe is committed to connecting the attached machine; publish
+      // the public progress phase before the (potentially slow) USB
+      // open/handshake so the discovery UI shows the connection in
+      // progress instead of a stale idle/error state. The concrete
+      // transport is not known until the probe result is available.
+      _publishStatus(
+        currentStatus.copyWith(
+          phase: ConnectionPhase.connectingMachine,
+          pendingAmbiguity: () => null,
+          activeTargetTransport: () => null,
+        ),
+      );
       final AttachProbeResult result;
       try {
         result = await probe.connectAttachedMachine(event);
@@ -463,6 +476,7 @@ class ConnectionManager {
             _log.warning('Adopting the attached machine failed', e, st);
             return false;
           }
+          adopted = true;
           return true;
         case AttachProbeUnsupported():
           _log.fine(
@@ -503,6 +517,14 @@ class ConnectionManager {
     } finally {
       _attachProbeInFlight = false;
       _isConnecting = false;
+      // If the probe published connectingMachine but did not hand off to an
+      // adopted machine (unsupported/failed/unavailable/exception/adoption
+      // failure), leave that transient phase unless a newer status already
+      // replaced it.
+      if (!adopted &&
+          currentStatus.phase == ConnectionPhase.connectingMachine) {
+        _publishStatus(currentStatus.copyWith(phase: ConnectionPhase.idle));
+      }
     }
   }
 
--- a/lib/src/controllers/connection_manager.dart
+++ b/lib/src/controllers/connection_manager.dart
@@ -436,6 +436,10 @@ class ConnectionManager {
     _isConnecting = true;
     _attachProbeInFlight = true;
     var adopted = false;
+    // Preserve the interrupted picker so a probe that does not adopt a
+    // machine can restore it instead of dropping the user into the generic
+    // results view.
+    final interruptedAmbiguity = currentStatus.pendingAmbiguity;
     try {
       final probe = _attachProbe;
       if (probe == null) return false;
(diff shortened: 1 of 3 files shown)


ODevStudio (changes requested):
The core fix is sound, and CI is green, but I found two state-machine edge cases.

[P2] Restore pendingAmbiguity when the attach probe does not adopt a machine. The new pre-probe transition unconditionally clears pendingAmbiguity. If a USB attach occurs while a machine/scale picker is active and the probe returns Unsupported, Unavailable, fails, or throws, finally only restores phase: idle; it does not restore the ambiguity. This matters because the selection session remains active, while selectMachine() explicitly rejects selections unless pendingAmbiguity == machinePicker. ScanFlowView also switches away from its actionable single-device picker when that flag disappears, leaving the user with the generic results view and requiring a rescan. The earlier USB-latch design explicitly says unsupported probes should resume interrupted work and that explicit flows are not superseded. I’d save the previous ambiguity before publishing the transient progress state and restore it on non-adoption, unless another status genuinely superseded the probe.
[P2] Don’t publish idle if adoption failed after the machine actually connected. _adoptAttachedMachine() calls de1Controller.adoptDevice() and waits for the machine stream before persisting preferredMachineId and doing subsequent setup. Any failure in those later operations is caught as an “adoption failure,” leaving adopted == false; the new finally then changes connectingMachine to idle. At that point _machineConnected can already be true—the disconnect supervisor updates that state when the DE1 stream emits, but its connected callback does not publish ready. The onboarding/discovery flow only advances on ConnectionPhase.ready, so this can leave a genuinely connected machine stuck on discovery. The cleanup should use the actual connection state, e.g. settle to ready when _machineConnected is already true, rather than keying solely on the local adopted boolean.

The main-path regression test is good: it verifies connectingMachine is observable while connectAttachedMachine() is still blocked and that the stale non-sticky error is cleared. I’d add tests for picker → unsupported probe → picker restored and machine adopted → preference persistence throws → connected state does not settle to idle before merging.

ODevStudio (approved):
Re-reviewed the current head. The two previously blocking state-machine issues are addressed and covered by regression tests, and CI is green.

[P3] There is one remaining non-blocking concurrency edge case: attach-probe cleanup identifies ownership only by `currentStatus.phase == ConnectionPhase.connectingMachine`. Because an explicit `connectMachine()` is allowed to overlap an in-flight attach probe and publishes the same phase, an unsupported/unavailable probe can finish first and reset the public status to `idle` while the explicit machine connection is still in progress. That can cause a temporary discovery/onboarding UI mismatch until the explicit connect eventually publishes its next state.

This requires a fairly narrow ordering—USB probe in flight, a separate explicit REST/WS/direct machine connect starting during it, and the probe finishing before that explicit connect—so I consider it P3/non-blocking for #738. A follow-up hardening change could exclude an active `_isConnectingMachine` from the probe cleanup or use operation/generation ownership for the transient status.

Approved.

PR #742, opened 2026-08-31
feat(scale): show connected Skale firmware revision

### LINKED ISSUE #714: feat(scale): show the connected Skale firmware revision
#### Problem

The Skale exposes its firmware revision, but Decaid does not display or publish
it. This makes support and compatibility reports harder to diagnose.

#### Proposed scope

- Read the standard Device Information Firmware Revision String characteristic
  (`0x180A` / `0x2A26`) after connecting.
- Treat the value as an opaque device-provided string such as `R029`.
- Show it in the Devices UI and expose it in nested REST/WebSocket `deviceInfo`.
- Clear it on disconnect and do not persist it as remembered-device truth.

#### Acceptance criteria

- A connected Skale's firmware string appears in the UI and device inventory.
- Missing or malformed firmware data does not fail the connection.
- No firmware update or DFU behavior is introduced.

#### Firmware documentation limit

Atomax publishes SDK versions but no public scale-firmware catalog, firmware
changelog, image archive, or compatibility matrix. The official Android app
compares the hexadecimal suffix of `R0xx` with a private cloud manifest and can
perform Nordic DFU, but that is insufficient to implement a safe third-party
updater. Decaid should display the reported revision only.

#### Evidence

- [Bluetooth SIG Device Information Service firmware field](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/DIS_v1.2/out/en/index-en.html)
- [Bluetooth SIG assigned number for `0x2A26`](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Assigned_Numbers/out/en/index-en.html)
- [SkaleKit iOS client-SDK changelog](https://github.com/atomaxinc/SkaleKit/blob/624c07436a577706dcc26cb2d62ca76ad18f7da9/README.md#L157-L165)
- [SkaleKit Flutter client-SDK changelog](https://github.com/atomaxinc/SkaleKit-flutter/blob/7df4a9d9bd975367c3ac27502323225ac61b116d/CHANGELOG.md#L8-L54)
- [SkaleKit iOS releases](https://github.com/atomaxinc/SkaleKit/releases)

The official Android app is proprietary, so its updater code cannot be given a
public source deeplink. The reproducibility anchor for the local static analysis
is app version 2.3.0 (version code 13), base APK SHA-256
`b28c9e1f118a73a616007197a99127045316b501841028edb37d0d0ef59e7c2d`.
It reads `0x2A26`; its updater fetches `fw.plist`, compares `version_number`
against the hexadecimal suffix of `R0xx`, downloads `nrf51822.zip`, scans for
`DfuTarg`, and passes the image to Nordic's Android DFU library. This evidence
supports documenting the behavior, but not implementing an updater without a
public firmware contract.

MaxRink:
## Summary

Read the connected Skale firmware revision from the standard Device Information Service and show the opaque value in Devices and `GET /api/v1/scale/info`. Metadata belongs to the current connection and clears on disconnect.

## Linked Issue

Fixes #714

## Verification

- Full Flutter 3.44.2 Linux suite: 4,170 passed, 1 skipped, 0 failed.
- `dart format lib test`: no changes; `flutter analyze`: no issues.
- Focused fake-transport and handler tests cover metadata clearing, valid/unknown values, and REST/WS inventory exclusion.
- Simulated integration of the native Skale series exercised the metadata endpoint through `sb-dev`, including hot reload. The companion USB branch also verified settings persistence across cold restart.
- Physical Skale hardware was not retested for this revision.

## Impact

Firmware metadata uses the narrow `DeviceInformationCapable` capability; other scales do not need no-op metadata getters. Device inventory REST/WS contracts remain unchanged. No firmware update or DFU behavior is added. API and device documentation are updated.

## Contributor Responsibility


tadelv (changes requested):
I think the API boundary here should be changed before merge.

`GET /api/v1/devices` is an inventory/discovery surface: it can contain connected devices, currently available devices, and remembered devices that are no longer present. Adding connection-scoped `deviceInfo` to those inventory entries makes the resource ambiguous, because the enclosing device can persist while the metadata only exists for an active connection.

We already have a clearer precedent with `GET /api/v1/machine/info`, which represents information about the currently connected machine. I would prefer the scale side to follow the same model:

- do not add `deviceInfo` to `/api/v1/devices` or `/ws/v1/devices`;
- add `GET /api/v1/scale/info` for information obtained from the currently connected scale;
- keep `firmwareVersion` opaque and optional;
- let #743 extend the same `ScaleInfo` contract with nullable/optional `batteryLevel`;
- if live scale-info updates are eventually needed, prefer a dedicated `/ws/v1/scale/info` surface rather than causing the device-inventory stream to change for metadata refreshes. I would not add that WebSocket until there is a concrete need.

This keeps a useful API distinction going forward: `/api/v1/devices` answers what devices exist and their connection state; role-specific `/machine/info` and `/scale/info` endpoints answer what Decaid has learned from the currently connected device. That also keeps this generic across future scale implementations without coupling the public API to Skale.

doc/DeviceManagement.md
```diff
@@ -809,6 +810,20 @@ of vanishing. Cross-transport (BLE/USB/WiFi) by construction.
   path). Moving a USB device to a different physical port yields a new id (new
   remembered entry); Forget removes the stale one.
 
+### Connected-session device information
+
+Devices may implement the optional `DeviceInformationCapable` interface for
+connected-session metadata. The devices REST and WebSocket inventories follow
+that stream and include a nested `deviceInfo` object when data is available.
```
tadelv: This still documents the pre-review API contract. The revised implementation/tests and `AI_API_NOTES` intentionally keep `/api/v1/devices` and `/ws/v1/devices` inventory-only, but this paragraph says those inventories follow `deviceInformation` and include nested `deviceInfo`. Please update it so only the Devices UI follows the connection-scoped metadata stream, while REST/WS clients use `/api/v1/scale/info`. Leaving the architecture guide contradictory will mislead future changes.

lib/src/models/device/scale.dart
```diff
@@ -14,6 +15,18 @@ abstract class Scale extends Device {
   Future<void> resetTimer() async {}
 }
 
+class ScaleInfo {
+  final String? firmwareVersion;
+  final int? batteryLevel;
```
tadelv: Please keep `batteryLevel` out of #742. The requested split was for this PR to introduce `ScaleInfo` with optional opaque `firmwareVersion`, then let #743 extend the same contract with `batteryLevel`. Adding it here also puts the Dart `toJson()` model ahead of this PR's OpenAPI schema. Remove the field/serialization here and let #743 add it together with its API/spec behavior.

tadelv (changes requested):
The revised API boundary is now aligned with the requested design: connected scale metadata lives under `/api/v1/scale/info`, while device inventory remains inventory-only. I still see two code/docs issues called out inline, plus the current head fails the repository's `Dart format check`, which is a hard pre-PR/verification gate in `AGENTS.md` and prevents analyze/tests from running in CI. Please run `dart format lib test`, push the formatter output, and address the inline items.

doc/DeviceManagement.md
MaxRink: Corrected in fdfc8bce. Only the native Devices page follows the connection-scoped metadata stream; REST clients use `/api/v1/scale/info`, and both device inventory surfaces explicitly exclude `deviceInfo` and metadata-refresh emissions. Corrected the same stale wording on the related branches as well.

lib/src/models/device/scale.dart
MaxRink: Removed `batteryLevel` from the ScaleInfo field, constructor, and serialization in fdfc8bce. #742 now introduces only optional opaque firmwareVersion; #743 retains the battery extension. Ran `dart format lib test`, the formatter check, and `flutter analyze` on all five Decaid PR branches; all passed. The 34 targeted ScaleInfo/Skale tests passed. The full #742 suite has one remaining unrelated LAN-hostname WebUI redirect failure (expected 307, got 500). New CI checks are not yet listed on this head.

lib/src/services/webserver/devices_handler.dart
```diff
@@ -10,7 +10,7 @@ class DevicesStateAggregator {
 
   final List<StreamSubscription> _subscriptions = [];
 
-  final Map<String, (Device, StreamSubscription)> _deviceStateSubs = {};
+  final Map<String, (Device, List<StreamSubscription>)> _deviceStateSubs = {};
```
tadelv: This `List<StreamSubscription>` shape is leftover scaffolding from when the PR was going to subscribe device inventory to metadata. The corrected design no longer does that, so every device still has exactly one connection-state subscription. Please revert this map and the associated loops to the original `(Device, StreamSubscription)` bookkeeping rather than carrying an unrelated complexity increase into #742.

tadelv (changes requested):
The two previous blockers are resolved. One cleanup remains from the abandoned inventory-metadata implementation: please revert the unused multi-subscription bookkeeping in `DevicesStateAggregator` back to the original single connection-state subscription shape. With inventory no longer subscribing to metadata, this refactor has no purpose and conflicts with the repository guidance to avoid adjacent refactors that the task does not require.

lib/src/services/webserver/devices_handler.dart
MaxRink: Restored the original `(Device, StreamSubscription)` bookkeeping in f860b0e2, including the single cancellation calls. `devices_handler.dart` now has no diff against the base implementation. Applied the same cleanup to #743 and #745.

tadelv (changes requested):
The API boundary and the previously requested cleanup are now in the right shape. I still want two things corrected before merge:

1. Please remove the broad no-op `ScaleInfo? get scaleInfo => null` overrides from every non-Skale scale implementation/test double. `Scale` already provides that default, so these edits add churn across unrelated device implementations without changing behavior. This is exactly the kind of adjacent change AGENTS.md asks us to avoid; only implementations that actually provide scale info should override it.
2. Please drop the remaining unrelated diff while doing that (`tare`/timer warning additions in `scale_handler.dart` and the newline-only `websocket_v1.yml` change). They are not needed for #714.

Also, please verify the result against current `main` before considering this ready. #820 has now merged the host-owned BLE/plugin-driver checkpoint, while this branch's local/full-suite evidence predates that merge and the current PR workflow is still `action_required` with no jobs. A rebase plus the relevant Scale/PluginScale/API tests (or a green merge-result CI run) is enough; no extra hardware testing is required for this review.

I do not think #742 itself needs to be held for the generic-driver work: `/api/v1/scale/info` is a generic host/domain surface and the Skale firmware read is narrowly scoped. But please keep the new metadata abstraction minimal here; richer plugin/device capabilities should continue to be defined through #809 rather than expanded from this Skale-specific PR.

tadelv (approved):

PR #743, opened 2026-08-31
fix(scale): make Skale battery reporting nullable and refresh it

### LINKED ISSUE #713: fix(scale): make Skale battery reporting nullable and refresh it
#### Problem

Skale battery data is currently initialized as `0%` and read only once. A failed
or unsupported read is therefore indistinguishable from an empty battery, and a
long-running connection never refreshes its value. Some Atomax firmware also
reports a fixed `100%`, so the value must remain explicitly device-reported.

Battery reporting does work on the observed R029 scale: the official Skale app
reported 82%, and Decaid reported 80% after reconnecting. The bug is therefore
not “R029 has no usable battery data”; it is Decaid's unavailable/error semantics
and lack of refresh during a long-lived connection. The historical fixed-`100%`
behavior should remain documented as firmware-generation-specific rather than
generalized to current units.

#### Proposed fix

- Represent unavailable battery data as `null`, not `0%`.
- Validate the Bluetooth Battery Level value as `0..100`.
- Read on connect and refresh every 30 minutes without overlapping reads.
- Clear connected-session battery metadata on disconnect.
- Treat read failures as best-effort metadata failures, not connection failures.

#### Acceptance criteria

- Failed, empty, and out-of-range reads do not publish a percentage.
- Valid reads appear in the Devices UI and nested REST/WebSocket `deviceInfo`.
- Battery metadata refreshes while connected and disappears on disconnect.

#### Evidence

- [iOS battery callback and read API](https://github.com/atomaxinc/SkaleKit/blob/624c07436a577706dcc26cb2d62ca76ad18f7da9/Distribution/SkaleKit.xcframework/ios-arm64/SkaleKit.framework/Headers/SKSkale.h#L26-L28)
- [iOS battery read method](https://github.com/atomaxinc/SkaleKit/blob/624c07436a577706dcc26cb2d62ca76ad18f7da9/Distribution/SkaleKit.xcframework/ios-arm64/SkaleKit.framework/Headers/SKSkale.h#L81-L85)
- [Android battery callback documentation](https://github.com/atomaxinc/SkaleKitAndroid/blob/faae8a96879f45c935712c11f25674e4d5dc9e82/README.md#L52-L62)
- [Android battery request documentation](https://github.com/atomaxinc/SkaleKitAndroid/blob/faae8a96879f45c935712c11f25674e4d5dc9e82/README.md#L99-L109)
- [de1app's historical fixed-100% firmware note](https://github.com/decentespresso/de1app/blob/24851f5b1b0faf77b92fdd837b2932a8d8e26e15/de1plus/bluetooth.tcl#L2645-L2652)

MaxRink:
## Summary

Report unavailable Skale battery metadata as unknown rather than a measured 0%. Accept one-byte values from 0 through 100, refresh every 30 minutes, prevent overlapping reads, and discard results from retired connections.

## Linked Issue

Fixes #713

## Verification

- Full Flutter 3.44.2 Linux suite: 4,177 passed, 1 skipped, 0 failed.
- `dart format lib test`: no changes; `flutter analyze`: no issues.
- Focused fake-transport and handler tests cover metadata clearing, valid/unknown values, and REST/WS inventory exclusion.
- Simulated integration of the native Skale series exercised the metadata endpoint through `sb-dev`, including hot reload. The companion USB branch also verified settings persistence across cold restart.
- Physical Skale hardware was not retested for this revision.

## Impact

Connected firmware/battery metadata uses `DeviceInformationCapable` and the existing `/api/v1/scale/info` projection. Device inventory REST/WS contracts remain unchanged. Native Skale weight snapshots retain their existing fallback behavior; low-battery warning policy is outside this PR. This branch includes the firmware foundation from #742. API and device documentation are updated.

## Contributor Responsibility


assets/api/rest_v1.yml
```diff
@@ -5442,6 +5467,22 @@ components:
           additionalProperties: true
       additionalProperties: true
 
+    DeviceInformation:
```
tadelv: Please remove this `DeviceInformation` schema (and the corresponding WebSocket schema). After the API-boundary correction it is not referenced by any endpoint or inventory payload; leaving an orphan public schema named for connection metadata makes the superseded `deviceInfo` contract look supported. `ScaleInfo` is the public connected-scale metadata contract now.

tadelv (changes requested):
The corrected `/api/v1/scale/info` boundary looks good. One stale artifact remains from the old nested-inventory design: both API specs still define a public `DeviceInformation` component even though no REST or WebSocket surface references or emits it. Please remove those unused schemas and keep the public contract centered on `ScaleInfo`. Propagate the same cleanup to #745, which currently extends that stale schema further.

assets/api/rest_v1.yml
MaxRink: Removed both unused DeviceInformation components in 53686c1e. The REST and WebSocket specs no longer declare or reference that superseded public contract; ScaleInfo remains unchanged. Propagated the cleanup to #745 in 4c103e06, including its orphan power-source properties. Internal native-UI metadata and the manual per-device USB settings contract are unchanged.

tadelv (approved):

PR #744, opened 2026-08-31
feat(scale): add opt-in Skale physical-button actions

### LINKED ISSUE #715: feat(scale): add opt-in Skale physical-button actions
#### Problem

Atomax exposes the Skale's circle and square button notifications, but Decaid
currently ignores them.

#### Proposed scope

- Map button value `1` to circle and value `2` to square.
- Make circle request the existing scale tare action.
- Add a default-off setting that lets square start espresso when idle and stop
  it while brewing.
- Keep these as application actions, not a new public button-event API.

#### Acceptance criteria

- Unknown button values are ignored safely.
- Circle uses the existing tare path.
- Square does nothing by default and toggles espresso only after opt-in.
- The setting persists and survives settings export/import.

#### Evidence

- [iOS button callback API](https://github.com/atomaxinc/SkaleKit/blob/624c07436a577706dcc26cb2d62ca76ad18f7da9/Distribution/SkaleKit.xcframework/ios-arm64/SkaleKit.framework/Headers/SKSkale.h#L43-L53)
- [Android button IDs 1 and 2](https://github.com/atomaxinc/SkaleKitAndroid/blob/faae8a96879f45c935712c11f25674e4d5dc9e82/README.md#L52-L62)
- [Android example's circle-button tare](https://github.com/atomaxinc/SkaleKitAndroid/blob/faae8a96879f45c935712c11f25674e4d5dc9e82/example/src/main/java/com/atomaxinc/skalekit/example/MainActivity.kt#L159-L165)
- [de1app's circle/square action mapping](https://github.com/decentespresso/de1app/blob/24851f5b1b0faf77b92fdd837b2932a8d8e26e15/de1plus/bluetooth.tcl#L3100-L3118)

MaxRink:
Superseded by draft #846, which moves the Skale button policy into JavaScript as discussed. The native branch is retained for reference. The replacement uses the guarded API proposed in #845/#853 and does not implement the rejected wake path.

## Summary

What changed, and why?

- Handle Skale button notifications and map protocol value `1` to circle and `2` to square; ignore empty and unknown values.
- Route circle presses through the existing scale tare path.
- Add a persisted, exported, REST-configurable, default-off square-button action setting keyed by exact device ID.
- Map square presses from idle to espresso start on machines without an active group-head controller, wake eligible sleeping machines before starting, and stop active espresso on every machine when the currently connected Skale has opted in.
- Send button-triggered state requests to the captured DE1 connection so an unfinished initialization write cannot block the physical control or redirect a follow-up command to a replacement machine.
- Do not send remote start requests to active-GHC machines, where DE1 firmware requires the operation to start at the group head for UL compliance.
- Keep button handling as application actions instead of adding a public button-event API.

## Linked Issue

Fixes #715

## Verification

- [x] Targeted backend/API correction suites: 63 tests passed.
- [x] `flutter analyze` completed with no issues.
- [x] Actual `ScaleHandler.addRoutes()` coverage verifies `GET /api/v1/scale/info`, disconnected `503`, connected metadata serialization, REST and WebSocket inventory metadata exclusion, and no inventory emission for metadata refresh.
- [ ] Hardware/end-to-end smoke testing was not run. The macOS environment lacks `xcodebuild`; Android requires a missing `debug.keystore`. An existing port-8080 ADB issue is unrelated and unchanged.

## Impact

- Circle presses invoke the existing tare behavior.
- Square start/stop actions require explicit opt-in for the connected Skale. They start from idle, safely wake and start from sleeping only when no active GHC is present, stop active espresso on all machines, and ignore other states; gateway `full` is excluded.
- Active-GHC machines require dangerous operations to start at the group head for UL compliance. Decaid no longer arms the white ready indication with a square-button start request.
- `scaleButtonStartsEspressoByDevice` participates in persistence, export/import, and the REST settings contract. Only enabled device IDs are stored.
- Two Skales can retain different square-button behavior without changing each other.
- Queued notifications from a disconnected scale cannot trigger an action because routing uses the current connection rather than historical device identity.
- Physical-button actions remain active while Decaid's service is connected, including when the UI is backgrounded. This is called out for explicit maintainer review.
- No public button-event API or WebSocket setting event is added.

## Contributor Responsibility

## API boundary

Connected-scale metadata is exposed through `GET /api/v1/scale/info` as `ScaleInfo`. The inventory REST endpoint `/api/v1/devices` and the `/ws/v1/devices` stream do not expose `deviceInfo` or emit metadata-only inventory refreshes; device-specific settings and native metadata UI remain supported separately.


tadelv (changes requested):
Please keep #715 to the accepted behavior: square-button start from `idle`, stop while `espresso`, default off. The new `sleeping -> idle -> espresso` path is a material remote wake-and-brew expansion that is not in the issue acceptance criteria and carries a different safety/intent profile. Remove the sleeping start path from this PR; it can be proposed separately if we want that behavior. The GHC gating and per-device opt-in otherwise look directionally correct.

tadelv (commented):
I think we should put this PR on hold rather than continue iterating on the native implementation for now.

The main reason is #809 and the work already underway in #813. #813 is explicitly building the generic boundary we want: Decaid owns BLE discovery/connection/session lifecycle and safe GATT access, while JavaScript plugin drivers own device protocols and device-specific behavior. Once that work is complete, the Skale button path should be expressible without adding Skale-specific machine-action plumbing to core:

`BLE notification -> Skale plugin -> Decaid HTTP API -> machine action`

The plugin can decode the Skale button notification through the host-owned BLE/GATT subscription and then request the existing Decaid machine API (for example, an espresso state request). That keeps button semantics, opt-in policy, and other Skale-specific behavior in the driver/plugin instead of spreading it through `ScaleController`, `De1StateManager`, settings, and native device abstractions.

Given the relatively narrow user impact of this feature and the amount of native cross-cutting machinery it currently introduces, I don't think it makes sense to invest further in this approach before #809 is complete. Please hold #744 for now; we can revisit #715 afterward and I expect much of it can be implemented as a small Skale plugin feature instead.

There may still be a useful generic idea behind some of this work: making the device/peripheral API more capability-driven so scale/grinder/sensor drivers can declare richer capabilities and possibly driver-specific data/API surfaces. I would like to take that question to the generic-driver discussion rather than establish those abstractions in this Skale-specific PR.

PR #746, opened 2026-08-31
fix(connection): stop adoptDevice() emitting an interim null on re-adopt

allofmeng:
## Summary

What changed, and why?

- Fixes a race where the skin shows "Machine disconnected unexpectedly"
  while a USB-connected DE1 is still attached and streaming data.
  `De1Controller.adoptDevice()` unconditionally resets (emits `null` on
  the `de1` stream) before emitting the newly adopted device, even when
  re-adopting an already-connected machine (e.g. a fresh transport
  object for the same physical USB device after re-enumeration).
  `DisconnectSupervisor` only suppresses that internal churn while
  `isConnectingMachine()` is true, but that guard was only ever set by
  the manual `connectMachine()` path — the automatic quick-connect
  (`_tryQuickConnectMachine`) and USB attach-probe
  (`_adoptAttachedMachine`) paths called `adoptDevice()` without holding
  it, so their reset-then-adopt churn could be misread as an unexpected
  disconnect.
- Fixed at the source rather than widening the guard:
  `De1Controller.adoptDevice()` no longer emits an interim `null` at all
  when replacing an already-connected device. The internal reset
  (generation bump, `dataInitialized` reset, subscription teardown) is
  factored into `_resetForNewDevice()`, which `adoptDevice()` now uses
  directly, going straight from the old device to the new one in a
  single emission. There is no churn left for `DisconnectSupervisor` to
  misread. `connectToDe1()` (the `connectMachine()`-only path) is
  unchanged; it was never implicated.
- Removing the interim null surfaced a second, independent, pre-existing
  race: `_releaseSupersededAutomaticMachine()` reads `_machineConnected`
  (backed by `DisconnectSupervisor`'s async-updated `_latestDe1`)
  immediately after `adoptDevice()` returns. The old code's
  null-then-device pair was accidentally providing microtask headroom
  for the supervisor to catch up; without it a usb-attach test failed.
  Fixed with `DisconnectSupervisor.waitForMachine(deviceId)`, a `Future`
  resolved from inside the supervisor's own pre-existing listener
  rather than a fresh subscription (a new `de1.firstWhere(...)` listener
  would race cross-listener delivery ordering on the async broadcast
  stream), so completion is ordered correctly by construction.
  `_tryQuickConnectMachine` awaits it before returning.

## Linked Issue

N/A — reported by a user via chat log analysis, not a tracked issue.

## Verification

- Diagnosed against a user-provided app log
  (`de1crops/log.txt`, 2026-08-31): `Quick-connect: machine adopted` →
  `machine connected, proceeding to ready` → SEVERE
  `machineDisconnected` all within ~1ms, followed by the machine
  reconnecting on its own a couple ms later — confirming the DE1 never
  actually dropped.
- Added a regression test in `test/controllers/de1_controller_test.dart`
  (`adoptDevice` group) that drives the real `De1Controller.adoptDevice()`
  with two real `TestDe1` instances over the real `BehaviorSubject`-backed
  `de1` stream, asserting the emission sequence has no interim null when
  replacing an already-adopted device. Confirmed it fails (hangs/times
  out) against the pre-fix `adoptDevice()` and passes against the fix.
- Added focused unit tests for `DisconnectSupervisor.waitForMachine()` in
  `test/controllers/connection/disconnect_supervisor_test.dart`:
  resolves immediately when already current, resolves when the awaited
  device arrives, ignores an unrelated device, resolves on dispose
  instead of hanging.
- `dart format`, `flutter analyze` — clean on all changed files.
- `flutter test test/controllers/connection/ test/controllers/de1_controller_test.dart`
  — all pass (72 tests).
- Full `flutter test` — no regressions; the only failures are
  pre-existing/unrelated (missing `assets/plugins/shot-upload.reaplugin/`
  directory in this environment, reproduces identically on unmodified
  `origin/main`).
- Reviewed by an independent Opus 5 pass and the maintainer review on
  this PR; the previous review round's correctness blockers (unreliable
  `firstWhere` flush, over-broad `_isConnectingMachine` guard) are
  resolved by the root-cause fix described above.

## Impact

- User-visible: fixes spurious "Machine disconnected unexpectedly"
  banners/reconnect churn on the skin during automatic quick-connect
  and USB attach-probe adoption, while the machine is actually still
  connected.
- Behavior change (edge case): `adoptDevice()` no longer resets the
  `de1` stream's exposed value through an interim `null` when replacing
  an already-connected device. Any downstream listener that depended on
  observing a `null` between two adopted devices would see the direct
  transition instead; the only in-tree listener (`DisconnectSupervisor`)
  is the one the bug was about. Internal `_onDisconnect()` bookkeeping
  is unchanged for genuine disconnects.
- No API/spec changes (`assets/api/*.yml` untouched — internal
  connection-manager behavior only).

## Contributor Responsibility

### DIFF
--- a/lib/src/controllers/connection/disconnect_supervisor.dart
+++ b/lib/src/controllers/connection/disconnect_supervisor.dart
@@ -33,11 +33,6 @@ class DisconnectSupervisor {
   Completer<void>? _machineSeenCompleter;
   String? _awaitedMachineId;
 
-  /// Resolves once this supervisor's own machine-stream listener has
-  /// processed a non-null device with the given [deviceId]. Unlike
-  /// subscribing separately to the machine stream, this is driven by the
-  /// listener already registered in [_start], so it cannot race a fresh
-  /// subscription against events already queued for this one.
   Future<void> waitForMachine(String deviceId) {
     if (_latestDe1?.deviceId == deviceId) return Future<void>.value();
     final completer = Completer<void>();
--- a/lib/src/controllers/connection/disconnect_supervisor.dart
+++ b/lib/src/controllers/connection/disconnect_supervisor.dart
@@ -30,6 +30,22 @@ class DisconnectSupervisor {
   device.ConnectionState _latestScaleState = device.ConnectionState.discovered;
   String? _lastKnownMachineId;
 
+  Completer<void>? _machineSeenCompleter;
+  String? _awaitedMachineId;
+
+  /// Resolves once this supervisor's own machine-stream listener has
+  /// processed a non-null device with the given [deviceId]. Unlike
+  /// subscribing separately to the machine stream, this is driven by the
+  /// listener already registered in [_start], so it cannot race a fresh
+  /// subscription against events already queued for this one.
+  Future<void> waitForMachine(String deviceId) {
+    if (_latestDe1?.deviceId == deviceId) return Future<void>.value();
+    final completer = Completer<void>();
+    _awaitedMachineId = deviceId;
+    _machineSeenCompleter = completer;
+    return completer.future;
+  }
+
   StreamSubscription<De1Interface?>? _machineSub;
   StreamSubscription<device.ConnectionState>? _scaleSub;
 
@@ -80,6 +96,11 @@ class DisconnectSupervisor {
       _latestDe1 = de1;
       if (de1 != null) {
         _lastKnownMachineId = de1.deviceId;
+        if (de1.deviceId == _awaitedMachineId) {
+          _awaitedMachineId = null;
+          _machineSeenCompleter?.complete();
+          _machineSeenCompleter = null;
+        }
         _onMachineConnected?.call();
         return;
       }
@@ -173,5 +194,8 @@ class DisconnectSupervisor {
     _scaleSub?.cancel();
     _machineSub = null;
     _scaleSub = null;
+    _awaitedMachineId = null;
+    _machineSeenCompleter?.complete();
+    _machineSeenCompleter = null;
   }
 }
--- a/lib/src/controllers/connection_manager.dart
+++ b/lib/src/controllers/connection_manager.dart
@@ -513,26 +513,21 @@ class ConnectionManager {
         activeTargetTransport: () => machine.transportType,
       ),
     );
-    _isConnectingMachine = true;
-    try {
-      de1Controller.adoptDevice(machine);
-      await de1Controller.de1.firstWhere(
-        (connected) => connected == machine,
-        orElse: () => machine,
-      );
-      await settingsController.setPreferredMachineId(machine.deviceId);
-      _log.info('Attach probe: machine adopted (${machine.deviceId})');
-      if (machine is BengleInterface) {
-        await _attachBengleVirtualScale(machine);
-      } else if (!_scaleConnected) {
-        if (settingsController.preferredScaleId != null) {
-          _ensureScaleReacquisition();
-        } else {
-          _armPostQuickConnectScaleScan();
-        }
+    de1Controller.adoptDevice(machine);
+    await de1Controller.de1.firstWhere(
+      (connected) => connected == machine,
+      orElse: () => machine,
+    );
+    await settingsController.setPreferredMachineId(machine.deviceId);
+    _log.info('Attach probe: machine adopted (${machine.deviceId})');
+    if (machine is BengleInterface) {
+      await _attachBengleVirtualScale(machine);
+    } else if (!_scaleConnected) {
+      if (settingsController.preferredScaleId != null) {
+        _ensureScaleReacquisition();
+      } else {
+        _armPostQuickConnectScaleScan();
       }
-    } finally {
-      _isConnectingMachine = false;
     }
     _publishStatus(currentStatus.copyWith(phase: ConnectionPhase.ready));
   }
@@ -963,10 +958,7 @@ class ConnectionManager {
       final device = await deviceScanner.tryQuickConnect(remembered);
       if (device is De1Interface) {
         de1Controller.adoptDevice(device);
-        await de1Controller.de1.firstWhere(
-          (connected) => connected == device,
-          orElse: () => device,
-        );
+        await _disconnectSupervisor.waitForMachine(device.deviceId);
         _log.info('Quick-connect: machine adopted (${device.deviceId})');
         return device;
       }
@@ -1008,13 +1000,7 @@ class ConnectionManager {
       _publishStatus(
         currentStatus.copyWith(phase: ConnectionPhase.connectingMachine),
       );
-      final De1Interface? qcMachine;
-      _isConnectingMachine = true;
-      try {
-        qcMachine = await _tryQuickConnectMachine();
-      } finally {
-        _isConnectingMachine = false;
-      }
+      final qcMachine = await _tryQuickConnectMachine();
       if (qcMachine != null) {
         if (await _releaseSupersededAutomaticMachine()) return;
         _log.info('Quick-connect: machine connected, proceeding to ready');
--- a/lib/src/controllers/connection_manager.dart
+++ b/lib/src/controllers/connection_manager.dart
@@ -513,21 +513,26 @@ class ConnectionManager {
         activeTargetTransport: () => machine.transportType,
       ),
     );
-    de1Controller.adoptDevice(machine);
-    await de1Controller.de1.firstWhere(
-      (connected) => connected == machine,
-      orElse: () => machine,
-    );
-    await settingsController.setPreferredMachineId(machine.deviceId);
-    _log.info('Attach probe: machine adopted (${machine.deviceId})');
-    if (machine is BengleInterface) {
-      await _attachBengleVirtualScale(machine);
-    } else if (!_scaleConnected) {
-      if (settingsController.preferredScaleId != null) {
-        _ensureScaleReacquisition();
-      } else {
-        _armPostQuickConnectScaleScan();
+    _isConnectingMachine = true;
+    try {
+      de1Controller.adoptDevice(machine);
+      await de1Controller.de1.firstWhere(
+        (connected) => connected == machine,
+        orElse: () => machine,
+      );
+      await settingsController.setPreferredMachineId(machine.deviceId);
+      _log.info('Attach probe: machine adopted (${machine.deviceId})');
+      if (machine is BengleInterface) {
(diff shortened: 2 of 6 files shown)


lib/src/controllers/connection_manager.dart
```diff
@@ -958,6 +963,10 @@ class ConnectionManager {
       final device = await deviceScanner.tryQuickConnect(remembered);
       if (device is De1Interface) {
         de1Controller.adoptDevice(device);
+        await de1Controller.de1.firstWhere(
```
tadelv: Blocking: this `firstWhere` is not a reliable flush of `DisconnectSupervisor`. `de1` is a default-async `BehaviorSubject`, so it is an async broadcast stream; Dart does not guarantee delivery ordering across separate listeners. This newly-created listener can receive the cached/adopted `device` and complete while the already-subscribed supervisor still has the earlier `null` queued. `_isConnectingMachine` can then be cleared before the supervisor processes that `null`, reproducing the original spurious disconnect. Please use synchronization that actually guarantees the supervisor's queued churn has been delivered (or remove the reset/adopt churn), rather than a fresh subscription as a barrier.

lib/src/controllers/connection_manager.dart
```diff
@@ -513,21 +513,26 @@ class ConnectionManager {
...
+    try {
+      de1Controller.adoptDevice(machine);
+      await de1Controller.de1.firstWhere(
+        (connected) => connected == machine,
+        orElse: () => machine,
+      );
+      await settingsController.setPreferredMachineId(machine.deviceId);
+      _log.info('Attach probe: machine adopted (${machine.deviceId})');
+      if (machine is BengleInterface) {
+        await _attachBengleVirtualScale(machine);
```
tadelv: Blocking: `_isConnectingMachine` remains true beyond the internal adopt churn while preference persistence and, for Bengle, virtual-scale attachment are awaited. A genuine machine `null` emitted during this window is not merely delayed: `DisconnectSupervisor` updates `_latestDe1` to null, skips the disconnect callbacks because the guard is true, and never re-evaluates when the flag later becomes false. That can permanently swallow the disconnect/recovery. Please scope the guard narrowly to the reset/adopt churn or explicitly preserve/replay a suppressed real disconnect after the guard closes.

test/controllers/connection/disconnect_supervisor_test.dart
```diff
@@ -0,0 +1,119 @@
...
+      await machineController.close();
+    });
+
+    // De1Controller.adoptDevice() unconditionally resets (emits null) before
+    // emitting the newly adopted De1Interface, even when re-adopting a
+    // machine that was already connected (e.g. a fresh transport object for
+    // the same physical USB device). Any automatic reconnect path that
+    // drives this churn must hold isConnectingMachine for its duration, or
+    // the reset half of the churn reads as an unexpected disconnect.
+    test('reset-then-adopt churn while isConnectingMachine is held through a '
```
tadelv: These tests don't exercise either changed `ConnectionManager` path, `De1Controller.adoptDevice()`, or the production `BehaviorSubject` listener ordering. They would all pass on the pre-PR implementation, so they are not regression tests for this bug. Please add a test through the actual automatic quick-connect/attach adoption path that produces the reset→adopt churn and fails before this fix, then assert that no `machineDisconnected` error is emitted. This is especially important because the bug is specifically about asynchronous cross-listener timing.

test/controllers/connection/disconnect_supervisor_test.dart
```diff
@@ -0,0 +1,119 @@
...
+      );
+    });
+
+    tearDown(() async {
+      supervisor.dispose();
+      statusPublisher.dispose();
+      await machineController.close();
+    });
+
+    // De1Controller.adoptDevice() unconditionally resets (emits null) before
```
tadelv: Repository-guidance issue: `AGENTS.md` says not to add explanatory comments to new/substantially rewritten code and to put rationale/debugging history in the matching `doc/AI_*_NOTES.md`. This file adds several long explanatory blocks/inline comments. Please remove those comments (the test names and assertion reasons can carry the contract) or move the rationale to the appropriate notes doc.

tadelv (changes requested):
There are two correctness issues in the connection guard/synchronization that need addressing before merge, and the new test file also conflicts with the repository's comment-style guidance. CI is green, but the added tests exercise `DisconnectSupervisor` in isolation and would pass without the `ConnectionManager` changes, so they do not establish that this race is fixed in the real path.

lib/src/controllers/connection_manager.dart
allofmeng: Agreed, and fixed at the source rather than trying to make the flush reliable: `De1Controller.adoptDevice()` (de1_controller.dart) no longer emits an interim `null` on the `de1` stream at all when replacing an already-connected device. `_onDisconnect()`'s internal reset (generation bump, dataInitialized reset, subscription teardown) is factored into `_resetForNewDevice()`, which `adoptDevice()` now uses directly, going straight from the old device to the new one in a single emission. There's no more churn for `DisconnectSupervisor` to observe, so nothing to flush for this race. The `firstWhere` call is removed entirely.

This did surface a second, independent, pre-existing race once the interim null was gone: `_releaseSupersededAutomaticMachine()` reads `_machineConnected` (backed by `DisconnectSupervisor`'s own async-updated `_latestDe1`) immediately after `adoptDevice()` returns, and the old code's extra null-then-device pair was accidentally providing enough microtask headroom for the supervisor to catch up first — removing it exposed the race (a usb-attach test started failing). Fixed properly this time: added `DisconnectSupervisor.waitForMachine(deviceId)`, a `Future` resolved from *inside* the supervisor's own pre-existing listener callback rather than a second, independently-scheduled subscription, so there's no cross-listener ordering question at all.

lib/src/controllers/connection_manager.dart
allofmeng: This guard is gone — reverted the `_isConnectingMachine` bracketing entirely from both `_tryQuickConnectMachine` and `_adoptAttachedMachine`. With the root-cause fix (no interim null from `adoptDevice()`), there's no churn left that needs suppressing, so the swallow-a-genuine-disconnect hazard you flagged doesn't apply — the guard was the wrong mechanism and it's not needed once the source stops emitting the spurious event.

test/controllers/connection/disconnect_supervisor_test.dart
allofmeng: Replaced. Removed the DisconnectSupervisor-only tests. Added a real regression test in `test/controllers/de1_controller_test.dart` (`adoptDevice` group) that drives the actual `De1Controller.adoptDevice()` with two real `TestDe1` instances over the real `BehaviorSubject`-backed `de1` stream, asserting the emission sequence has no interim null when replacing an already-adopted device. Confirmed it fails (hangs/times out) against the pre-fix `adoptDevice()` and passes cleanly against the fix.

Also added focused unit tests for the new `DisconnectSupervisor.waitForMachine()` primitive itself in a rewritten `disconnect_supervisor_test.dart`, since that's new synchronization logic worth covering directly (resolves immediately when already current, resolves when the awaited device arrives, ignores an unrelated device, resolves on dispose instead of hanging).

test/controllers/connection/disconnect_supervisor_test.dart
allofmeng: Fixed — the old file is gone (replaced per the reply above). The new `disconnect_supervisor_test.dart` has no explanatory comment blocks, matching the style of the existing files in that directory (e.g. `disconnect_expectations_test.dart`).

One doc comment remains, on the new `DisconnectSupervisor.waitForMachine` method itself, explaining why it's driven by the supervisor's existing listener rather than a fresh subscription. I read that as within AGENTS.md's carve-out ("a hidden constraint... behavior that would surprise a reader") since it's exactly the kind of thing someone might 'simplify' back into a `firstWhere` and reintroduce the bug — but happy to trim or drop it if you'd still rather it not be there.

lib/src/controllers/connection/disconnect_supervisor.dart
```diff
@@ -30,6 +30,22 @@ class DisconnectSupervisor {
   device.ConnectionState _latestScaleState = device.ConnectionState.discovered;
   String? _lastKnownMachineId;
 
+  Completer<void>? _machineSeenCompleter;
+  String? _awaitedMachineId;
+
+  /// Resolves once this supervisor's own machine-stream listener has
```
tadelv: This is still an explanatory/rationale comment in newly added code. The current root `AGENTS.md` does not contain the carve-out mentioned in the reply: it explicitly says not to add explanatory comments to new/substantially rewritten code and to put rationale/debugging history in the matching `doc/AI_*_NOTES.md`. Please keep the method name/API clear here and move the important cross-listener-ordering footgun to `doc/AI_BLE_NOTES.md` (or remove the rationale if you don't think it warrants documentation).

tadelv (changes requested):
The previous correctness blockers are resolved on this revision. The root fix in `De1Controller.adoptDevice()` removes the synthetic `null`, the overly broad `_isConnectingMachine` guard is gone, and the new regression/synchronization tests cover the revised mechanism. I did not find another non-edge-case behavioral blocker in the updated code.

Two required repository-guidance/history fixes remain before merge:

1. The PR title and description are now materially stale. They still say the fix is to hold `_isConnectingMachine` and use `de1.firstWhere(...)`, and the Impact section still describes the widened guard. None of that is true on the current head: the implementation now changes `adoptDevice()` semantics and uses `DisconnectSupervisor.waitForMachine()`. Please update the title, Summary, Verification, and Impact so the PR accurately records what will actually be merged. The PR template is an explicit AGENTS.md pre-PR hard gate.
2. The new explanatory `waitForMachine` comment still conflicts with AGENTS.md's explicit code-style rule. I left an inline comment on that. The cross-listener ordering rationale is useful, but it belongs in the matching AI notes rather than new code comments per repo guidance.

tadelv (approved):
Re-review complete on the updated head. The previous correctness and repository-guidance blockers are resolved: `adoptDevice()` now fixes the spurious disconnect at the source, quick-connect synchronization is tied to `DisconnectSupervisor`'s own listener, the regression coverage exercises the revised behavior, the explanatory rationale has moved to `doc/AI_BLE_NOTES.md`, and the PR title/body now accurately describe the implementation. I found no new non-edge-case blocker in this pass. Latest CI is still running, so this approval is subject to the required checks completing successfully.

PR #750, opened 2026-08-31
chore(shots): track createdAt and updatedAt on shot records

### LINKED ISSUE #676: Add createdAt/updatedAt tracking to ShotRecord for reliable incremental sync
## Problem

`ShotRecord` currently has a `timestamp` for when the extraction happened, but unlike newer mutable entities such as `Bean`, `BeanBatch`, `Grinder`, and `ProfileRecord`, it has no persisted `createdAt` / `updatedAt` metadata.

This makes it impossible for consumers to reliably determine whether an already-seen shot has changed after its initial creation.

A concrete example is the bundled `shot-upload` plugin:

1. a shot is recorded and uploaded;
2. the plugin marks it via `annotations.extras.uploaded_to_decent`;
3. the user later edits notes, enjoyment, bean/batch association, or other annotations;
4. the backlog scanner sees the existing upload marker and skips the shot;
5. the updated shot is never synced.

The `shotUpdated` plugin event helps while Decaid is running, but it is not a durable reconciliation mechanism across restarts/offline periods.

## Proposed change

Bring `ShotRecord` in line with the other mutable persisted entities by adding:

```dart
final DateTime createdAt;
final DateTime updatedAt;
```

Persist both fields in the Drift `shot_records` table and expose them in shot JSON/API responses.

Suggested semantics:

- `timestamp`: when the extraction happened;
- `createdAt`: when the shot record entered Decaid storage;
- `updatedAt`: when meaningful user/content data on the shot last changed.

For new shots, initialize `createdAt` and `updatedAt` to the persistence time.

For existing rows during migration, use `timestamp` as the best available initial value for both fields.

## Important sync/bookkeeping distinction

Not every internal metadata write should necessarily advance the content revision.

For example, `shot-upload` currently writes `annotations.extras.uploaded_to_decent` after a successful upload. If that bookkeeping PUT automatically advances `updatedAt`, then a naive sync rule such as:

```text
updatedAt > uploadedAt
```

would make the act of marking a shot synced immediately make it look dirty again.

We should therefore define one of the following explicitly:

- `updatedAt` tracks meaningful/content mutations and bookkeeping-only sync metadata does not bump it; or
- use a separate revision/hash/sync marker that avoids this feedback loop.

The first option is probably the simpler storage contract.

## Follow-up use in shot-upload

Once durable modification tracking exists, `shot-upload` should be able to distinguish:

```text
never uploaded
```

from:

```text
uploaded previously, but modified locally since then
```

instead of treating the mere presence of `uploaded_to_decent` as permanently synced.

The same model should also make future bean / bean-batch / grinder syncing easier, since those entities already carry `createdAt` / `updatedAt`.

## Acceptance criteria

- [ ] `ShotRecord` exposes `createdAt` and `updatedAt`.
- [ ] `shot_records` persists both fields via Drift.
- [ ] A schema migration initializes existing rows safely, using the shot `timestamp` as the fallback historical value.
- [ ] New shot persistence initializes both fields.
- [ ] Meaningful shot edits advance `updatedAt`.
- [ ] Shot API responses include the timestamps.
- [ ] Semantics for bookkeeping-only metadata writes are documented/tested so sync markers do not create a dirty-loop.
- [ ] Tests cover creation, migration, normal updates, and bookkeeping-only updates.

## Related observation

Bean and bean-batch records already persist `createdAt` / `updatedAt`, and their normal HTTP update paths advance `updatedAt`. Their current backup/export pagination is based on `createdAt`, however, so any future incremental sync should use modification tracking deliberately rather than reusing the backup cursor semantics.

tadelv:
## Summary

What changed, and why?

- Adds `createdAt` and `updatedAt` timestamps to shot records (drift columns, generated code, mapper, model), defaulting to the existing shot timestamp for legacy records.
- Enables durable reconciliation for consumers like the shot-upload plugin: an already-seen shot that later gets edited (notes, bean/batch, annotations) can now be detected via `updatedAt` instead of being skipped on the `uploaded_to_decent` marker.
- Schema version bump 4 → 5 with a migration adding the two nullable columns.`n- Treats `decent_upload_rejected` as explicit sync bookkeeping so uploader rejection markers do not advance `updatedAt`.

## Linked Issue

Fixes #676

## Verification

- `flutter analyze --no-pub` clean.`n- `flutter test --no-pub test/models/shot_record_test.dart test/webserver/shots_handler_test.dart`: 33 passed.
- `flutter test` on the touched suites (de1_controller, shot_sequencer, workflow_device_sync, universal_ble discovery/recovery, remembered_devices, firmware_mmr_exclusion, generate_simulation_assets): 151 passed, 1 skipped.

## Impact

- Drift schema migration (v4 → v5) adds two nullable columns; no data rewrite, legacy shots fall back to their extraction timestamp.
- API/spec documentation now lists `decent_upload_rejected` with the existing bookkeeping keys; no user-visible UI change.

## Contributor Responsibility

### DIFF
--- a/lib/src/models/data/shot_record.dart
+++ b/lib/src/models/data/shot_record.dart
@@ -1,9 +1,15 @@
+import 'dart:convert';
+
 import 'package:reaprime/src/models/data/shot_annotations.dart';
 import 'package:reaprime/src/models/data/shot_snapshot.dart';
 import 'package:reaprime/src/models/data/workflow.dart';
 import 'package:shadcn_ui/shadcn_ui.dart';
 
 class ShotRecord {
+  /// Keys in `annotations.extras` reserved for plugin sync state. Writes
+  /// confined to these keys do not count as content changes.
+  static const bookkeepingExtrasKeys = {'uploaded_to_decent', 'visualizerId'};
+
   final String id;
   final DateTime timestamp;
   final DateTime? createdAt;
@@ -41,8 +47,8 @@ class ShotRecord {
     return {
       "id": id,
       "timestamp": timestamp.toIso8601String(),
-      "createdAt": createdAt?.toIso8601String(),
-      "updatedAt": updatedAt?.toIso8601String(),
+      "createdAt": createdAt?.toUtc().toIso8601String(),
+      "updatedAt": updatedAt?.toUtc().toIso8601String(),
       "measurements": measurements.map((e) => e.toJson()).toList(),
       "workflow": workflow.toJson(),
       if (annotations != null) "annotations": annotations!.toJson(),
@@ -56,8 +62,8 @@ class ShotRecord {
     return {
       "id": id,
       "timestamp": timestamp.toIso8601String(),
-      "createdAt": createdAt?.toIso8601String(),
-      "updatedAt": updatedAt?.toIso8601String(),
+      "createdAt": createdAt?.toUtc().toIso8601String(),
+      "updatedAt": updatedAt?.toUtc().toIso8601String(),
       "workflow": workflow.toJson(),
       if (annotations != null) "annotations": annotations!.toJson(),
       if (stopReason != null) "stopReason": stopReason,
@@ -79,17 +85,17 @@ class ShotRecord {
 
     final legacyTime = DateTime.parse(json["timestamp"]);
     final parsedCreatedAt = json["createdAt"] != null
-        ? DateTime.parse(json["createdAt"] as String)
+        ? DateTime.parse(json["createdAt"] as String).toUtc()
         : null;
     final parsedUpdatedAt = json["updatedAt"] != null
-        ? DateTime.parse(json["updatedAt"] as String)
+        ? DateTime.parse(json["updatedAt"] as String).toUtc()
         : null;
 
     return ShotRecord(
       id: json["id"],
       timestamp: legacyTime,
-      createdAt: parsedCreatedAt ?? legacyTime,
-      updatedAt: parsedUpdatedAt ?? parsedCreatedAt ?? legacyTime,
+      createdAt: parsedCreatedAt ?? legacyTime.toUtc(),
+      updatedAt: parsedUpdatedAt ?? parsedCreatedAt ?? legacyTime.toUtc(),
       measurements: (json["measurements"] as List)
           .map((e) => ShotSnapshot.fromJson(e))
           .toList(),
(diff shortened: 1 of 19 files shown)


lib/src/models/data/shot_record.dart
```diff
@@ -20,11 +22,15 @@ class ShotRecord {
     required this.timestamp,
     required this.measurements,
     required this.workflow,
+    DateTime? createdAt,
+    DateTime? updatedAt,
     this.annotations,
     this.stopReason,
     String? shotNotes,
     Map<String, dynamic>? metadata,
```
ODevStudio, line 40: I think this fallback is too broad for new records. Issue #676 specifies persistence time for newly created records, with the extraction `timestamp` used only as the legacy/migration fallback. As written, normal recordings and historical imports that construct `ShotRecord` without these fields inherit the extraction time, so a shot imported today can have an `updatedAt` from years ago and be missed by an incremental reconciliation cursor. Could we initialize these timestamps at persistence/insertion time and reserve the `timestamp` fallback for legacy rows/JSON?

lib/src/services/webserver/shots_handler.dart
```diff
@@ -320,4 +328,25 @@ class ShotsHandler {
...
+  }
+
+  Map<String, dynamic> _contentOnly(Map<String, dynamic> json) {
+    final copy = Map<String, dynamic>.from(json)
+      ..remove('updatedAt')
+      ..remove('measurements')
+      ..remove('metadata');
+    final annotations = copy['annotations'];
+    if (annotations is Map<String, dynamic>) {
+      copy['annotations'] = Map<String, dynamic>.from(annotations)
```
ODevStudio: I don't think we can exclude all of `annotations.extras` from modification tracking. Elsewhere this field is a general extension surface (tags, flags, custom/plugin data), not only sync bookkeeping. A meaningful change such as editing tags would then persist without advancing `updatedAt`. Could we instead exclude an explicit set of bookkeeping keys, or reserve a dedicated sync namespace?

lib/src/services/webserver/shots_handler.dart
```diff
@@ -320,4 +328,25 @@ class ShotsHandler {
...
+  bool _contentEquals(
+    Map<String, dynamic> merged,
+    Map<String, dynamic> existing,
+  ) {
+    return jsonEncode(_contentOnly(merged)) ==
+        jsonEncode(_contentOnly(existing));
+  }
+
+  Map<String, dynamic> _contentOnly(Map<String, dynamic> json) {
+    final copy = Map<String, dynamic>.from(json)
```
ODevStudio: `measurements` are excluded from change detection here, but the PUT path still deep-merges arbitrary request fields into the full shot and can therefore persist a changed `measurements` array. That would modify shot content without advancing `updatedAt`. If measurements are intentionally not editable through this endpoint, I think we should reject/strip that field rather than only ignoring it for revision calculation.

lib/src/services/webserver/shots_handler.dart
```diff
@@ -216,6 +216,14 @@ class ShotsHandler {
       _synchronizeLegacyAnnotationAliases(merged);
       merged['id'] = id;
 
+      final contentChanged = !_contentEquals(
+        merged,
+        existingShot.toJsonWithoutMeasurements(),
+      );
+      merged['updatedAt'] =
+          (contentChanged ? DateTime.now() : existingShot.updatedAt)
```
ODevStudio: The API docs call these timestamps ISO-8601 UTC, but this creates a local `DateTime`, and `ShotRecord.toJson()` serializes it directly. Could we normalize mutation timestamps with `DateTime.now().toUtc()` (and do the same in the History edit path), or otherwise guarantee UTC serialization?

ODevStudio (changes requested):
Thanks for putting this together — the overall direction makes sense, and the migration/mapping work looks solid. I’m requesting changes because there are a few edge cases in the `createdAt`/`updatedAt` contract that could make durable reconciliation miss real changes. I’ve left the concrete issues inline. None of these require changing the overall approach, but I’d like to tighten the timestamp semantics before we depend on them for sync.

ODevStudio (approved):
Re-reviewed the current head after the follow-up changes. The earlier blocking concerns are addressed:

- new records are stamped at persistence time rather than inheriting an old extraction timestamp;
- bookkeeping suppression is limited to the explicit `uploaded_to_decent` / `visualizerId` keys, so ordinary `annotations.extras` edits still advance `updatedAt`;
- `measurements` writes are rejected instead of being silently excluded from revision tracking;
- mutation timestamps are normalized to UTC, including the History edit path;
- tests cover insertion timestamps, mapper preservation, bookkeeping-only PUTs, normal content PUTs, non-bookkeeping extras, and measurement rejection.

I also checked for additional failure modes with a normal/non-adversarial client in mind. I don't see another practical blocker. A client can technically send unusual fields such as a different `createdAt` through the generic PUT merge, but that requires an atypical caller deliberately altering record metadata and is not something I'd block this PR on.

Approved.

ODevStudio (approved):
Re-reviewed current head `c66022b`, including the two issues raised in comment #5504552587.

Both are addressed:

- `ShotExportSection` now compares full content with `includeMeasurements: true`, so an overwrite that changes only measurements advances `updatedAt`. The focused regression covers this case.
- The v4→v5 migration now backfills `created_at` and `updated_at` from `timestamp`, and the migration test opens a real SQLite file at `user_version = 4`, verifies the persisted backfill, and checks the mapper result.

Full PR checks are green, including Flutter analyze/test and the Linux build. I don't see a remaining code blocker.

Non-blocking: the PR description's Impact section still says "no data rewrite"; that should be updated now that the migration performs the timestamp backfill.

PR #759, opened 2026-09-01
feat(plugins): permission-gated host.transport WebSocket/TCP/TLS transports (#146)

### LINKED ISSUE #146: Plugin host: permission-gated outbound WebSocket + TCP/TLS transports
## Summary

Expose generic, permission-gated outbound network transports to Decaid JavaScript plugins.

Decaid should provide **safe native transport primitives**. Plugins and bundled JavaScript libraries should implement higher-level application protocols such as MQTT.

This issue implements:

- WebSocket (`ws://` / `wss://`), preserving text vs binary frames and supporting WebSocket subprotocol negotiation;
- raw TCP byte streams;
- raw TLS byte streams using the platform trust store.

The workflow event previously included here has been split into #756.

Discussion in #681 is the main concrete consumer: MQTT should be implementable as a plugin using MQTT.js or another JS protocol implementation rather than adding MQTT protocol logic to Decaid core.

## Current state

Verified in `lib/src/plugins/plugin_manager.dart` and the existing plugin runtime:

- `flutter_js` / `dart_js` is initialized with `xhr: false`;
- Decaid supplies its own permission-gated `fetch()` bridge;
- the existing HTTP bridge is request/response oriented and is not a generic bidirectional transport;
- no WebSocket, TCP or TLS API is exposed to plugins;
- the JS↔Dart bridge is JSON-string based, so binary data needs an explicit representation.

For v1, binary data crosses the bridge as base64.

## Architecture boundary

```text
Dart/native transport
        ↓
permission-gated Decaid host API
        ↓
thin JS transport/library adapter
        ↓
MQTT.js / other protocol or device library
        ↓
Decaid plugin
```

Decaid owns connection lifecycle, permission enforcement, resource bounds and safe native cleanup. It must not implement MQTT packet semantics, reconnect policy, QoS, subscriptions, device protocols, or Node-style stream compatibility as part of this issue.

## Public plugin API

The v1 API is `host.transport`. These names and shapes are part of the plugin contract and should not be renamed during implementation.

### `host.transport.open(options)`

Returns a Promise that resolves **once the connection is established**:

```js
{
  handle: "opaque-host-handle",
  protocol: "mqtt" // WebSocket only, omitted/empty when no subprotocol was negotiated
}
```

The handle is opaque and owned by the plugin id + plugin generation that opened it.

#### WebSocket

```js
const opened = await host.transport.open({
  kind: "websocket",
  url: "wss://broker.example/mqtt",
  protocols: ["mqtt"],
});

const handle = opened.handle;
```

WebSocket options for v1:

```text
kind:       "websocket"
url:        required string, ws:// or wss://
protocols:  optional array of WebSocket subprotocol strings
```

Arbitrary custom WebSocket headers are **not** part of v1. Subprotocols use the explicit `protocols` field.

`wss://` uses ordinary platform certificate validation and requires only the WebSocket permission described below.

#### TCP

```js
const { handle } = await host.transport.open({
  kind: "tcp",
  host: "192.168.1.10",
  port: 1883,
});
```

#### TLS

```js
const { handle } = await host.transport.open({
  kind: "tls",
  host: "broker.example",
  port: 8883,
});
```

TCP/TLS options for v1:

```text
kind:  "tcp" or "tls"
host:  required hostname/IP string
port:  required integer 1..65535
```

Raw TLS uses the platform trust store only. Custom CA material and client certificate/private-key authentication are explicitly out of scope and tracked in #758.

### `host.transport.onEvent(handle, callback)`

Registers the single event listener for a handle. Calling it again replaces the previous listener for that handle.

Events that arrive after `open()` resolves but before a listener is registered are retained in the bounded inbound queue and delivered in order once the listener is registered.

#### Data events

WebSocket text frame:

```js
{
  type: "data",
  dataType: "text",
  data: "plain text payload"
}
```

WebSocket binary frame, TCP bytes or TLS bytes:

```js
{
  type: "data",
  dataType: "binary",
  data: "<base64>"
}
```

TCP and TLS are byte streams, not framed text protocols. They only emit `dataType: "binary"`. Text encoding, line framing and request/response semantics belong to plugin/library adapters.

#### Error event

A failure that occurs after a connection has opened is reported as:

```js
{
  type: "error",
  code: "transport_error",
  message: "human-readable description"
}
```

Resource-limit failures use `code: "transport_resource_limit"`.

A terminal transport error must be followed by deterministic transport cleanup; a close event may follow when applicable.

#### Close event

```js
{
  type: "close",
  code: 1000,       // WebSocket close code when available; otherwise omitted
  reason: "..."     // when available; otherwise omitted
}
```

Successful resolution of `host.transport.open()` is the connection-open notification; there is no separate `open` event in v1. Connection failures before establishment reject the `open()` Promise.

### `host.transport.send(handle, payload)`

Returns a Promise.

WebSocket text send:

```js
await host.transport.send(handle, {
  type: "text",
  data: "hello",
});
```

WebSocket binary / TCP / TLS send:

```js
await host.transport.send(handle, {
  type: "binary",
  data: "<base64>",
});
```

For TCP/TLS, `type: "text"` is invalid; adapters must encode text to bytes themselves.

The Promise resolves once the complete payload has been accepted into the native transport's bounded outbound write path. It does **not** mean that the remote peer received or processed the payload.

A send is atomic with respect to the queue limit: if accepting the whole payload would exceed the outbound limit, reject the complete send with `error.code === "transport_resource_limit"`; do not partially enqueue or silently drop it.

### `host.transport.close(handle)`

Returns a Promise that resolves after the native transport is closed and the handle has been released.

Closing an already-closed/stale handle may reject with a normal transport error; it must never affect another plugin or another generation of the same plugin.

## Permissions

Add these manifest permissions exactly:

```text
network.websocket
network.tcp
network.tls
```

Rules:

- `kind: "websocket"` requires `network.websocket` for both `ws://` and `wss://`;
- `kind: "tcp"` requires `network.tcp`;
- `kind: "tls"` requires `network.tls`;
- `wss://` does **not** additionally require `network.tls`;
- permission failure occurs before any DNS lookup or connection attempt and follows the existing `PluginPermissionError` behavior;
- permissions must be surfaced through the existing plugin permission UI/manifest path like other plugin permissions.

Do not expose an unrestricted global `WebSocket`, `Socket`, `dart:io`, or equivalent native networking object in the shared JS runtime.

## Ownership and lifecycle

Every transport handle is owned by `(pluginId, pluginGeneration)`.

Required behavior:

- a handle cannot be used by another plugin;
- a handle cannot be reused by a newer generation after plugin reload;
- stale native events from retired generations are dropped;
- plugin unload closes all transports owned by that generation and removes listeners;
- `PluginManager.dispose()` closes all remaining transports;
- transport cleanup must remain deterministic even when JS `onUnload()` throws or does not explicitly close its connections;
- localhost, LAN and private-address destinations are allowed in v1; there is no hostname/CIDR allow-list in this issue.

## Resource bounds / backpressure

This is not a mission-critical transport and v1 should use simple bounded behavior rather than a complex flow-control API.

Use these implementation defaults:

- maximum **8 live transports per plugin generation**, including connecting/open/closing transports;
- maximum **1 MiB pending outbound data per transport**;
- maximum **1 MiB queued inbound data per transport** waiting for JS delivery.

If opening another transport would exceed the connection limit, reject `open()` with:

```js
error.code === "transport_resource_limit"
```

If a send would exceed the outbound limit, reject that send with the same code. Do not block indefinitely and do not silently drop data.

If inbound queued data exceeds 1 MiB, close that transport with a `transport_resource_limit` error rather than dropping data silently.

The numeric limits are implementation constants, **not configurable fields in the public JS API**. They may be tuned later without changing the plugin contract.

No connection-rate limiter, bandwidth throttling, global per-plugin byte quota or adaptive backpressure protocol is required for v1.

## Binary bridge representation

For v1:

```text
JS bytes -> base64 -> JSON bridge -> Dart Uint8List
Dart bytes -> base64 -> JSON bridge -> JS bytes
```

WebSocket text frames remain strings and must not be base64 encoded.

This deliberately favors compatibility with both QuickJS and JavaScriptCore over maximum throughput. A typed/binary bridge can be optimized separately if profiling later shows a need.

## Implementation guidance

Follow `AGENTS.md` and `doc/AI_REPO_MAP.md`: this is a plugin API/permissions change and should be test-first.

Likely touch points:

### `lib/src/plugins/plugin_manifest.dart`

- add `network.websocket`, `network.tcp`, `network.tls`.

### Plugin host / manager

- expose the exact `host.transport` API above;
- keep native transport state in a small testable bridge/service owned by `PluginManager` rather than adding protocol-specific logic;
- `WebSocket`/`Socket`/`SecureSocket` from Dart are appropriate native primitives;
- authenticate JS bridge messages using the existing plugin bridge token + generation model;
- track handles by plugin id + generation;
- preserve WebSocket text/binary frame type;
- base64 encode/decode binary bridge payloads;
- enforce permissions and resource limits before native work;
- guarantee unload/dispose cleanup.

### `doc/Plugins.md`

Document:

- all three permissions;
- the exact `host.transport` methods/options/event shapes;
- text vs binary representation;
- base64 behavior;
- resource-limit behavior;
- lifecycle/ownership behavior;
- security implications;
- small WebSocket text/binary echo and TCP binary echo examples.

### Contract / permissions

- [ ] `network.websocket`, `network.tcp`, and `network.tls` are recognized manifest permissions; unknown permissions remain rejected.
- [ ] Missing permission rejects before DNS/network activity using existing `PluginPermissionError` behavior.
- [ ] No unrestricted global socket/WebSocket API is added to the JS runtime.

### WebSocket

- [ ] A permitted test plugin opens `ws://` and round-trips a text frame with `dataType: "text"`.
- [ ] A permitted test plugin round-trips random binary bytes with `dataType: "binary"` + base64 bridge representation.
- [ ] WebSocket subprotocol negotiation works through `protocols` and the negotiated protocol is returned by `open()`; `mqtt` is a useful test case.
- [ ] `wss://` works with normal platform certificate validation using only `network.websocket`.
- [ ] A permitted test plugin connects through the host transport to Decaid's own loopback `/ws/v1/sensors/{id}/snapshot` endpoint and receives the text JSON snapshot from a simulated/mock Sensor without hardware.

### TCP/TLS

- [ ] A permitted test plugin opens a TCP echo connection and round-trips random bytes.
- [ ] A permitted test plugin opens a TLS connection using normal platform trust validation and round-trips bytes.
- [ ] TCP/TLS reject text-frame sends; adapters must supply binary/base64 payloads.

### Lifecycle / isolation

- [ ] Handles cannot be used across plugins or plugin generations.
- [ ] Events arriving before `onEvent()` registration are delivered in order once the listener is registered, subject to the inbound bound.
- [ ] Plugin unload closes all owned connections and listeners even if plugin cleanup is faulty.
- [ ] Manager disposal closes all remaining transports.
- [ ] Late events after unload/reload are ignored.

### Resource limits

- [ ] A ninth live transport for one plugin generation is rejected with `transport_resource_limit`.
- [ ] A send that would take pending outbound data over 1 MiB is rejected atomically with `transport_resource_limit`.
- [ ] More than 1 MiB of undelivered inbound data closes that transport with `transport_resource_limit`; data is never silently discarded to stay under the bound.

### Documentation / verification

- [ ] `doc/Plugins.md` documents the final contract and examples.
- [ ] Relevant tests, `flutter analyze`, and full `flutter test` pass per `AGENTS.md`.

## Explicitly out of scope

- `workflowUpdated` / workflow events — #756;
- MQTT protocol implementation or an MQTT service in Decaid core — #681 should consume this capability from a plugin;
- requiring a working MQTT.js plugin as acceptance for this issue;
- custom CA certificates, client certificates or private keys — #758;
- BLE/GATT access from plugins;
- UDP;
- arbitrary WebSocket request headers;
- Node `net`, `tls`, `stream` or browser `WebSocket` compatibility layers;
- a typed/native binary JS bridge;
- hostname/CIDR allow-lists, bandwidth throttling or connection-rate limiting.

The standards-compatible timer work tracked in `tadelv/dart_js#9` has been implemented but is not yet validated here. MQTT.js remains useful downstream integration validation, but it must not block completion of #146.

## Architectural goal

**Decaid owns safe native capabilities; plugins own protocols and integration logic.**

tadelv:
## Summary

Implements #146: permission-gated outbound WebSocket/TCP/TLS transports for Decaid plugins via `host.transport`.

- `lib/src/plugins/plugin_manifest.dart`: new `network.websocket`, `network.tcp`, `network.tls` manifest permissions; unknown permissions stay rejected.
- `lib/src/plugins/plugin_transport_service.dart` (new): owns native `WebSocket`/`Socket`/`SecureSocket` connections keyed by (pluginId, pluginGeneration) — opaque handles, bounded inbound queue with in-order delivery after `onEvent()` registration, flush-based outbound accounting with atomic `transport_resource_limit` rejection, max 8 live transports per generation, deterministic cleanup on unload/dispose.
- `lib/src/plugins/plugin_manager.dart`: per-plugin `host.transport` wrapper with manifest permission gating (JS-side + Dart-side re-check before any DNS/connect), bridge-token-authenticated `transport` message channel, base64 binary bridge representation, listener/pending-op cleanup on unload and disposal. No unrestricted global socket/WebSocket API is exposed.
- `doc/Plugins.md`: documents all three permissions, the exact `host.transport` contract (methods, options, event shapes, text vs base64 binary, resource limits, lifecycle/ownership, security), and WebSocket text/binary + TCP binary echo examples.

## Linked Issue

Fixes #146

## Verification

- New `test/plugins/plugin_manager_transport_test.dart` (19 tests): WS text/binary round-trips, subprotocol negotiation (`mqtt`), loopback `/ws/v1/sensors/{id}/snapshot` text-JSON delivery, TCP echo round-trip, TCP text-send rejection, permission gating before network activity (incl. direct bridge-call re-check), in-order delivery of events arriving before `onEvent()`, handle isolation across plugins and generations, unload/dispose connection cleanup, ninth-transport rejection, atomic oversize-send rejection, inbound-overflow close with `transport_resource_limit`.
- New `test/plugins/plugin_manager_transport_tls_test.dart` (2 tests): TLS and wss reject untrusted self-signed certificates via platform validation.
- Extended `plugin_manifest_permissions_test.dart` for the three new permissions.
- `flutter analyze`: no issues. Full `flutter test`: 3749 passed, 1 pre-existing skip.

## Impact

New plugin API surface (`host.transport`) and three new manifest permissions. New capabilities are opt-in via manifest; existing plugins are unaffected. No endpoint/spec changes; `doc/Plugins.md` updated.

## Contributor Responsibility

### DIFF
--- a/lib/src/plugins/plugin_manager.dart
+++ b/lib/src/plugins/plugin_manager.dart
@@ -9,6 +9,7 @@ import 'package:logging/logging.dart';
 import 'plugin_manifest.dart';
 import 'plugin_decent_proxy_bridge.dart';
 import 'plugin_runtime.dart';
+import 'plugin_transport_service.dart';
 import 'plugin_types.dart';
 import '../services/storage/kv_store_service.dart';
 import '../controllers/de1_controller.dart';
@@ -83,9 +84,11 @@ class PluginManager {
   Future<void>? _disposeFuture;
   final Map<String, int> _retiringPluginGenerations = {};
   final Map<(String, int), Set<Future<void>>> _pluginStorageOperations = {};
+  late final PluginTransportService _transportService;
 
   De1Controller? get de1Controller => _de1controller;
   PluginManagerLifecycle get lifecycle => _lifecycle;
+  int get liveTransportCount => _transportService.liveTransportCount;
   int get attachmentGeneration => _attachmentGeneration;
   int get activeSubscriptionCount =>
       (_de1Subscription == null ? 0 : 1) +
@@ -106,6 +109,9 @@ class PluginManager {
     _decentProxyBridge = PluginDecentProxyBridge(
       decentProxyService: decentProxyService,
       log: _log,
+    );
+    _transportService = PluginTransportService(
+      eventSink: _dispatchTransportEvent,
     );
     _bootstrapJs();
   }
@@ -538,6 +544,79 @@ class PluginManager {
           __timers.clear();
         };
 
+        // Transport bridge: plugin-owned outbound WebSocket/TCP/TLS handles.
+        const __transportListeners = new Map();
+        const __transportPending = new Map();
+        function __sendTransportMessage(message) {
+          __nativeSendMessage("transport", __nativeJsonStringify(message));
+        }
+
+        globalThis.__transportRequest = function (bridgeToken, generation, requestId, type, payload) {
+          __sendTransportMessage({
+            bridgeToken: bridgeToken,
+            generation: generation,
+            requestId: requestId,
+            type: type,
+            payload: payload
+          });
+        };
+
+        globalThis.__transportRegister = function (requestId, entry) {
+          __transportPending.set(requestId, entry);
+        };
+
+        globalThis.__transportSetListener = function (handle, pluginId, listener) {
+          __transportListeners.set(handle, { pluginId: pluginId, listener: listener });
+        };
+
+        globalThis.__handleTransportReply = function (msg) {
+          const pending = __transportPending.get(msg.requestId);
+          if (!pending || pending.bridgeToken !== msg.bridgeToken) return;
+          __transportPending.delete(msg.requestId);
+          if (msg.error) {
+            const error = new __NativeError(msg.error);
+            if (msg.code) error.code = msg.code;
+            pending.reject(error);
+            return;
+          }
+          pending.resolve(msg.result || {});
+        };
+
+        globalThis.__dispatchTransportEvent = function (pluginId, handle, event) {
+          const record = __transportListeners.get(handle);
+          if (!record || record.pluginId !== pluginId || !record.listener) return;
+          try {
+            record.listener(event);
+          } catch (e) {
+            console.error("Transport listener error", pluginId, e);
+          }
+        };
+
+        globalThis.__rejectTransportPendingForToken = function (bridgeToken, error) {
+          for (const [requestId, pending] of __transportPending) {
+            if (pending.bridgeToken !== bridgeToken) continue;
+            __transportPending.delete(requestId);
+            pending.reject(new __NativeError(error));
+          }
+        };
+
+        globalThis.__rejectAllTransportPending = function (error) {
+          for (const pending of __transportPending.values()) {
+            pending.reject(new __NativeError(error));
+          }
+          __transportPending.clear();
+        };
+
+        globalThis.__clearTransportListenersForPlugin = function (pluginId) {
+          for (const [handle, record] of __transportListeners) {
+            if (record.pluginId === pluginId) __transportListeners.delete(handle);
+          }
+        };
+
+        globalThis.__clearAllTransportListeners = function () {
+          __transportListeners.clear();
+        };
+
         Object.defineProperty(globalThis, "host", {
           value: __nativeFreeze({
             log(bridgeToken, generation, message) {
@@ -676,7 +755,8 @@ class PluginManager {
                 permissionDenied: globalThis.host.permissionDenied,
                 fetch: globalThis.__fetchFor,
                 timerSet: globalThis.__timerSet,
-                timerClear: globalThis.__timerClear
+                timerClear: globalThis.__timerClear,
+                transportRequest: globalThis.__transportRequest
               }),
               writable: false,
               configurable: false,
@@ -721,6 +801,18 @@ class PluginManager {
         }
       } catch (e, st) {
         _log.warning("Invalid fetch message", e, st);
+      }
+    });
+
+    js.onMessage("transport", (raw) {
+      try {
+        final msg = raw as Map<String, dynamic>;
+        final pluginId = _pluginIdForBridgeMessage(msg, 'transport');
+        if (pluginId != null) {
+          unawaited(_handleTransportSafely(pluginId, msg));
+        }
+      } catch (e, st) {
+        _log.warning("Invalid transport message", e, st);
       }
     });
   }
@@ -782,6 +874,179 @@ class PluginManager {
       await _handleFetch(pluginId, msg);
     } catch (e, st) {
       _log.warning("Invalid fetch message", e, st);
+    }
+  }
+
+  Future<void> _handleTransportSafely(
+    String pluginId,
+    Map<String, dynamic> msg,
+  ) async {
+    await Future<void>.delayed(Duration.zero);
+    if (_lifecycle != PluginManagerLifecycle.active) return;
+    try {
+      await _handleTransportMessage(pluginId, msg);
+    } catch (e, st) {
+      _log.warning("Invalid transport message", e, st);
+    }
+  }
+
+  Future<void> _handleTransportMessage(
+    String pluginId,
+    Map<String, dynamic> msg,
+  ) async {
+    final requestId = msg['requestId'];
+    if (requestId is! String) {
+      _log.warning('Transport message from $pluginId missing requestId');
+      return;
+    }
+    final bridgeToken = msg['bridgeToken'];
+    final generation = msg['generation'] is num
+        ? (msg['generation'] as num).toInt()
+        : 0;
+    if (generation != _pluginGenerations[pluginId] ||
+        _plugins[pluginId]?.isAlive != true) {
+      _replyTransport(
+        requestId,
+        bridgeToken,
+        error: 'Plugin generation changed',
+      );
+      return;
+    }
+    final type = msg['type'];
+    final payload = msg['payload'];
+    if (payload is! Map) {
+      _replyTransport(
+        requestId,
+        bridgeToken,
+        error: 'Invalid transport message payload',
+      );
+      return;
+    }
+    try {
+      switch (type) {
+        case 'open':
+          final kind = payload['kind'];
+          final permission = switch (kind) {
+            'websocket' => PluginPermissions.networkWebsocket,
+            'tcp' => PluginPermissions.networkTcp,
+            'tls' => PluginPermissions.networkTls,
+            _ => null,
+          };
+          if (permission == null) {
+            _replyTransport(
+              requestId,
+              bridgeToken,
+              error: 'Unknown transport kind',
+            );
+            return;
+          }
+          if (!_hasPermission(pluginId, permission)) {
+            _replyTransport(
+              requestId,
+              bridgeToken,
+              error: _permissionError(pluginId, permission),
+            );
+            return;
+          }
+          final result = await _transportService.open(
+            pluginId: pluginId,
+            generation: generation,
+            options: Map<String, dynamic>.from(payload),
+          );
+          if (!_isCurrentPluginMessage(pluginId, {'generation': generation})) {
+            unawaited(
+              _transportService.close(pluginId, generation, result.handle),
+            );
+            return;
+          }
+          _replyTransport(
+            requestId,
+            bridgeToken,
+            result: {
+              'handle': result.handle,
+              if (result.protocol != null) 'protocol': result.protocol,
+            },
+          );
+        case 'send':
+          final handle = payload['handle'];
+          final sendPayload = payload['payload'];
+          if (handle is! String || sendPayload is! Map) {
+            _replyTransport(
+              requestId,
+              bridgeToken,
+              error: 'Invalid transport send message',
+            );
+            return;
+          }
+          await _transportService.send(
+            pluginId,
+            generation,
+            handle,
+            Map<String, dynamic>.from(sendPayload),
+          );
+          _replyTransport(requestId, bridgeToken, result: const {});
+        case 'close':
+          final handle = payload['handle'];
+          if (handle is! String) {
+            _replyTransport(
+              requestId,
+              bridgeToken,
+              error: 'Invalid transport close message',
+            );
+            return;
+          }
+          await _transportService.close(pluginId, generation, handle);
+          _replyTransport(requestId, bridgeToken, result: const {});
+        case 'onEvent':
+          final handle = payload['handle'];
+          if (handle is String) {
+            _transportService.onEventRegistered(pluginId, generation, handle);
+          }
+        default:
+          _log.warning('Unknown transport message type from $pluginId: $type');
+      }
+    } on PluginTransportException catch (e) {
+      _replyTransport(requestId, bridgeToken, error: e.message, code: e.code);
+    } catch (e) {
+      _replyTransport(requestId, bridgeToken, error: e.toString());
+    }
+  }
+
+  void _replyTransport(
+    String requestId,
+    Object? bridgeToken, {
+    Map<String, dynamic>? result,
+    String? error,
+    String? code,
+  }) {
+    try {
+      js.evaluate(
+        'globalThis.__handleTransportReply('
+        '${jsonEncode({'requestId': requestId, 'bridgeToken': ?bridgeToken, 'result': result, 'error': error, 'code': code})});',
+      );
+      while (js.executePendingJob() > 0) {}
+    } catch (e, st) {
+      _log.warning('Failed to deliver transport reply', e, st);
+    }
+  }
+
+  void _dispatchTransportEvent(
+    String pluginId,
+    int generation,
+    String handle,
+    Map<String, dynamic> event,
+  ) {
+    if (_lifecycle != PluginManagerLifecycle.active) return;
+    if (_plugins[pluginId]?.isAlive != true) return;
+    if (_pluginGenerations[pluginId] != generation) return;
+    try {
+      js.evaluate(
+        'globalThis.__dispatchTransportEvent('
+        '${jsonEncode(pluginId)}, ${jsonEncode(handle)}, ${jsonEncode(event)});',
+      );
+      while (js.executePendingJob() > 0) {}
+    } catch (e, st) {
+      _log.warning('Failed to dispatch transport event to $pluginId', e, st);
     }
   }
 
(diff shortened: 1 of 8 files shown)


lib/src/plugins/plugin_manager.dart
```diff
@@ -538,6 +550,82 @@ class PluginManager {
...
+          __sendTransportMessage({
+            bridgeToken: bridgeToken,
+            generation: generation,
+            requestId: requestId,
+            type: type,
+            payload: payload
+          });
+        };
+
+        globalThis.__transportRegister = function (requestId, entry) {
```
ODevStudio: **Blocking:** could we harden these transport helpers before merging? The plugin runtime is shared, and this installs `__transportRegister` as a writable/configurable global. Another plugin can replace it; when a different plugin later calls `host.transport.*`, this helper receives an entry containing that plugin’s `bridgeToken`. That exposes the victim token, after which the attacker can call the frozen native `transportRequest` using the victim identity and inherit its transport permissions.

The existing bridge already uses non-writable/non-configurable bootstrap hooks for token-sensitive paths. I think the transport registration/reply helpers should follow the same pattern (and ideally avoid echoing the token in replies), with a two-plugin tampering regression test.

lib/src/plugins/plugin_transport_service.dart
```diff
@@ -0,0 +1,601 @@
...
+      );
+      return;
+    }
+    record.inboundQueue.add(_QueuedEvent(event, size));
+    record.inboundBytes += size;
+    _drain(record);
+  }
+
+  void _drain(_TransportRecord record) {
+    if (!record.delivering) return;
```
ODevStudio, line 429: There’s a retention/resource-bound edge case here. If the remote side errors/closes before the plugin ever registers `onEvent()`, `_terminate()` marks the record terminal and queues the final events, then `_drain()` returns here. That means the record is never removed. Since `_checkLiveLimit()` ignores terminal records, a plugin can churn dead handles (and potentially retained inbound buffers) without those records counting toward the 8-transport limit until unload.

Could we keep terminal-but-undrained records bounded — e.g. count them toward the quota, expire/release them, or otherwise cap the retained terminal state — and add a regression test for the no-listener case?

lib/src/plugins/plugin_transport_service.dart
```diff
@@ -0,0 +1,601 @@
...
+  Future<void> _runOutboundDrain(_TransportRecord record) async {
+    try {
+      while (!record.terminal && record.outboundQueue.isNotEmpty) {
+        final frame = record.outboundQueue.first;
+        final ws = record.webSocket;
+        if (ws == null) break;
+        try {
+          await ws.addStream(Stream<dynamic>.value(frame.data));
+        } on Object {
+          break;
```
ODevStudio, line 315: If `ws.addStream()` fails, breaking here leaves the frame at the head of `outboundQueue` and leaves its byte reservation in `outboundBytes`. `send()` has already resolved by this point, so the plugin gets no failure signal; later sends keep retrying the same frame and can eventually hit `transport_resource_limit` because the accounting never clears.

Would it be better to treat a native write failure as a terminal transport error and deterministically release/clear the queued accounting? A failing-WebSocket regression test would catch this.

test/plugins/plugin_manager_transport_tls_test.dart
```diff
@@ -0,0 +1,326 @@
...
+        ),
+        onError: (Object _) {},
+      );
+      addTearDown(() async => server.close());
+
+      final tlsManager = PluginManager(
+        kvStore: FakeKeyValueStoreService(),
+        fetchTimeout: const Duration(seconds: 2),
+        connectSecureSocket: (host, port) =>
+            SecureSocket.connect(host, port, onBadCertificate: (_) => true),
```
ODevStudio, line 223: Small test-coverage gap: this positive TLS case accepts any bad certificate, so it verifies the transport plumbing but not the production “platform trust store only” success path. The positive WSS test below uses the same bypass. The rejection tests already cover an untrusted cert nicely.

Could we add a positive trusted-certificate path that exercises certificate validation rather than `onBadCertificate` / `badCertificateCallback`? I’d treat this as secondary to the bridge/security fix.

ODevStudio (changes requested):
Thanks for putting this together — this is a substantial change, and the overall direction looks good. The API shape, permission re-checks, ownership model, lifecycle handling, and coverage are all thoughtful.

I’m requesting changes because there’s one cross-plugin security boundary issue in the shared JS bridge that I think we need to fix before merging. I’ve left that blocker inline, along with two resource/accounting edge cases and one smaller TLS/WSS test-coverage gap.

Once the bridge-token boundary is hardened, the remaining items look fairly contained.

lib/src/plugins/plugin_manager.dart
tadelv: Addressed in 7b1baf1a. All transport bridge helpers (__transportRegister, __handleTransportReply, __transportSetListener, __transportRemoveListener, __transportRequest, __dispatchTransportEvent, __rejectTransportPendingForToken, __rejectAllTransportPending, __clearTransportListenersForPlugin, __clearAllTransportListeners) are now installed via Object.defineProperty as non-writable, non-configurable, non-enumerable globals, matching the existing token-sensitive bootstrap hooks. __transportRegister and __handleTransportReply no longer accept replacement, so a plugin cannot intercept another plugin's bridgeToken (and the frozen pluginHostBridge.transportRequest cannot be redirected). Added a two-plugin regression test: the attacker attempts Object.defineProperty redefinition of all six helpers, must observe :blocked for each, and the victim's WebSocket round trip still works.

lib/src/plugins/plugin_transport_service.dart
tadelv: Addressed in 7b1baf1a. Terminal-but-undrained records now count toward the 8-transport quota: _checkLiveLimit counts records that are either live or terminal with undelivered queued events (a peer-closed transport that has not yet been drained via onEvent registration). A plugin can therefore retain at most 8 dead handles (each bounded by the 1 MiB inbound cap) instead of an unbounded number until unload. Added a regression test: 8 listenerless transports closed by the peer make the 9th open reject with transport_resource_limit, with liveTransportCount == 0.

lib/src/plugins/plugin_transport_service.dart
tadelv: Addressed in 7b1baf1a. A failed ws.addStream() is now treated as a terminal transport error: the drain clears the outbound queue and byte accounting and terminates the record with an error event (delivered to a registered onEvent listener), so send() callers get a terminal signal and the accounting cannot accumulate into a phantom transport_resource_limit. TCP socket.add() failures release their reservation before rethrowing. Added a regression test with a failing fake WebSocket asserting the error event, a rejected follow-up send, and liveTransportCount == 0.

test/plugins/plugin_manager_transport_tls_test.dart
tadelv: Addressed in 7b1baf1a (documented; code unchanged for this one). I verified empirically on this platform (macOS, Flutter's dart:io) that a positive trusted-certificate path cannot be exercised offline: both SecureSocket.connect(host, port, context: SecurityContext()..setTrustedCertificates(ca)) and WebSocket.connect(url, customClient: HttpClient(context: trustContext)) fail with CERTIFICATE_VERIFY_FAILED for a local server cert signed by that CA — dart:io ignores user-supplied trust anchors here. The rejection tests prove the production default connectors validate against the platform store, and the seam-based positive tests exercise the full transport path end-to-end; the limitation is now documented in doc/AI_TESTING_NOTES.md. A genuine trusted-cert success test would require a cert the OS trust store accepts, which cannot be minted offline without modifying system trust.

lib/src/plugins/plugin_manager.dart
```diff
@@ -538,6 +553,81 @@ class PluginManager {
...
+            bridgeToken: bridgeToken,
+            generation: generation,
+            requestId: requestId,
+            type: type,
+            payload: payload
+          });
+        });
+        __frozenTransportGlobal("__transportRegister", function (requestId, entry) {
+          __transportPending.set(requestId, entry);
+        });
```
ODevStudio, line 591: One remaining blocker in the hardening: these private maps still use normal `Map` method dispatch (`__transportPending.set/get/delete`, and likewise the listener map). Because plugins share the JS runtime, an earlier plugin can replace `Map.prototype.set` before a later plugin calls `host.transport.*`. The replacement then receives this `entry`, including the victim’s `bridgeToken`, so the token can still be captured even though `__transportRegister` itself is frozen.

This is deterministic for a malicious plugin, not a low-probability transport edge case. The bootstrap already captures `Map.prototype.set/get/delete` as `__nativeMap*` and wraps them in `__mapSet/__mapGet/__mapDelete` for the existing bridge. Could we use those captured helpers for the transport maps too (and ideally test prototype tampering, not only redefining the globals)?

### AI worker prompt

Please fix this finding on the current PR branch. Follow `AGENTS.md` and keep the change narrowly scoped to hardening the existing `host.transport` bridge; do not redesign the plugin runtime, change the public `host.transport` contract, alter permissions, or refactor unrelated code.

**Goal:** no plugin-controlled mutation of `Map.prototype` (including iteration-related methods) may expose another plugin's transport pending entries, listener records, or bridge token, or break transport bookkeeping through the new transport maps.

**Implementation requirements:**
1. In `PluginManager._bootstrapJs`, use the native `Map` methods captured before any plugin code runs. Reuse the existing `__mapSet`, `__mapGet`, and `__mapDelete` helpers for every access to `__transportPending` and `__transportListeners`.
2. Harden the remaining transport-map operations too. Do not leave direct calls such as `.clear()`, `.forEach()`, `.values()`, `.entries()`, or `for...of map` where a plugin can replace a `Map.prototype` method/iterator. Capture the minimal additional native primitive(s), preferably `Map.prototype.clear` and `Map.prototype.forEach`, and invoke them through the already captured `Reflect.apply` pattern. This avoids relying on `Map.prototype[Symbol.iterator]` during cleanup.
3. Preserve all current behavior: request/reply matching, token checks, listener replacement/removal, unload cleanup, manager-dispose cleanup, event ordering, permissions, and the public JS API must remain unchanged.
4. Do not solve this by exposing the private maps, moving tokens into globals, weakening token validation, or removing the cross-plugin isolation checks.

**Regression test first:** extend `test/plugins/plugin_manager_transport_test.dart` with a two-plugin prototype-tampering test. Have an attacker plugin save the original native `Map` methods, then replace relevant `Map.prototype` methods (at minimum `set`, `get`, `delete`, `clear`, and the iterator/iteration path) with wrappers that still delegate to the originals but record any transport pending entry containing a `bridgeToken`/`resolve`/`reject`. Load/use a victim plugin afterwards and perform a real WebSocket transport operation. Assert that:
- the victim transport still opens/sends/receives normally;
- the attacker captures no victim bridge token or pending entry;
- cleanup/unload still works;
- the test fails against the vulnerable implementation, not merely against a completely broken monkey patch.

Keep the existing global-redefinition regression test; this new test should specifically cover prototype tampering.

**Verification:** after the change run, in this order:
- `dart format lib test`
- `flutter test test/plugins/plugin_manager_transport_test.dart`
- `flutter analyze`
- full `flutter test`

Fix any failures caused by the change. When replying here, summarize the exact hardening performed, name the regression test added, and include the verification results. Do not resolve this thread based only on code inspection; the prototype-tampering regression must pass.

ODevStudio (changes requested):
Re-reviewed the latest head. The earlier resource/lifecycle findings are addressed, and I’ve resolved those threads. I’m also fine with the documented TLS/WSS trust-store test limitation; I don’t consider that a merge blocker.

There is one remaining security-boundary issue in the bridge hardening itself. This one is not a rare runtime edge case: a plugin can deterministically monkey-patch `Map.prototype` before another plugin uses the transport API and observe the token-bearing pending entry. I’ve left the specific fix inline. Once that is hardened using the already-captured native Map helpers, I don’t see another blocker in this pass.

lib/src/plugins/plugin_manager.dart
tadelv: Addressed in 96dd279f. Every access to __transportPending and __transportListeners now goes through the captured native helpers: __mapSet/__mapGet/__mapDelete (existing) plus newly captured Map.prototype.clear and Map.prototype.forEach (__nativeMapClear/__nativeMapForEach, invoked via __mapClear/__mapForEach through the __nativeReflectApply pattern). No direct .set/.get/.delete/.clear/.values()/.entries() or for...of remains on the transport maps, including in __rejectTransportPendingForToken, __rejectAllTransportPending, __clearTransportListenersForPlugin, and __clearAllTransportListeners. Behavior is unchanged: request/reply matching, token checks, listener replacement/removal, unload/dispose cleanup, event ordering, permissions, and the public JS API are identical.

Regression test added: "a plugin cannot monkey-patch Map.prototype to capture another plugin token" in test/plugins/plugin_manager_transport_test.dart. An attacker plugin saves the native Map methods, then replaces Map.prototype set/get/delete/clear/forEach/has/values/entries/keys/[Symbol.iterator] with delegating wrappers that sniff any value carrying resolve+reject+bridgeToken (a transport pending entry). A victim plugin loaded afterwards performs a real WebSocket open/send/echo. Assertions: victim round-trip works, attacker captures nothing, and per-plugin unload cleanup succeeds. The test fails against the pre-hardening implementation (captures the victim token via the patched set during __transportRegister) and passes with the fix; the global-redefinition test is kept as-is.

Verification (in order): dart format lib test clean; flutter test test/plugins/plugin_manager_transport_test.dart 27 passed; flutter analyze no issues; full flutter test 3759 passed, 1 pre-existing skip.

ODevStudio (approved):
Re-reviewed the latest head. The transport bridge hardening now uses captured native Map operations throughout the pending/listener bookkeeping, including cleanup iteration, and the new two-plugin regression test specifically covers Map.prototype tampering while exercising a real WebSocket round-trip and unload cleanup.

The previous robustness findings are also addressed, and the current PR checks are green (format, analyze, Flutter tests, Linux build smoke). I don’t see a remaining merge blocker under the project’s trusted-plugin / trusted-LAN operating assumptions. Further protection against deliberately hostile shared-runtime mutation would be defense-in-depth rather than something I’d hold this PR for.

Approved.

PR #762, opened 2026-09-01
feat(sensor): expose Bengle EBus tap as a raw Sensor

tadelv:
## Summary

- expose the supported Bengle USB tap through the existing Sensor API
- preserve raw bytes for snapshot and write operations
- support desktop and Android discovery while preserving existing machine identity
- handle duplicate USB descriptors using physical ownership

## Safety

- discovery uses explicit USB identity and does not actively probe the tap
- binary transport has single-reader connection semantics and skips text decoding
- no protocol parsing, capture, compression, or upload behavior is included

## Validation

- `flutter analyze`
- focused Sensor, serial discovery, identity, and factory tests
- full `flutter test`: 3777 passed, 1 skipped
- Android hardware validated; additional hardware details available out of band

### DIFF
--- a/lib/src/services/serial/serial_service_desktop.dart
+++ b/lib/src/services/serial/serial_service_desktop.dart
@@ -384,6 +384,7 @@ class SerialServiceDesktop implements DeviceDiscoveryService {
     int? vid;
     int? pid;
     int? interfaceNumber;
+    String? productName;
     try {
       vid = port.vendorId;
     } catch (_) {}
@@ -393,12 +394,24 @@ class SerialServiceDesktop implements DeviceDiscoveryService {
     try {
       interfaceNumber = port.interfaceNumber;
     } catch (_) {}
-    if (isBengleEbusTap(vid: vid, pid: pid, interfaceNumber: interfaceNumber)) {
+    try {
+      productName = port.productName;
+    } catch (_) {}
+    if (isBengleEbusTap(
+      vid: vid,
+      pid: pid,
+      interfaceNumber: interfaceNumber,
+      productName: productName,
+    )) {
       _log.info(
         "Bengle EBus tap on interface $interfaceNumber ($id)"
         " — no protocol probe",
       );
-      final transport = _DesktopSerialPort(port: port, dtrOn: true);
+      final transport = _DesktopSerialPort(
+        port: port,
+        dtrOn: true,
+        decodeUtf8Text: false,
+      );
       _portPathToTransport[id] = transport;
       final device = BengleDebugPort(transport: transport);
       _portPathToDeviceId[id] = device.deviceId;
@@ -559,6 +572,7 @@ const int _serialWriteTimeoutMs = 500;
 class _DesktopSerialPort implements SerialTransport {
   final SerialPort _port;
   final bool _dtrOn;
+  final bool _decodeUtf8Text;
   late Logger _log;
   final BehaviorSubject<ConnectionState> _open = BehaviorSubject.seeded(
     ConnectionState.discovered,
@@ -570,9 +584,13 @@ class _DesktopSerialPort implements SerialTransport {
   late final String _cachedId = _computeId();
   late final String _cachedName = _safePortName() ?? "Unknown port";
 
-  _DesktopSerialPort({required SerialPort port, bool dtrOn = false})
-    : _port = port,
-      _dtrOn = dtrOn {
+  _DesktopSerialPort({
+    required SerialPort port,
+    bool dtrOn = false,
+    bool decodeUtf8Text = true,
+  }) : _port = port,
+       _dtrOn = dtrOn,
+       _decodeUtf8Text = decodeUtf8Text {
     _log = Logger("SerialPort:${port.name}");
     _cachedId;
     _cachedName;
@@ -729,6 +747,7 @@ class _DesktopSerialPort implements SerialTransport {
       _portSubscription = reader.stream.listen(
         (data) {
           _rawStreamController.add(data);
+          if (!_decodeUtf8Text) return;
           try {
             final input = utf8.decode(data);
             _log.finest("received serial input: $input");
--- a/lib/src/services/serial/serial_service_desktop.dart
+++ b/lib/src/services/serial/serial_service_desktop.dart
@@ -11,6 +11,7 @@ import 'package:reaprime/src/models/device/impl/bengle/bengle.dart';
 import 'package:reaprime/src/models/device/impl/de1/de1.models.dart';
 import 'package:reaprime/src/models/device/impl/de1/unified_de1/unified_de1.dart';
 import 'package:reaprime/src/models/device/impl/decent_scale/scale_serial.dart';
+import 'package:reaprime/src/models/device/impl/sensor/bengle_debug_port.dart';
 import 'package:reaprime/src/models/device/impl/sensor/debug_port.dart';
 import 'package:reaprime/src/models/device/impl/sensor/sensor_basket.dart';
 import 'package:reaprime/src/models/device/transport/serial_port.dart';
@@ -333,6 +334,7 @@ class SerialServiceDesktop implements DeviceDiscoveryService {
     int? vid;
     int? pid;
     String? serial;
+    int? interfaceNumber;
     try {
       name = port.name ?? path;
     } catch (_) {}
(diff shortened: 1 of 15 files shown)


lib/src/services/serial/usb_ids.dart
```diff
@@ -23,3 +23,39 @@ UsbDeviceModel? matchUsbDevice(
...
+/// The installed usb_serial fork treats the requested interface as the CDC
+/// bulk-data interface and derives the control interface as `iface - 1`.
+/// The tap's Android bulk-data interface is therefore 3 (Linux control
+/// interface 2). Do NOT change this back to 2 — interface 2 is the control
+/// interface, which has no bulk endpoints and makes `UsbPort.open()` fail.
+const int bengleEbusAndroidDataInterface = 3;
+
+/// True only for the Bengle EBus tap: VID 0x2e8a, PID 0x000a, interface 2.
+/// Everything else — interface 0, missing metadata, and devices without
+/// interface 2 — is false.
```
ODevStudio: **Blocker — the VID/PID is shared, so this can open the wrong RP2040 device.**

`0x2e8a:0x000a` is Raspberry Pi’s standard Pico SDK CDC UART allocation, not a Decent-specific product identity. Raspberry Pi’s USB PID registry documents that these standard IDs are shared and recommends distinguishing products with USB string descriptors. Right now the tap helpers accept only VID/PID plus interface layout. On Android that means any unrelated composite Pico using this standard PID and enough interfaces can be exposed as a Bengle tap; the Sensor controller then auto-connects it, which opens the interface and asserts DTR. This does not require a bad actor — it is ordinary USB misidentification with an active side effect.

Decaid already uses the exact USB product name `Bengle` as a device discriminator elsewhere. Keep the strict VID/PID/interface checks, but also require the Bengle product string. Do **not** replace this with a protocol probe or discovery-time write.

Reference: Raspberry Pi USB PID registry: https://github.com/raspberrypi/usb-pid

### DeepSeek worker prompt

You are fixing PR #762 in `decentespresso/decaid`. Implement only the targeted identity hardening below; do not redesign discovery.

1. Strengthen Bengle EBus tap identity so VID `0x2e8a` + PID `0x000a` + interface layout alone is **not** sufficient. Require the exact USB product name `Bengle` as well.
2. Update `isBengleEbusTap` and `isBengleCompositeWithTap` (or one small shared helper they call) to accept/check `productName`. A null or different product name must return false.
3. Update desktop call sites to pass `port.productName`. Update Android call sites to pass `device.productName`.
4. Preserve the existing logical interface `2` identity and Android bulk-data interface `3`. Do not change machine IDs, duplicate-device ownership, raw-byte behavior, or DTR behavior for a correctly identified Bengle.
5. Do not add active probing, writes, or path-based identity during discovery.
6. Add regression tests for a device whose VID/PID are exactly `0x2e8a/0x000a` and whose interface/interface-count would otherwise match, but whose product name is `Pico` (or another non-`Bengle` value). It must not expose a Bengle tap. Keep positive `Bengle` cases passing on desktop/helper and Android paths where practical.
7. Update comments/docs that currently state VID/PID+interface alone uniquely identify the tap.
8. Run `dart format`, `flutter analyze`, the focused tap/serial tests, and full `flutter test`. Do not weaken existing tests to make the change pass.

lib/src/services/serial/serial_service_desktop.dart
```diff
@@ -370,6 +381,30 @@ class SerialServiceDesktop implements DeviceDiscoveryService {
...
+    } catch (_) {}
+    try {
+      interfaceNumber = port.interfaceNumber;
+    } catch (_) {}
+    if (isBengleEbusTap(vid: vid, pid: pid, interfaceNumber: interfaceNumber)) {
+      _log.info(
+        "Bengle EBus tap on interface $interfaceNumber ($id)"
+        " — no protocol probe",
+      );
+      final transport = _DesktopSerialPort(port: port, dtrOn: true);
```
ODevStudio: **Blocker — the desktop tap still enters the UTF-8/text path for every binary chunk.**

Android explicitly constructs the tap transport with `decodeUtf8Text: false`, but the desktop constructor here only sets `dtrOn: true`. `_DesktopSerialPort` still does `utf8.decode(data)` for every read after forwarding the raw bytes. EBus is arbitrary binary, so invalid UTF-8 is normal rather than exceptional; this will repeatedly throw/catch `FormatException` and log decode failures, while valid chunks are unnecessarily copied into the unused text stream. It also contradicts this PR’s stated safety/transport contract that the binary path skips text decoding.

The clean fix is to mirror the Android opt-out while leaving every existing text serial device unchanged.

### DeepSeek worker prompt

You are fixing PR #762 in `decentespresso/decaid`. Make the desktop Bengle EBus tap genuinely binary-only without changing behavior for any other serial device.

1. Add a private boolean such as `_decodeUtf8Text` to `_DesktopSerialPort`, with constructor parameter `decodeUtf8Text` defaulting to `true`.
2. At the Bengle tap construction site, pass `decodeUtf8Text: false` together with `dtrOn: true`.
3. In the serial reader callback, always publish the incoming `Uint8List` to `_rawStreamController` first. Immediately return when `_decodeUtf8Text` is false. Only when it is true should the existing `utf8.decode`, text logging, and `_readController.add` logic run.
4. Do not change framing, chunk boundaries, raw bytes, write behavior, DTR, or the default behavior of DE1/Bengle-machine/HDS/other serial transports.
5. Add a focused regression test if the existing desktop serial test seams can cover this without building a large FFI mocking layer. The important assertion is that binary mode forwards raw bytes unchanged and does not emit to the text stream or attempt UTF-8 decoding. Do not overengineer a new test harness solely for this.
6. Run `dart format`, `flutter analyze`, focused serial/tap tests, full `flutter test`, and ensure the Linux build smoke path still succeeds. Do not suppress decode errors globally; the text path should remain unchanged for normal serial devices.

ODevStudio (changes requested):
CI is green, and the core raw-byte/lifecycle work is well covered. I’m requesting changes for two issues that occur in normal operation rather than adversarial or once-in-a-blue-moon scenarios:

1. The Bengle tap identity gate treats Raspberry Pi’s shared `0x2e8a:0x000a` Pico SDK CDC VID/PID as if it were Decent-specific. Because discovered Sensors are auto-connected, an unrelated composite RP2040 device can be opened and have DTR asserted. This needs an additional product discriminator without adding an active protocol probe.
2. The desktop tap still runs every raw chunk through UTF-8 decoding, unlike Android’s explicit binary-only path. Arbitrary EBus bytes make decode failures normal, producing avoidable exception/log churn and contradicting the PR’s stated binary transport behavior.

I am not blocking this PR on speculative races, null-detach cleanup, or other very-low-frequency edge cases that are not dangerous. Each inline blocker below ends with a concrete DeepSeek worker prompt.

lib/src/services/serial/usb_ids.dart
tadelv: Fixed in 9feb6f67. Tap identity now additionally requires the exact USB product name `Bengle` on helper, desktop, and Android paths. Null/non-matching products are rejected without probing or writes, with focused regressions.

lib/src/services/serial/serial_service_desktop.dart
tadelv: Fixed in 9feb6f67. The desktop tap now opts out of UTF-8 decoding while preserving raw-first delivery; text decoding remains enabled by default for every other serial transport. There is no narrow desktop FFI test seam, so I kept the implementation surgical rather than adding a broad harness. Analyze and the full test suite are green; Linux smoke remains covered by CI.

ODevStudio (approved):
Re-reviewed `9feb6f6` against the two requested blockers and the resulting full PR diff.

Both blockers are resolved:

1. Bengle tap identity now requires the exact `Bengle` product string in addition to VID/PID and the platform-specific interface evidence. Desktop and Android pass the product metadata through, null/non-matching products fail closed, and the focused regressions verify that shared Pico VID/PID devices are not opened. Discovery still performs no probe or write.
2. The desktop tap now sets `decodeUtf8Text: false`; raw chunks are published before the binary-mode early return, while `_DesktopSerialPort` keeps `true` as the default for every existing text transport.

I found no new realistic correctness or safety blocker in the fix. I am not blocking on the absent desktop FFI regression harness or speculative low-frequency cleanup races. Current-head CI is green: format, Flutter analyze/tests, and Linux build/AppImage smoke all passed.

PR #773, opened 2026-09-03
fix(ble): harden background scale-watch ownership

### LINKED ISSUE #451: tech debt(ble): harden background scale-watch ownership and teardown semantics
## Context

PR #447 introduced the persistent Android background scale watch and fixed the four lifecycle gaps raised during review. A post-merge re-audit of the merged implementation (`143443f`, PR head `0e99c65e`) found four remaining ownership/lifecycle edge cases.

These are follow-up technical debt rather than a request to revert the feature. The highest-risk cases can violate the feature's core invariant: **the native BLE scan must be known to be stopped before GATT connection work or another scan starts**.

### 1. P1 — `ScaleWatch` can subscribe while a burst/full scan owns `deviceStream`

`ConnectionManager._connectImpl()` calls `_cancelScaleReacquisition()`, but watch disarm is fire-and-forget. A machine connection observed during the full scan can call `_ensureScaleReacquisition()` before `ScanOrchestrator.runScan()` and its policy phase finish.

`UniversalBleDiscoveryService.startDeviceWatch()` records `_watchRequested` and returns successfully when `_isScanning` is true. `ScaleWatch.arm()` interprets that return as an active watch and installs its `deviceStream` subscription. It can therefore consume advertisements belonging to the burst and race `EarlyConnectWatcher`.

The cached-device path also connects directly from `scanner.devices` without checking whether the full-scan policy currently owns connection decisions.

**Desired behavior**

- Full scans exclusively own preferred-device discovery and connection decisions until their policy phase completes.
- A watch requested during a burst is represented as queued, not active.
- `ScaleWatch` must not subscribe until a real watch lease is active.

### 2. P1 — native `stopScan()` failure is treated as successful ownership handoff

`UniversalBleDiscoveryService._deactivateWatchScan()` clears `_watchScanActive` before awaiting `UniversalBle.stopScan()`, catches stop failures, and returns normally. `DeviceController.stopScaleWatch()` suppresses service-level errors as well.

Consequences:

- `ScaleWatch._onSighting()` may start scale GATT connection work even though the native scan could still be active.
- `_pauseWatchForBurst()` may start a low-latency burst after failing to stop the balanced watch.
- Local state says `inactive` even though native ownership is unknown.

**Desired behavior**

- Ownership-transfer stops used before GATT or another scan must propagate failure.
- Best-effort suppression should be reserved for final disposal.
- A failed stop should enter an explicit `faulted`/`unknown` state.
- No GATT connection or new scan may begin while native ownership is unknown.

### 3. P1 — initial watch-start failure can leave a latent `_watchRequested`

`startDeviceWatch()` stores `_watchRequested` before attempting the native start. If the initial `startScan()` throws, the request is not cleared.

`ScaleWatch` marks itself unarmed and activates the legacy fallback, but `disarm()` can return early because no local subscription exists. A later burst completion or adapter recovery can therefore resurrect the lower-level watch request without a `ScaleWatch` consumer, potentially alongside the legacy reconnect loop.

**Desired behavior**

- Initial watch start must be transactional.
- Clear the request on failure when it still belongs to the failed request generation.
- `disarm()` must always release lower-layer requested state, even when local `_armed == false`.

### 4. P2 — stale subscriptions can be installed after `disarm()`

Both `ScaleWatch.arm()` and the failed-connect restart path await `_startWatchScan()` and then call `_listen(gen)` without rechecking `_armed`, `_generation`, or `_shouldWatch()`.

If `disarm()` runs during the await, the old continuation can install `_sub` and `_failureSub` afterward. The generation guard prevents connection attempts, but the subscriptions remain allocated and lifecycle teardown is no longer deterministic.

**Desired behavior**

After every awaited start/restart, validate:

```dart
if (!_armed || gen != _generation || !_shouldWatch()) {
  await _scanner.stopScaleWatch();
  return;
}
```

Only then install listeners.

## Field evidence: #598 scale hard-power-cycle after machine sleep

Issue #598 provides a field reproduction that is consistent with this ownership/lifecycle class and gives us a concrete end-to-end regression scenario.

Observed sequence on Android:

1. DE1 and preferred Decent Scale initially connect normally.
2. User puts the machine into sleep.
3. The USB-powered scale is physically powered off, then powered back on.
4. Machine is woken.
5. Scale does not reconnect reliably.
6. `Retry` reports `Scale not found`; `Reconnect` does nothing; the scale is marked `Unavailable` and disappears from the device list.
7. Restarting Decaid changes the behavior and allows recovery again.

The reporter reproduced this on 0.7.16, 0.7.17, and 0.8.0-beta.1. The symptom is stronger than a failed scale GATT initialization: while wedged, Decaid's own discovery path no longer sees the advertising scale. That points first at scan/watch lifecycle or native scan ownership rather than the Decent Scale protocol itself.

There is a second possible contributor worth distinguishing in tests: `UniversalBleDiscoveryService` de-duplicates devices by `deviceId` while an existing `Device` remains in `_devices`. If Android misses the disconnect callback during hard power loss, a stale device/transport instance may suppress fresh advertisements until its state is removed. App restart naturally clears that in-memory state.

### Diagnostic split for #598

After reproducing the failure, without restarting Decaid:

1. Run discovery-only (`/api/v1/devices/scan?connect=false`).
2. If the powered-on scale is still not discovered, treat the failure as scan/watch ownership/lifecycle first.
3. If it is rediscovered but fails to connect, investigate stale device/transport/GATT state separately.
4. Toggle the Android Bluetooth adapter off/on as an additional diagnostic. Recovery after adapter reset without an app restart would strongly support stale native scan/watch state.

### `universal_ble` version note

The original 0.7.16 report used `universal_ble` 2.2.5. Final Decaid v0.8.0 pins 2.2.6. That release includes an Android fix to report scans as active only after `BluetoothLeScanner` has actually started. Because #598 was last reproduced on 0.8.0-beta.1 rather than final 0.8.0, the first field retest should be on final 0.8.0 before attributing the remaining failure entirely to Decaid's higher-level watch coordinator.

## Suggested design direction

Avoid adding more independent booleans. Represent native scan ownership in one coordinator with explicit state, for example:

```text
owner: none | watch | burst
phase: idle | starting | active | stopping | faulted
request/transition generation: monotonically increasing token
```

A watch start should return an explicit result or lease, such as:

```text
active | queuedBehindBurst | failed
```

Only an active lease permits `ScaleWatch` to subscribe.

## Diagnostics / field verification

The ownership coordinator introduced by this work should also be the authoritative source for a small read-only BLE diagnostics snapshot. Avoid exposing the current collection of implementation booleans as a quasi-public contract before the coordinator exists.

### Read-only BLE diagnostics endpoint

Add a narrow internal endpoint:

```http
GET /api/v1/diagnostics/ble
```

This endpoint is diagnostic surface, not a general device API. It should serialize the state required to distinguish scan/watch lifecycle failures from stale discovery/transport/GATT state, including at least:

- adapter state
- coordinator `owner`, `phase`, and transition generation
- native `UniversalBle.isScanning`
- latest native scan failure / failure count
- `ScaleWatch` armed state and generation
- whether watch device/failure subscriptions are installed
- preferred scale id and reconnect-policy decision (`shouldRetryPreferredScale`, power-mode block)
- current discovery cache entries and their connection states
- enough identity/instance information to detect a stale cached transport for the preferred scale
- advertisement counters/timestamps for the preferred scale, so we can distinguish “Android is advertising but Decaid ignores it” from “native scan is not producing results”

The endpoint must remain read-only. Existing mutation paths (`/api/v1/devices/scan`, machine state, adapter control outside Decaid) remain the way tests drive state.

### Tiny #598 diagnostics skin

Bundle a temporary/minimal **BLE Reconnect Diagnostics** skin in diagnostic builds. Keep it deliberately small (a single `index.html` with inline CSS/JS is sufficient) and use it as a guided field-reproduction harness rather than a product UI.

The skin should combine:

1. `/api/v1/info` for exact Decaid version/commit/build metadata.
2. `/ws/v1/devices` for device state and structured BLE errors.
3. `/ws/v1/machine/state` or the machine snapshot stream for sleep/wake transitions.
4. `/api/v1/diagnostics/ble`, sampled while a capture is active.
5. Explicit user markers for physical actions Decaid cannot observe directly (`Scale powered OFF`, `Scale powered ON`).

Provide guided controls for the #598 sequence:

- start/stop capture
- put machine to sleep
- marker: scale powered off
- marker: scale powered on
- wake machine
- discovery-only probe (`/api/v1/devices/scan?connect=false`)
- scan + connect probe (`/api/v1/devices/scan?connect=true`)
- snapshot-now

The UI should show the relevant state live (machine, scale, scan owner/phase, native scanning, watch armed/generation) and maintain one timestamped event timeline.

### Diagnostic report

The skin should export one JSON report containing:

- Decaid version/commit/build metadata
- capture start/end timestamps and duration
- machine/device WebSocket events
- diagnostics snapshots/transitions
- native scan failures
- explicit physical-action markers
- probe requests and their results

The goal is that a reporter can attach one JSON file to #598 without manually copying logs or running curl commands.

The report should make the main diagnostic split obvious:

- powered-on scale not rediscovered by discovery-only probe → scan/watch/native ownership path
- scale rediscovered but connection fails → stale transport/GATT path
- fresh preferred-scale advertisements observed while a stale cache entry suppresses replacement → stale discovery-cache path
- coordinator/watch says active while `UniversalBle.isScanning == false` → native scan/watch ownership path

The diagnostic skin is also the preferred hardware regression harness after the #451 fix: run the same guided #598 sequence before and after the coordinator change and compare reports.

## Acceptance criteria

- [ ] A full scan and `ScaleWatch` can never consume the same burst advertisement.
- [ ] Watch cancellation is complete before a full scan begins, or the full scan waits for cancellation.
- [ ] A queued watch request does not appear active to its consumer.
- [ ] Native stop failure prevents both GATT connection and subsequent scan startup.
- [ ] Native stop failure produces an observable fault/unavailable state.
- [ ] Initial start failure clears the matching watch request.
- [ ] Adapter recovery and post-burst resume cannot resurrect an orphan request.
- [ ] Disarm during initial start or restart leaves no subscriptions and no native watch.
- [ ] Hard power loss of a preferred scale while the DE1 sleeps can recover after wake without restarting Decaid.
- [ ] A fresh advertisement for a hard-power-cycled scale cannot be suppressed indefinitely by stale discovery/transport state.
- [ ] The coordinator exposes a read-only diagnostic snapshot sufficient to identify owner/phase/generation and native scan disagreement.
- [ ] `/api/v1/diagnostics/ble` exposes the diagnostic snapshot plus scale-watch, reconnect-policy, cache, advertisement, and latest scan-failure state without mutating BLE state.
- [ ] A minimal bundled BLE diagnostics skin can guide the #598 reproduction, timestamp physical scale power markers, run discovery-only vs scan+connect probes, and export a single JSON report.
- [ ] The exported report contains enough information to classify a failure as scan/watch lifecycle, stale discovery cache, or post-discovery transport/GATT.

## Required regression tests

1. Machine appears before scale during a full scan: only `EarlyConnectWatcher` handles the scale advertisement.
2. Scale appears before machine during a full scan: cached-device watch logic does not bypass post-scan policy.
3. Native `stopScan()` throws before a watch-driven connection: zero GATT attempts.
4. Native `stopScan()` throws during burst preemption: zero burst `startScan()` calls.
5. Initial watch start fails, then a fallback burst and adapter recovery occur: no latent watch starts.
6. Disarm while initial start is pending: no stale device/failure subscriptions.
7. Disarm while failed-connect restart is pending: no stale subscriptions or watch lease.
8. #598 sequence: connected preferred scale → machine sleep → scale hard power loss → scale power-on → machine wake → scale is rediscovered and reconnects without process restart.
9. Same #598 sequence with a missed/displaced disconnect callback: a fresh advertisement for the same device id still permits stale-state recovery rather than permanent de-duplication.
10. Diagnostics snapshot accurately reports coordinator owner/phase/generation and detects disagreement with native `isScanning`.
11. Diagnostics endpoint is read-only and reports stale preferred-scale cache identity plus fresh-advertisement counters when that condition is simulated.
12. Diagnostics-skin report fixture contains build metadata, ordered events, physical-action markers, probe results, and diagnostics snapshots required to classify the failure.

## References

- Follow-up to #447
- Related field report: #598
- Merged implementation: `143443fcb1f36c180226d13b9dcc34ed43d4b919`
- Reviewed PR head: `0e99c65ea0ad68466b32b0c6c987c3bdc6e61207`

tadelv:
## Summary

Harden background scale-watch ownership so BLE bursts and the persistent watch have one explicit native scan owner. Queued watch requests stay queued, ownership-transfer failures fault the coordinator before GATT work, and watch start/disarm continuations are generation-safe.

Added read-only BLE diagnostics at `/api/v1/diagnostics/ble` and a temporary `BLE Reconnect Diagnostics` skin for diagnostic builds.

## Linked Issue

Fixes #451

## Verification

- `dart format lib test`
- `flutter analyze`
- `flutter test` — all tests passed
- Focused ownership, stale-cache, fault-recovery, diagnostics endpoint, and bundled-skin tests passed
- `flutter test --dart-define=diagnostics=true test/diagnostics_skin_test.dart` passed
- Hardware retest not run in this environment

## Impact

- No database migration or breaking device API change.
- Adds the documented read-only REST diagnostics endpoint and updates `assets/api/rest_v1.yml`, `doc/Api.md`, and `doc/Skins.md`.
- The temporary diagnostic skin is bundled only with `--dart-define=diagnostics=true`.
- BLE cache replacement uses native link state to discard stale connected instances after a fresh advertisement, while preserving live links.
- Faulted BLE ownership resets only after a powered-off/powered-on adapter boundary.

## Contributor Responsibility


ODevStudio (changes requested):
DeepSeek worker: please treat this review as the implementation prompt. Fix the issues below on the current PR branch, add regression coverage for each one, and do not resolve this review by weakening invariants or tests.

## Goal

Keep the core ownership invariant from #451 intact: **before GATT work or a new BLE scan starts, the previous native scan must be known to be stopped.** At the same time, the coordinator must recover cleanly from expected lifecycle events instead of leaving Decaid permanently wedged.

## Required changes

### 1. Fix `ScaleWatch` state semantics around direct/cached connects and failed watch-driven connects

There are two related bugs in `lib/src/controllers/connection/scale_watch.dart`.

**Cached preferred-scale path:** `arm()` sets `_armed = true` before `_tryConnect(existing, gen)`. If that direct connect fails and `_tryConnect()` starts a lower-level watch, `_activate()` sees `_armed == true` and returns without installing `_sub` / `_failureSub`. The native watch can be active while `ScaleWatch` has no consumer subscriptions.

**Watch-driven sighting path:** `_onSighting()` stops the native watch and clears `_armed` before `_tryConnect()`. If the connect fails, `_tryConnect()` starts a replacement watch, but its stale/restart guard currently treats `!_armed` as a reason to cancel that replacement. This breaks the intended "failed connect -> keep watching" behavior.

Refactor these transitions so `_armed` means one thing only: **an active lower-level watch lease for which the data/failure subscriptions are installed.** A direct cached-device connect is not an armed watch. A stopped watch during a GATT attempt is not armed. Restart success must transition back to armed exactly once and install the subscriptions.

Do not fix this by simply deleting generation checks or by allowing stale continuations to subscribe. Preserve generation safety and `_shouldWatch()` checks after every awaited start/restart.

Add tests covering at least:
- cached preferred scale exists -> direct connect fails -> watch starts -> later preferred-scale advertisement triggers another connect;
- watch sighting -> watch stops -> connect fails -> replacement watch remains active and subscribed;
- disarm racing either restart still leaves no native watch and no subscriptions.

### 2. Handle a missed disconnect callback in stale BLE cache replacement

`UniversalBleDiscoveryService._deviceScanned()` currently keeps an existing cached device whenever its domain `connectionState` reports `connected` or `connecting`. That does not satisfy #451/#598's missed-disconnect case: after hard power loss, the stale object may continue reporting `connected`, causing every fresh advertisement for the same device id to be ignored indefinitely.

Use fresh advertisement evidence plus an authoritative/native transport/link check to decide whether the old instance is genuinely live. If the old instance is stale, evict it, cancel its connection subscription, dispose/disconnect it as appropriate, and allow the fresh advertisement to create a new transport/device instance.

Do **not** blindly replace genuinely connected devices on every advertisement. The fix must distinguish a live connection from a stale cached object.

Add a regression test for the exact scenario required by #451: existing preferred-scale instance still reports connected because the disconnect callback was missed, a fresh advertisement for the same id arrives after hard power-cycle, and discovery replaces/recreates the stale instance so reconnect can proceed without restarting the app.

### 3. Give the BLE ownership `faulted` state a safe recovery boundary

The new fail-stop behavior is correct while native ownership is unknown, but currently `BleScanPhase.faulted` appears permanent: later burst scans/watch starts reject immediately, and an adapter off/on transition does not clear the fault.

Keep the safety property: never start GATT or another scan while ownership is unknown. But define a recovery transition where ownership becomes knowable again. An adapter transition that conclusively tears down the native scanner is a reasonable boundary if supported by the platform semantics. After that boundary, clear/reset coordinator ownership state in a controlled way and permit a fresh scan/watch attempt.

Do not "recover" by clearing `faulted` immediately after a failed `stopScan()`; that would reintroduce the ownership race this PR is meant to fix.

Add tests for at least:
- stop/start ownership failure -> coordinator is faulted and new scan is blocked;
- adapter powered off -> powered on -> fault is safely reset -> explicit burst scan can start again;
- no orphan `_watchRequested` is resurrected during that recovery.

### 4. Fix the diagnostics build flag contract

Docs say `--dart-define=diagnostics=1`, but `WebUIStorage` uses `bool.fromEnvironment('diagnostics')`. Use one consistent contract. Prefer `--dart-define=diagnostics=true` with `bool.fromEnvironment`, unless there is an intentional repository-wide convention requiring `1`, in which case parse the string explicitly.

Add a test that verifies the **registration/gating behavior**, not only that the asset exists in `rootBundle`.

## Constraints / things not to do

- Do not weaken the native scan ownership invariant to make tests pass.
- Do not suppress `stopScan()` failures on ownership-transfer paths. Best-effort suppression is only appropriate for teardown/disposal where no subsequent GATT/scan depends on the stop succeeding.
- Do not use `_armed`, `_requested`, and `_lowerWatchRequested` interchangeably. Define their meanings and keep transitions consistent.
- Do not remove generation checks around asynchronous watch start/stop/restart continuations.
- Do not solve the stale-cache case by always replacing connected devices; preserve legitimate live connections.
- Do not make broad unrelated refactors in this PR.
- Do not close #451 acceptance gaps by changing tests to accept weaker behavior.

## Verification before marking this review resolved

Run and report:

```text
dart format lib test
flutter analyze
flutter test test/controllers/connection/scale_watch_test.dart
flutter test test/services/universal_ble_discovery_service_test.dart
flutter test test/ble_diagnostics_handler_test.dart
flutter test test/diagnostics_skin_test.dart
flutter test
```

Then manually re-read the final async transitions for these three races: disarm during start, burst preempting watch, and failed GATT followed by watch restart. Confirm that every successful watch activation has the expected subscriptions and every stale generation leaves no native watch/subscriptions behind.

Please keep the final implementation aligned with #451's acceptance criteria rather than only the currently existing tests.

ODevStudio (approved):
Re-reviewed current head `ffcd34d5c3eb544fb081f4a0ef89ba4c227fc2f9` against the prior requested changes and #451's core ownership/reconnect requirements.

The previously blocking common-path issues are addressed:
- `ScaleWatch` now treats `_armed` as an active watch lease rather than a direct-connect attempt, and failed cached/watch-driven connects correctly re-arm with subscriptions installed.
- stale cached BLE instances are checked against native link state before replacement, preserving a genuinely live connection while allowing hard-power-cycle recovery when the domain disconnect callback was missed.
- faulted scan ownership now has a controlled recovery boundary across adapter powered-off -> powered-on rather than being permanently wedged.
- the diagnostics build flag and registration test now use the same `diagnostics=true` contract.
- the added regression tests cover the relevant restart/disarm races and stale-cache identity behavior.

I did notice the existing `UniversalBleTransport` pre-connect hook still suppresses a thrown scan-stop error internally. That concern requires a native `stopScan()` failure path and is not a normal/common-path blocker under this re-review criterion, so I am not holding this PR for it.

Approved.

PR #775, opened 2026-09-03
feat(plugins): register sensor-backed devices

tadelv:
## Summary

What changed, and why?

- Add manifest-declared `sensor` drivers and a generation-owned `host.devices` registration bridge.
- Feed plugin-backed sensors through `DeviceController` and `SensorController` so the existing device, sensor REST, and sensor WebSocket APIs expose them without new routes.
- Prove the boundary with a fake localhost WebSocket humidity sensor, including `sampleNow`, lifecycle cleanup, stale-generation fencing, validation, timeouts, and resource limits.

## Linked Issue

N/A — implements the first sensor-registration slice of [Discussion #749: Generic device drivers through the JavaScript extension system](https://github.com/decentespresso/decaid/discussions/749).

## Verification

- `dart format lib test`
- `flutter test` — 3867 passed, 1 skipped
- `./scripts/fetch_dye2_plugin.sh`
- Parsed `assets/api/rest_v1.yml` with Ruby YAML
- Fake localhost WebSocket integration covers `/api/v1/devices`, `/api/v1/sensors`, `/api/v1/sensors/:id/execute`, and `/ws/v1/sensors/:id/snapshot`

## Impact

- Plugin manifests may now declare `drivers`; the only supported type is `sensor`.
- Plugins gain `host.devices.register()`, publication, disconnect reporting, and unregister operations. Driver declarations remain separate from transport permissions.
- Existing plugins remain compatible; no persistence migration or new REST/WebSocket route is introduced.
- Updated the REST schema and plugin, API, and device-management documentation.
- Grinder, BLE driver access, remembered plugin devices, probing, and generic device UI remain out of scope.

## Contributor Responsibility

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -6671,7 +6671,7 @@ components:
                 type: string
                 enum: ["quick", "full"]
             required: ["mode"]
-        resultSchema:
+        resultsSchema:
           type: object
           additionalProperties: true
           example:
@@ -7123,7 +7123,21 @@ components:
           type: array
           items:
             type: string
-            enum: [log, api, emit, pluginStorage, events.machine, events.shots, proxy.decent_api, proxy.decent_api.write]
+            enum: [log, api, emit, pluginStorage, events.machine, events.shots, proxy.decent_api, proxy.decent_api.write, network.websocket, network.tcp, network.tls]
+        drivers:
+          type: array
+          maxItems: 8
+          description: Device classes this plugin may register. Driver declarations do not grant transport permissions.
+          items:
+            type: object
+            required: [id, type]
+            properties:
+              id:
+                type: string
+                pattern: '^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$'
+              type:
+                type: string
+                enum: [sensor]
         settings:
           type: object
           description: >
--- a/doc/Api.md
+++ b/doc/Api.md
@@ -395,6 +395,12 @@ same request prevent all fields from being stored (validation is atomic).
 | GET | `/api/v1/sensors/:id` | Get sensor manifest | |
 | POST | `/api/v1/sensors/:id/execute` | Execute sensor command | |
 
+Plugin-backed sensors registered through `host.devices` use this same API and
+the device inventory. Their stable IDs have the form
+`plugin:<pluginId>:<driverId>:<instanceId>`. Sensor manifests expose command
+result schemas as `resultsSchema`. See `Plugins.md` for registration and
+lifecycle rules.
+
 #### `Bengle EBus Tap` sensor
 
 The Bengle EBus tap (USB interface `2` of a composite Bengle, VID `0x2e8a` /
--- a/doc/DeviceManagement.md
+++ b/doc/DeviceManagement.md
@@ -743,6 +743,15 @@ sensor is connected, skins can call the `measure` command through the existing
 Sensors API and read TDS, temperature, refractive index, and status values from
 the sensor data stream.
 
+`PluginDeviceService` is a `DeviceDiscoveryService` that contributes sensors
+registered by plugin generations. This keeps plugin-backed sensors on the same
+`DeviceController` → `SensorController` path as native sensors. Public identity
+comes from plugin id, declared driver id, and plugin-local instance id; unload
+removes the retiring generation without changing that identity for a later
+reload. Plugin connection handlers must complete protocol initialization before
+the sensor reports `connected`. Registrations are runtime-only and are not added
+to remembered-device selection.
+
 ### Bengle EBus tap
 
 Bengle composite devices (VID `0x2e8a`, PID `0x000a`) may expose a second CDC
--- a/doc/Plugins.md
+++ b/doc/Plugins.md
@@ -582,16 +582,21 @@ handle with:
 
 All three handlers are required. Decaid calls `connect()` when the sensor joins
 the sensor registry; resolving means the driver is ready to serve commands, not
-merely that its transport opened. `disconnect()` performs normal driver cleanup.
-`execute({ commandId, params })` handles a declared command and returns an object.
-Handler failures propagate through the existing sensor command API. Calls time
-out after 10 seconds.
+merely that its transport opened. `disconnect()` performs normal driver cleanup
+and is run before a device is removed: on explicit `unregister()` and on plugin
+unload, so the driver can release its transport or other resources. A failing or
+timed-out `disconnect()` still removes the device and rejects in-flight
+commands; the failure is surfaced to the caller.
+`execute({ commandId, params })` handles a declared command and returns an
+object. Handler failures propagate through the existing sensor command API.
+Calls time out after 10 seconds.
 
 Definitions, snapshots, command parameters and command results are limited to
 64 KiB of JSON. A plugin generation can register at most 8 devices. Registration
-is not remembered across app restarts. On plugin unload, Decaid removes every
-device and rejects in-flight commands owned by the retiring generation, even if
-`onUnload()` fails. Late publications and command results from older generations
+is not remembered across app restarts. On plugin unload, Decaid runs each
+device's `disconnect()` handler, removes every device, and rejects in-flight
+commands owned by the retiring generation, even if `onUnload()` fails. Late
+publications and command results from older generations
 are ignored. BLE-backed drivers, discovery, probing and grinder registration are
 not supported by this first sensor registration contract.
 
--- a/doc/Plugins.md
+++ b/doc/Plugins.md
@@ -32,6 +32,12 @@ A Decaid plugin consists of two required files:
     "pluginStorage",
     "events.machine"
   ],
+  "drivers": [
+    {
+      "id": "humidity",
+      "type": "sensor"
+    }
+  ],
   "settings": {
     "SettingName": {
       "type": "string",
@@ -77,6 +83,7 @@ A Decaid plugin consists of two required files:
   - `network.websocket`: Open outbound WebSocket connections (`ws://` and `wss://`) through `host.transport`
   - `network.tcp`: Open outbound raw TCP connections through `host.transport`
   - `network.tls`: Open outbound TLS connections (platform trust store) through `host.transport`
+- **drivers**: Device classes the plugin may register. Each declaration has a plugin-local `id` and a `type`. The only currently supported type is `sensor`. A manifest may declare at most 8 drivers. Driver declarations authorize registration; they do not grant transport access. For example, a WebSocket-backed sensor also needs `network.websocket`.
 - **settings**: User-configurable options with `type` (`string`, `number`, `boolean`, `enum`), an optional `label` giving the setting a human-friendly name, an optional `description` explaining what the setting does, an optional `default`, and an optional `secure` flag for credentials such as passwords. Enum `values` are a JSON array of strings. Secure values use platform credential storage, are supplied in memory to `onLoad(settings)`, and are never returned by the REST API.
 
   `GET /api/v1/plugins` returns this schema verbatim under `settings`, so a skin can render a settings form — labels, help text and defaults included — without reading the plugin's repository. `GET /api/v1/plugins/:id/settings` returns the stored values only.
@@ -488,6 +495,106 @@ host.transport.onEvent(opened.handle, (event) => {
 await host.transport.send(opened.handle, { type: "binary", data: btoa("\x00\x01\x02") });
 ```
 
+## Device Registration (`host.devices`)
+
+A plugin with a declared sensor driver can register runtime sensor instances.
+Registered sensors join Decaid's normal device inventory and sensor registry; no
+plugin-specific endpoint is created. They are available through:
+
+- `GET /api/v1/devices`
+- `GET /api/v1/sensors`
+- `GET /api/v1/sensors/:id`
+- `POST /api/v1/sensors/:id/execute`
+- `/ws/v1/sensors/:id/snapshot`
+
+Registration and transport authorization are independent. The manifest below
+allows a sensor registration and an outbound WebSocket connection:
+
+```json
+{
+  "drivers": [{ "id": "humidity", "type": "sensor" }],
+  "permissions": ["network.websocket"]
+}
+```
+
+Register the sensor from `onLoad()`:
+
+```js
+let sensor;
+let transportHandle;
+
+async function registerSensor() {
+  sensor = await host.devices.register({
+    driverId: "humidity",
+    instanceId: "office",
+    name: "Office humidity",
+    vendor: "Example",
+    dataChannels: [
+      { key: "relativeHumidity", type: "number", unit: "%RH" }
+    ],
+    commands: [
+      {
+        id: "sampleNow",
+        name: "Sample now",
+        paramsSchema: { type: "object" },
+        resultsSchema: {
+          type: "object",
+          properties: { relativeHumidity: { type: "number" } }
+        }
+      }
+    ]
+  }, {
+    async connect() {
+      const opened = await host.transport.open({
+        kind: "websocket",
+        url: "ws://sensor.local/readings"
+      });
+      transportHandle = opened.handle;
+    },
+    async disconnect() {
+      if (transportHandle) await host.transport.close(transportHandle);
+    },
+    async execute(command) {
+      if (command.commandId === "sampleNow") {
+        return { relativeHumidity: 52.4 };
+      }
+      throw new Error("unknown command");
+    }
+  });
+
+  await sensor.publish({ relativeHumidity: 52.4 });
+}
+```
+
+Call `registerSensor()` from the plugin's `onLoad()` method.
+
+`host.devices.register(definition, handlers)` returns a Promise for a device
+handle with:
+
+- `deviceId`: stable public identity derived from plugin id, driver id, and
+  instance id; plugin generation is not part of the identity.
+- `publish(snapshot)`: publishes one complete snapshot. Every declared channel
+  must be present, values must match their declared JSON type, and undeclared
+  channels are rejected.
+- `reportDisconnected()`: marks the sensor disconnected after an unexpected
+  transport or protocol failure.
+- `unregister()`: removes the sensor from the device inventory.
+
+All three handlers are required. Decaid calls `connect()` when the sensor joins
+the sensor registry; resolving means the driver is ready to serve commands, not
+merely that its transport opened. `disconnect()` performs normal driver cleanup.
+`execute({ commandId, params })` handles a declared command and returns an object.
+Handler failures propagate through the existing sensor command API. Calls time
+out after 10 seconds.
+
+Definitions, snapshots, command parameters and command results are limited to
+64 KiB of JSON. A plugin generation can register at most 8 devices. Registration
+is not remembered across app restarts. On plugin unload, Decaid removes every
+device and rejects in-flight commands owned by the retiring generation, even if
+`onUnload()` fails. Late publications and command results from older generations
+are ignored. BLE-backed drivers, discovery, probing and grinder registration are
+not supported by this first sensor registration contract.
+
 ## Plugin Lifecycle
 
 1. **Initialization**: Plugin directory is copied to app storage
@@ -635,7 +742,7 @@ it in Decaid UI.
 - **Global Functions**: `fetch()`, `btoa()`, `setTimeout()`, `clearTimeout()`
 - **Objects**: `Promise`, `JSON`, `Math`, `Date`, `Array`, `Object`
 - **Constants**: `undefined`, `null`, `Infinity`, `NaN`
-- **Host API**: `host.log()`, `host.emit()`, `host.storage()`, `host.decentProxy()`, `host.transport`
+- **Host API**: `host.log()`, `host.emit()`, `host.storage()`, `host.decentProxy()`, `host.transport`, `host.devices`
 
 ### Not Available
 
--- a/doc/Plugins.md
+++ b/doc/Plugins.md
@@ -544,8 +544,8 @@ async function registerSensor() {
       }
     ]
   }, {
-    async connect() {
-      const opened = await host.transport.open({
+    async connect(transport) {
+      const opened = await transport.open({
         kind: "websocket",
         url: "ws://sensor.local/readings"
       });
@@ -580,9 +580,12 @@ handle with:
   transport or protocol failure.
 - `unregister()`: removes the sensor from the device inventory.
 
-All three handlers are required. Decaid calls `connect()` when the sensor joins
-the sensor registry; resolving means the driver is ready to serve commands, not
-merely that its transport opened. `disconnect()` performs normal driver cleanup
+All three handlers are required. Decaid calls `connect(transport)` when the
+sensor joins the sensor registry. The invocation-bound `transport` has the same
+methods as `host.transport`; connections opened through it belong to that
+connect attempt across Promise continuations. Resolving means the driver is
+ready to serve commands, not merely that its transport opened. `disconnect()`
+performs normal driver cleanup
 and is run before a device is removed: on explicit `unregister()` and on plugin
 unload, so the driver can release its transport or other resources. A failing or
 timed-out `disconnect()` still removes the device and rejects in-flight
--- a/doc/plans/archive/plugin-backed-sensor-driver/design.md
+++ b/doc/plans/archive/plugin-backed-sensor-driver/design.md
@@ -1,0 +1,57 @@
+# Plugin-backed sensor driver
+
+**Date:** 2026-09-03
+**Discussion:** [#749](https://github.com/decentespresso/decaid/discussions/749)
+
+## Goal
+
+Prove the plugin device boundary with a humidity sensor whose protocol runs in JavaScript over the existing permission-gated WebSocket transport.
+
+## Decisions
+
+- Add static `drivers` contributions to plugin manifests; they are not permissions.
+- The first supported driver type is `sensor`.
+- A plugin may register only instances of drivers declared by its manifest.
+- Device identity is stable across reloads and derived from plugin id, driver id, and plugin-local instance id; generation is ownership metadata, not identity.
+- `ConnectionState.connected` means the plugin's connect handler completed protocol initialization and the sensor is ready.
+- Decaid invokes plugin connect, disconnect, and command handlers. Plugins own protocol framing, handshakes, and transport use.
+- Registered devices enter `DeviceController` through a plugin-backed `DeviceDiscoveryService`, so existing device inventory and Sensor APIs see them without new REST or WebSocket routes.
+- Registration definitions are immutable. State publications are complete, bounded snapshots.
+- Registrations, callbacks, and pending commands are owned by plugin id plus generation and are removed or rejected during unload even when `onUnload()` fails.
+
+## Public seams
+
+1. `PluginManifest.fromJson()` validates `drivers`.
+2. `host.devices.register(definition, handlers)` registers a declared sensor.
+3. The returned device publishes snapshots and reports unsolicited disconnects.
+4. Existing `/api/v1/devices`, `/api/v1/sensors`, `/api/v1/sensors/{id}/execute`, and `/ws/v1/sensors/{id}/snapshot` expose the device.
+5. Plugin unload/reload/disposal removes stale registrations and closes associated transports.
+
+## Tracer fixture
+
+A test-local WebSocket server:
+
+- sends `{"relativeHumidity":52.4}`;
+- accepts `{"command":"sampleNow"}`;
+- replies with a fresh humidity sample.
+
+The test plugin declares `network.websocket` plus one `sensor` driver, connects through `host.transport`, registers the sensor, publishes readings, and translates `sampleNow` commands.
+
+## Implementation order
+
+1. Manifest driver parsing and rejection tests.
+2. Plugin device service and DeviceController inventory test.
+3. `host.devices` registration/publication/command tests.
+4. End-to-end fake-WebSocket humidity sensor test through existing Sensor APIs.
+5. Generation, unload, malformed input, timeout, and resource-bound tests.
+6. Update `doc/Plugins.md`, `doc/DeviceManagement.md`, and correct affected REST schema mismatches.
+
+## Out of scope
+
+- Grinder semantics or `DeviceType.grinder`.
+- BLE scanning, matching, GATT access, or probing.
+- Migrating existing native devices.
+- Remembered/preferred plugin devices.
+- Generic device settings/debug UI.
+- New generic REST or WebSocket endpoints.
+- Shared JavaScript runtime isolation hardening.
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -39,6 +39,7 @@ import 'package:reaprime/src/controllers/workflow_device_sync.dart';
 import 'package:reaprime/src/models/data/workflow.dart';
 import 'package:reaprime/src/models/device/device.dart';
 import 'package:reaprime/src/models/device/simulated_device.dart';
+import 'package:reaprime/src/plugins/plugin_device_service.dart';
 import 'package:reaprime/src/plugins/plugin_loader_service.dart';
 import 'package:reaprime/src/plugins/plugin_source_service.dart';
 import 'package:reaprime/src/services/android_updater.dart';
@@ -278,6 +279,8 @@ void main(List<String> args) async {
   });
 
   final List<DeviceDiscoveryService> services = [];
+  final pluginDeviceService = PluginDeviceService();
+  services.add(pluginDeviceService);
 
   final bleDiscoveryService = UniversalBleDiscoveryService();
   if (!cliArgs.serial) {
@@ -503,6 +506,7 @@ void main(List<String> args) async {
     kvStore: HiveStoreService(defaultNamespace: "plugins")..initialize(),
     decentProxyService: decentProxyService,
     credentialStore: credentialStore,
+    deviceService: pluginDeviceService,
   );
   await pluginService.pluginManager.attachDe1Controller(de1Controller);
   persistenceController.onShotStored = (shotId) =>
--- a/lib/src/models/device/device_implementation.dart
+++ b/lib/src/models/device/device_implementation.dart
@@ -21,4 +21,5 @@ enum DeviceImplementation {
   debugPort,
   sensorBasket,
   bengleDebugPort,
+  plugin,
 }
--- a/lib/src/plugins/plugin_device_service.dart
+++ b/lib/src/plugins/plugin_device_service.dart
@@ -257,7 +257,9 @@ class _PluginSensor implements Sensor {
       StreamController.broadcast();
   Future<void>? _connecting;
   Future<void>? _disconnectCleanup;
+  int? _disconnectCleanupEpoch;
   int _connectionEpoch = 0;
+  int _connectionFence = 0;
   bool _disposed = false;
 
   @override
@@ -289,19 +291,38 @@ class _PluginSensor implements Sensor {
     if (_disposed || _connectionState.value == ConnectionState.connected) {
       return Future.value();
     }
-    return _connecting ??= _connect().whenComplete(() => _connecting = null);
+    return _connecting ??= _connectAfterCleanup().whenComplete(
+      () => _connecting = null,
+    );
+  }
+
+  Future<void> _connectAfterCleanup() async {
+    final cleanup = _disconnectCleanup;
+    final cleanupFence = _connectionFence;
+    if (cleanup != null) {
+      try {
+        await cleanup;
+      } catch (_) {}
+      if (cleanupFence != _connectionFence) return;
+      if (identical(_disconnectCleanup, cleanup)) {
+        _disconnectCleanup = null;
+        _disconnectCleanupEpoch = null;
+      }
+    }
+    await _connect();
   }
 
   Future<void> _connect() async {
-    final epoch = ++_connectionEpoch;
+    ++_connectionEpoch;
+    final fence = ++_connectionFence;
     _connectionState.add(ConnectionState.connecting);
     try {
       await _invoke(PluginDeviceOperation.connect, const {});
-      if (!_disposed && epoch == _connectionEpoch) {
+      if (!_disposed && fence == _connectionFence) {
         _connectionState.add(ConnectionState.connected);
       }
     } catch (_) {
-      if (!_disposed && epoch == _connectionEpoch) {
+      if (!_disposed && fence == _connectionFence) {
         _connectionState.add(ConnectionState.disconnected);
       }
       rethrow;
@@ -311,18 +332,35 @@ class _PluginSensor implements Sensor {
   @override
   Future<void> disconnect() {
     if (_disposed) return Future.value();
-    return _disconnectCleanup ??= _runDisconnectCleanup();
+    final epoch = _connectionEpoch;
+    final existing = _disconnectCleanup;
+    if (existing != null && _disconnectCleanupEpoch == epoch) {
+      if (_connecting != null) _connectionFence += 1;
+      return existing;
+    }
+    final cleanup = _runDisconnectCleanup(epoch);
+    _disconnectCleanup = cleanup;
+    _disconnectCleanupEpoch = epoch;
+    return cleanup;
   }
 
-  Future<void> _runDisconnectCleanup() async {
-    _connectionEpoch += 1;
+  Future<void> _runDisconnectCleanup(int epoch) async {
+    final connecting = _connecting;
+    _connectionFence += 1;
     if (_connectionState.value != ConnectionState.disconnected) {
       _connectionState.add(ConnectionState.disconnecting);
+    }
+    if (connecting != null) {
+      try {
+        await connecting;
+      } catch (_) {}
     }
     try {
       await _invoke(PluginDeviceOperation.disconnect, const {});
     } finally {
-      if (!_disposed) _connectionState.add(ConnectionState.disconnected);
+      if (!_disposed && epoch == _connectionEpoch) {
+        _connectionState.add(ConnectionState.disconnected);
+      }
     }
   }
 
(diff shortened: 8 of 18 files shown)


ODevStudio (changes requested):
Requesting changes for two blockers. CI is green, but these are runtime lifecycle/inventory regressions on normal paths.

### Blocker 1: a disconnected registered plugin sensor breaks the scan contract

`PluginDeviceService` publishes `List.unmodifiable(_registrations.values)` and implements `scanForDevices()` as a no-op. `DeviceController._runScan()` intentionally removes disconnected/stale devices from each discovery service's current list before invoking that service's scan. After a plugin calls `reportDisconnected()` (or the sensor is explicitly disconnected), the next normal device scan attempts `remove()` on the unmodifiable list and can throw `UnsupportedError`. If the list is only made mutable, the registration is then removed from `DeviceController` and never re-emitted because `PluginDeviceService.scanForDevices()` does nothing.

A runtime registration should remain discoverable until `unregister()`/plugin unload; scanning must not crash or permanently lose it.

### Blocker 2: `unregister()` / unload bypass the required `disconnect()` handler

`PluginDeviceService.unregister()`, `removeAllForPlugin()`, and `dispose()` call `_PluginSensor.dispose()` directly; `dispose()` only closes Dart streams and never invokes `PluginDeviceOperation.disconnect`. This breaks the documented handler lifecycle. In the documented WebSocket pattern, `disconnect()` owns `host.transport.close(...)`, so `device.unregister()` can remove the sensor while leaving its transport open. Plugin unload also clears device handlers before removing registrations, so the required disconnect handler cannot run there either.

## Agent worker prompt

Fix PR #775 only for the two blockers above. Keep the existing public API, stable device IDs, limits/timeouts, permission model, REST/WebSocket routes, and generation fencing. Do not add BLE/grinder/discovery persistence or unrelated refactors.

1. **Make `PluginDeviceService` obey `DeviceController`'s scan/discovery contract.**
   - Change `_publishDevices()` to emit a fresh **mutable snapshot** of the current registrations (for example `List<Device>.of(_registrations.values)`), not an unmodifiable list. `DeviceController` deliberately prunes the per-service list before a scan, so this snapshot must tolerate that mutation without mutating `_registrations` itself.
   - Implement `PluginDeviceService.scanForDevices()` so an active service re-publishes a fresh snapshot of all current registrations. Do not unregister or recreate registrations during a scan.
   - Add a regression test through `DeviceController`: register/connect a plugin sensor, call `reportDisconnected()`, wait for the disconnected state, run `deviceController.scanForDevices()`, and assert the scan completes without throwing and the same registered device ID remains present in the resulting/current inventory (and remains connectable through the existing sensor path).

2. **Run device `disconnect()` cleanup before a registration is torn down.**
   - For direct `device.unregister()`, invoke the registered plugin `disconnect` handler before disposing the sensor streams/removing its handler. Always remove the registration and publish the updated inventory even if disconnect fails or times out; surface the cleanup failure after removal rather than leaving a half-registered device.
   - If native unregister can reject after removal, update the JS `unregister()` wrapper so `__deviceRemoveHandlers(registrationHandle)` runs on both fulfillment and rejection, then rethrow the rejection. Do not leave a dead handler entry behind.
   - On plugin unload, run each retiring generation's disconnect handler **before** `__clearDeviceHandlersForPlugin(...)` and before `_transportService.closeAllForPlugin(...)`. Cleanup must still continue if `onUnload()` or a disconnect handler fails.
   - Be careful with the existing generation/liveness guards: `_unloadPlugin()` marks the runtime stopping and increments `_pluginGenerations` before cleanup, and `_handleDeviceMessage()` currently drops old-generation invocation results. Simply calling `sensor.disconnect()` from `removeAllForPlugin()` will therefore hang/fail unless you add a narrowly scoped cleanup path. Allow only the retiring generation's pending **disconnect** invocation/result to complete (or implement an equivalent cleanup-only mechanism). Do **not** reopen normal `register`, `publish`, `reportDisconnected`, `unregister`, or `execute` traffic for stale generations.
   - Preserve the existing 10-second device invocation timeout. Reject pre-existing in-flight commands as today; then perform disconnect cleanup; after disconnects settle/fail/timeout, clear handlers and close remaining transports. Cleanup failures must not prevent inventory removal or transport teardown.
   - Add tests in the plugin device/manager test suite proving: (a) direct `unregister()` calls `disconnect` exactly once and removes the device, and (b) plugin unload still calls `disconnect` and removes the device even when `onUnload()` throws. Keep the stale-generation and pending-command fencing tests passing.

Run `dart format lib test`, `flutter analyze`, the affected plugin/device tests, and the full `flutter test` suite.

ODevStudio (changes requested):
Re-review: the two previously reported blockers are fixed. The mutable discovery snapshot/rescan path is correct, and direct unregister + retiring-generation device disconnect invocation are now covered by focused tests.

One blocker remains on the ordinary documented WebSocket lifecycle:

### Blocker: unload-time `disconnect()` cannot close its transport

During `_unloadPlugin()`, the runtime is marked stopping and `_pluginGenerations[id]` is incremented before `PluginDeviceService.removeAllForPlugin()` invokes each retiring sensor's `disconnect()` handler. The new `_deviceDisconnectCleanup` exception correctly lets the retiring generation's **device disconnect invocation/result** pass.

However, if that handler performs the documented cleanup pattern:

```js
disconnect() {
  return host.transport.close(transportHandle);
}
```

`host.transport.close()` sends a transport message with the retiring generation. `_handleTransportMessage()` still requires `generation == _pluginGenerations[pluginId]` and `_plugins[pluginId]?.isAlive == true`, so it replies `Plugin generation changed`. The disconnect handler therefore rejects during normal plugin unload. `closeAllForPlugin()` subsequently closes the native transport, but that does not satisfy the handler contract and can make `unloadPlugin()` surface a cleanup failure for a correctly written plugin.

The current unload test only increments `disconnectCalls`, so it does not exercise the transport operation that `doc/Plugins.md` explicitly uses in `disconnect()`.

## Agent worker prompt

Fix only this remaining unload cleanup blocker in PR #775. Keep the prior scan fix, direct unregister behavior, generation fencing, permissions, limits/timeouts, and manager-disposal behavior unchanged. Do not broaden stale-generation access generally.

1. **Permit the retiring device disconnect handler to close its already-owned transport during active plugin unload.**
   - In `PluginManager._handleTransportMessage`, add a narrowly scoped cleanup exception for a transport `close` request when all of these are true (or equivalent invariants are enforced):
     - manager lifecycle is `active`;
     - `(pluginId, generation)` is currently in `_deviceDisconnectCleanup`;
     - `generation` is the plugin's retiring generation / runtime is stopping;
     - the transport operation is exactly `close`.
   - The current code reads `type` only after the generation/liveness guard; restructure as needed so the cleanup-close decision can be made before rejecting the message.
   - Do **not** allow retiring generations to `open` new transports, publish/register devices, execute commands, or otherwise bypass the existing stale-generation fences. A `close` on an existing handle is sufficient for the documented lifecycle case. `PluginTransportService.close(pluginId, generation, handle)` already enforces ownership of the handle; preserve that check and the existing transport permission checks.
   - Ensure the transport reply is delivered normally so the JavaScript `host.transport.close(...)` Promise resolves/rejects and the device `disconnect()` invocation can complete before handlers are cleared and `closeAllForPlugin()` runs.

2. **Add a regression test that exercises the real documented cleanup path.**
   - Load a plugin with a sensor driver and `network.websocket` permission.
   - Have `connect()` open a test/fake WebSocket transport and retain its handle.
   - Have `disconnect()` return/await `host.transport.close(handle)` and record that it completed (not merely that the handler was entered).
   - Call `manager.unloadPlugin(...)` while the manager is active.
   - Assert unload completes without a `Plugin generation changed` cleanup error, the disconnect completion marker is set exactly once, the plugin device is removed, and `manager.liveTransportCount == 0`.
   - Keep the existing test that `onUnload()` throwing still does not prevent device disconnect/removal, and keep stale-generation/pending-command tests passing.

3. Run `dart format lib test`, `flutter analyze`, the affected plugin device/transport tests, and the full `flutter test` suite.

ODevStudio (changes requested):
Requesting changes for two remaining lifecycle blockers. The previous scan-contract and unload-time transport-close fixes look good; keep those behaviors intact.

### Blocker 1: `reportDisconnected()` causes unregister/unload to skip the required `disconnect()` cleanup

`_PluginSensor.disconnect()` currently returns immediately when `connectionState == disconnected`. But `reportDisconnected()` sets that state after an unexpected transport/protocol failure, and a failed `connect()` also leaves the sensor disconnected. Both `PluginDeviceService.unregister()` and `removeAllForPlugin()` rely on `sensor.disconnect()` to run the plugin's cleanup handler.

As a result, this normal sequence is broken:

```text
connect -> reportDisconnected -> unregister
```

The device is removed, but the plugin `disconnect()` handler is never invoked. For direct `unregister()` this can leak a plugin-owned WebSocket/TCP/TLS transport because there is no plugin-wide `closeAllForPlugin()` sweep afterward. It also contradicts the documented contract that `disconnect()` runs before explicit unregister and plugin unload.

The existing service test already does `reportDisconnected()` followed by `unregister()`, but it only checks inventory removal; it does not assert that the disconnect operation ran.

### Blocker 2: direct `device.unregister()` does not reject/fence in-flight commands for that registration

Pending device invocations are tracked only by plugin id, generation, and operation. They do not retain the `registrationHandle`. The unregister path calls `deviceService.unregister()` and removes the handler, but it does not reject pending `execute()` invocations for that device.

Therefore a command Promise can remain pending after the device has been removed until the 10-second timeout. If the JavaScript handler resolves after unregister, its already-attached Promise callback can still send an `invocationResult`; because the plugin generation is still current, Dart can accept that late result and complete a command successfully for a sensor that no longer exists.

Device teardown must terminate that registration's outstanding command work and ignore late results, without rejecting unrelated devices belonging to the same plugin generation.

## Agent worker prompt

Fix PR #775 only for the two blockers above. Preserve the current public API, stable device IDs, manifest permissions, resource limits, 10-second invocation timeout, existing REST/WebSocket routes, mutable discovery snapshot/rescan behavior, retiring-generation fencing, unload-time transport-close exception, and manager-disposal behavior. Do not add BLE/grinder support, persistence, probing, generic discovery, or unrelated refactors.

### 1. Make teardown cleanup independent of the public connection-state value

The plugin `disconnect()` handler is a resource-cleanup callback, not merely a state transition. `ConnectionState.disconnected` must not by itself mean cleanup has already executed.

Implement lifecycle state that can distinguish at least:

- cleanup has never run;
- cleanup is currently in flight;
- cleanup has completed.

Requirements:

- `reportDisconnected()` may mark the sensor disconnected but must **not** mark plugin cleanup complete.
- A failed `connect()` may leave the sensor disconnected but must **not** suppress later cleanup.
- Explicit `device.unregister()` must invoke the plugin `disconnect()` handler exactly once before final disposal/removal, even if the sensor was already marked disconnected by `reportDisconnected()` or a failed connect.
- Plugin unload must likewise invoke the handler exactly once for every retiring registration whose cleanup has not already completed.
- Repeated/concurrent teardown paths must not double-invoke `disconnect()`; share an in-flight cleanup Future or equivalent.
- Preserve existing behavior that a disconnect failure/timeout does not leave the registration in inventory. Direct unregister should still surface the cleanup failure after removal; unload should continue removing devices and closing remaining transports while retaining the first cleanup error according to the current cleanup policy.
- Do not change normal reconnect semantics: a reported disconnect remains discoverable/reconnectable until unregister/unload.

Add focused regression coverage for both direct unregister and unload. At minimum:

1. Register and connect a sensor.
2. Call `reportDisconnected()` and wait for `ConnectionState.disconnected`.
3. Call `device.unregister()`.
4. Assert the plugin `disconnect()` handler ran exactly once and the device was removed.

Also add a transport-backed variant (WebSocket is fine) where `disconnect()` closes the retained transport, then assert direct unregister leaves `liveTransportCount == 0`. This specifically proves the leak is fixed, not only that a counter incremented.

Add/keep a case for failed `connect()` followed by teardown if practical; teardown must still get one cleanup attempt.

### 2. Reject and fence pending invocations by registration on direct unregister

Extend `_PendingDeviceInvocation` (or an equivalent structure) to retain the `registrationHandle` that owns the invocation. The registration identity must participate in teardown/fencing.

Requirements:

- Every invocation created by `_invokeDeviceHandler(...)` records the registration handle.
- Before or as direct unregister tears down that registration, reject all pre-existing pending invocations owned by `(pluginId, generation, registrationHandle)` with a deterministic `PluginDeviceException` such as `Plugin device unregistered`.
- Do **not** reject pending invocations for another device registered by the same plugin generation.
- The unregister cleanup invocation itself must still be allowed to run to completion; do not accidentally reject the `disconnect` operation you are starting for teardown.
- After unregister starts/removes the registration, any late `invocationResult` for an earlier command from that registration must be ignored and must not complete the already-rejected Dart Future.
- Keep generation-wide rejection during plugin unload unchanged in semantics.
- Preserve the 10-second timeout for work that remains valid and pending.

A clean implementation could add a registration-scoped rejection helper and ensure `_completeDeviceInvocation()` only accepts results for invocations that are still present. Do not broaden stale-generation access or weaken the existing cleanup-only exceptions.

Add regression tests that prove:

1. A device command returns a never-resolving Promise.
2. Start that command and then call the same device's `unregister()`.
3. The command Future rejects promptly because the device was unregistered, rather than waiting for the invocation timeout.
4. The unregister operation completes and the device disappears from `SensorController` / `DeviceController` inventory.
5. A second device registered by the same plugin can have an in-flight command that is **not** rejected by unregistering the first device.
6. Add a late-result test: arrange for an execute handler to resolve after unregister and verify the already-rejected command does not later complete successfully or otherwise alter state.

### 3. Verification

Run and report:

- `dart format lib test`
- `flutter analyze`
- affected plugin-device / plugin-manager / transport tests
- full `flutter test`

Keep all existing regression tests from the earlier review fixes passing, especially:

- disconnected registration survives a normal device scan and remains reconnectable;
- direct unregister runs disconnect and removes the device even when cleanup fails;
- plugin unload runs retiring-generation disconnect even if `onUnload()` throws;
- unload-time `disconnect()` can `await host.transport.close(handle)` successfully;
- stale-generation register/publish/reportDisconnected/unregister/execute/open/send traffic remains fenced;
- pre-existing plugin-generation commands are rejected during unload;
- manager disposal does not wait on JavaScript cleanup that cannot complete.

ODevStudio (changes requested):
Requesting changes for two remaining lifecycle blockers. The earlier scan, unload cleanup, transport-close, and registration-scoped pending-command fixes look good; keep those behaviors intact.

### Blocker 1: disconnect cleanup is one-shot for the lifetime of the registration

`_PluginSensor.disconnect()` memoizes `_disconnectCleanup` with `??=` and never resets it after a later successful `onConnect()`. That breaks the ordinary device lifecycle exposed through the existing `/api/v1/devices/connect` and `/api/v1/devices/disconnect` paths.

A normal sequence such as:

```text
connect -> disconnect -> connect -> disconnect
```

runs the plugin `disconnect()` handler only on the first disconnect. The second call returns the old completed cleanup Future, does not run the handler again, and can leave resources opened by the second `connect()` alive. The same problem applies if the device reconnects and is then unregistered or the plugin is unloaded: teardown can reuse stale cleanup state from the previous connection instead of cleaning up the current one.

### Blocker 2: disconnect/unregister can race an in-flight connect and leak its late side effects

`_runDisconnectCleanup()` starts the plugin `disconnect` handler without serializing against `_connecting`. Incrementing `_connectionEpoch` prevents a late connect from changing the Dart connection state, but it does not cancel the JavaScript `connect()` Promise or its transport operations.

For direct unregister, registration-scoped rejection rejects the pending Dart `connect` invocation, but that also does not cancel the already-running JS handler. A slow `connect()` can therefore finish after disconnect cleanup has completed, successfully open a WebSocket/TCP/TLS handle, have its late invocation result ignored, and leave an orphan transport. Direct unregister has no plugin-wide `closeAllForPlugin()` sweep to clean that up.

## Agent worker prompt

Fix PR #775 only for the two lifecycle blockers above. Preserve the current public API, stable device IDs, manifest permissions, resource limits, 10-second invocation timeout, REST/WebSocket routes, mutable discovery snapshot/rescan behavior, registration-scoped command fencing, retiring-generation fencing, unload-time transport-close exception, and manager-disposal behavior. Do not add BLE/grinder support, persistence, probing, generic discovery, or unrelated refactors.

### 1. Make disconnect cleanup per connection lifecycle, not per registration

Treat cleanup as belonging to the currently established/attempted connection epoch rather than permanently caching one Future for the whole `_PluginSensor` object.

Requirements:

- Concurrent/repeated calls to `disconnect()` for the same connection attempt must share one in-flight cleanup Future and invoke the plugin handler at most once for that epoch.
- After a later successful reconnect, a future `disconnect()` must be allowed to invoke the plugin `disconnect` handler again for that new connection.
- `reportDisconnected()` must not by itself mark cleanup complete; unregister/unload must still clean up that connection if cleanup has not run.
- A failed `connect()` must still permit one teardown cleanup attempt.
- Unregister/unload must clean up the most recent connection epoch exactly once before disposal/removal.
- Preserve current reconnect behavior: a reported disconnect stays registered/discoverable and can reconnect.
- Preserve failure semantics: cleanup failure/timeout must not leave the registration in inventory; direct unregister surfaces the cleanup error after removal, while unload continues teardown and retains the first error according to the current policy.

A reasonable implementation is to track a monotonically increasing connection/cleanup epoch plus the cleanup Future and the epoch it belongs to. Reset/advance cleanup eligibility when a new connect attempt starts (or when it successfully establishes a new resource-owning connection), while still deduplicating concurrent teardown for the same epoch.

Add a transport-backed regression test covering:

1. register/connect a sensor whose `connect()` opens a fake WebSocket and whose `disconnect()` closes its retained handle;
2. disconnect it through the normal device method/API and assert `liveTransportCount == 0` and disconnect count is 1;
3. reconnect the same registration and assert a new transport is live;
4. disconnect again, or unregister after the reconnect;
5. assert the plugin disconnect handler ran again for the new connection and `liveTransportCount == 0`.

Also keep the existing `reportDisconnected() -> unregister` cleanup test passing.

### 2. Serialize teardown with an in-flight connect, or otherwise fence all late connect side effects

Teardown must not finish while an earlier `connect()` can still create resources afterward.

Requirements:

- If `disconnect()`/unregister begins while `_connecting` is in flight, ensure that connect attempt cannot leave resources alive after cleanup finishes.
- Do not rely only on completing/rejecting the Dart invocation Future; that does not cancel the JavaScript Promise.
- Do not allow a late successful `host.transport.open()` from the same registration to survive teardown.
- Keep unrelated devices and current-generation plugin traffic unaffected.
- Do not weaken generation fencing or allow stale generations to open/send traffic.
- Keep the 10-second invocation timeout semantics.

The simplest safe model is usually to serialize teardown behind the active connect attempt and then run disconnect cleanup for that same connection epoch. If waiting for connect is undesirable, an equivalent cancellation/fencing design is acceptable only if it also guarantees that any transport/resources created by the late connect are deterministically closed before teardown completes.

Add a regression test with an intentionally delayed connect:

1. plugin sensor `connect()` starts and waits before or during `host.transport.open()`;
2. begin direct `device.unregister()` (and preferably also cover plain `disconnect()`);
3. let the delayed connect continue;
4. assert unregister/disconnect completes without hanging;
5. assert the registration is removed where applicable;
6. assert `manager.liveTransportCount == 0` after teardown;
7. assert no late connect result changes the sensor state or resurrects the device.

Keep the existing registration-scoped pending-command tests, unload cleanup tests, stale-generation fencing tests, and manager disposal tests passing.

### Non-blocking correctness cleanup

The device invocation JS bridge currently sends `result || {}`, so falsy handler results (`null`, `false`, `0`, `""`) are silently converted to `{}`. Dart also treats a null result as `{}` even though non-object results otherwise fail. For `execute()`, preserve the raw result and reject anything that is not an object so buggy plugin commands fail deterministically instead of appearing successful. Connect/disconnect return values can be ignored separately if desired.

### Verification

Run and report:

- `dart format lib test`
- `flutter analyze`
- affected plugin-device / plugin-manager / transport tests
- full `flutter test`

Keep all previously added lifecycle regression tests passing.

ODevStudio (changes requested):
Re-review: the previous findings are substantially fixed. Cleanup is now per connection epoch, reconnects run a fresh disconnect cleanup, direct unregister preserves an in-flight connect long enough to serialize behind it, and non-object `execute()` results are rejected correctly.

Two lifecycle blockers remain around connect retirement/timeouts.

### Blocker 1: unload during an in-flight connect waits for the connect invocation timeout even after the handler has already settled

The unload path now excludes pending `connect` invocations from `_rejectDeviceInvocations()`, which is necessary so `PluginDeviceService.disconnect()` can wait for `_connecting` before cleanup. However, after `_unloadPlugin()` advances the generation / marks the runtime stopping, `_handleDeviceMessage()` still only permits a stale-generation `invocationResult` through when the pending operation is `disconnect` and `(pluginId, generation)` is in `_deviceDisconnectCleanup`.

That means a connect that was already pending at unload cannot complete its Dart invocation normally once it resolves or rejects. Its late `invocationResult` is dropped as stale, while `_runDisconnectCleanup()` is awaiting `_connecting`. The only remaining completion path is the device invocation timer, so an ordinary unload/reload while a sensor is connecting can stall for the remaining 10-second timeout even if the JS `connect()` handler rejects immediately after unload (for example because its `host.transport.open()` is correctly rejected as stale-generation traffic).

The cleanup exception should allow the retiring generation's **already-pending connect invocation result** to settle as well as the cleanup disconnect result. This must not reopen stale-generation device requests or transport opens; it only needs to accept the result for an invocation that already exists in `_pendingDeviceInvocations` for that retiring generation.

### Blocker 2: direct unregister can still leak a transport when `connect()` outlives the 10-second invocation timeout

The new serialization works while the connect invocation finishes before `deviceInvocationTimeout`. It does not guarantee cleanup after that timeout.

`_runDisconnectCleanup()` waits on `_connecting`, but `_connecting` is the Dart invocation Future managed by `_invokeDeviceHandler()`. When the 10-second timer fires, that Future completes with `Plugin device invocation timed out` and the pending invocation is removed. The underlying JavaScript `connect()` Promise is **not cancelled**.

A normal slow-network sequence can therefore be:

```text
connect() -> await host.transport.open(...)
          -> device invocation reaches 10s timeout
unregister() -> _connecting is now failed/completed
             -> disconnect() runs before the transport handle exists
             -> registration/handler entry is removed
later -> the already-running host.transport.open() succeeds
      -> the old JS connect continuation receives/stores the handle
```

Because direct unregister leaves the plugin generation active and has no plugin-wide `closeAllForPlugin()` sweep, that late transport can remain live with no registered device left to clean it up. `PluginTransportService.open()` also creates/owns its transport record independently of the device invocation timer, so timing out the device callback does not cancel the pending transport open.

This is the same underlying rule as the prior race finding: teardown must guarantee that side effects of the retiring connection attempt cannot appear after cleanup has completed. Waiting only for the timeout-bounded public invocation Future is insufficient.

## Agent worker prompt

Fix PR #775 only for the two blockers above. Preserve the current public API, stable IDs, manifest permissions, resource limits, 10-second user-visible device invocation timeout, REST/WebSocket routes, per-connection cleanup behavior, reconnect behavior, registration-scoped execute fencing, retiring-generation transport-close exception, and manager-disposal behavior. Do not add BLE/grinder support, persistence, probing, generic discovery, or unrelated refactors.

### 1. Let a retiring generation's already-pending connect invocation settle during unload cleanup

Requirements:

- When `_deviceDisconnectCleanup` contains `(pluginId, generation)`, accept an `invocationResult` for an already-existing pending `connect` invocation owned by that same plugin/generation (and registration), in addition to the cleanup `disconnect` invocation result.
- Keep the normal generation/liveness fence for all new `register`, `publish`, `reportDisconnected`, `unregister`, `execute`, and transport traffic.
- Do not permit stale-generation `host.transport.open()` or `send()` merely because connect cleanup is in progress. The existing close-only transport exception should stay close-only.
- Do not accept arbitrary stale invocation IDs: require the invocation to still exist in `_pendingDeviceInvocations` and match plugin id/generation/operation.
- Once that connect settles/rejects, `_PluginSensor.disconnect()` should proceed immediately to the disconnect handler rather than waiting for the invocation timer.

Regression test:

1. Register a sensor whose `connect()` waits on a JS-controlled gate before attempting `host.transport.open()`.
2. Start unload while the sensor is still connecting.
3. After unload has entered retiring-generation cleanup, release the gate. The transport `open` should be rejected by the stale-generation fence and the connect handler should reject/settle.
4. Configure `deviceInvocationTimeout` to something deliberately long (for example 20-30 seconds) and assert unload completes promptly after the connect result arrives, rather than waiting for that timer.
5. Assert disconnect cleanup runs exactly once, the registration is removed, `liveTransportCount == 0`, and no stale open/send traffic was permitted.

### 2. Fence or own connect side effects beyond the user-visible invocation timeout

The 10-second timeout may fail the caller, but it must not make the underlying JS handler invisible while it is still capable of creating resources.

Required invariant:

> After direct `device.unregister()` (and plain disconnect, where applicable) finishes cleanup for a connection attempt, that attempt must be unable to create or retain a transport/resource later.

A solution may separate **caller completion** from **underlying handler settlement**, retain a teardown tombstone for timed-out connects, associate transport opens with the registration/connect attempt so they can be cancelled/closed, or use another equivalent design. The exact mechanism is flexible, but simply waiting on the current timeout-bounded `_connecting` Future is not sufficient.

Requirements:

- Preserve the 10-second user-visible invocation timeout: callers must still receive a timeout rather than waiting indefinitely.
- Continue tracking/fencing the underlying connect attempt after the caller-facing timeout if its JavaScript Promise is still running.
- A transport open that was initiated by that retired connection attempt and settles after timeout must be rejected/closed deterministically before it can become an orphan.
- Direct unregister must not solve this by calling `closeAllForPlugin()`, because that could tear down transports belonging to another registered device or unrelated plugin functionality.
- Do not reject/close unrelated devices' transports in the same plugin generation.
- Keep handler cleanup idempotent and do not resurrect the removed device or emit a late `connected` state.

Regression test using a delayed native transport connector:

1. Use a short `deviceInvocationTimeout` (for example 20-50 ms).
2. Have the sensor's `connect()` call `host.transport.open()` and retain the returned handle, but inject a `connectWebSocket`/socket connector whose Future is held by a test gate so the native open remains pending beyond the device invocation timeout.
3. Start direct `device.unregister()` while connect is in flight and allow the device invocation timeout to fire.
4. Let unregister perform its normal cleanup/removal.
5. Only then release the transport connector so the old `host.transport.open()` attempt can finish.
6. Assert `manager.liveTransportCount` remains/returns `0`, the device stays absent from `DeviceController`/`SensorController`, no late `connected` state is emitted, and cleanup is not double-invoked.
7. Keep a second device in the same plugin live if useful to prove the fix is registration-scoped rather than plugin-wide.

### Verification

Run and report:

- `dart format lib test`
- `flutter analyze`
- affected plugin-device / plugin-manager / transport tests
- full `flutter test`

Keep all previously added regressions passing, especially reconnect cleanup, `reportDisconnected() -> unregister`, direct unregister of an in-flight connect that settles before timeout, registration-scoped command rejection, unload-time transport close, stale-generation fencing, and manager disposal.

ODevStudio (changes requested):
Re-review: the two previous timeout/retirement blockers are fixed for the cases now covered by tests. Retiring connect invocation results can settle during unload, and a transport `open` that was already started by a connect invocation is now owned/retired correctly after the device invocation times out.

One blocker remains in the new connect-attribution mechanism.

### Blocker: transport ownership is lost after the first async yield inside `connect()`

`__dispatchDeviceInvocation()` sets `__activeDeviceConnect` only around the synchronous call to `handler(payload)`, then restores it immediately in `finally`:

```js
__activeDeviceConnect = { registrationHandle: handle, invocationId };
handlerResult = handler(payload);
__activeDeviceConnect = previousDeviceConnect;
```

`__transportRequest()` only adds `deviceRegistrationHandle` / `deviceInvocationId` while that global is set.

That works for the current regression test because `connect()` calls `host.transport.open()` synchronously before returning its Promise, even though the native connector itself resolves later.

It does **not** work for a normal async connect handler that yields before opening its transport, for example:

```js
async connect() {
  await waitForConfiguration();
  const opened = await host.transport.open({ ... });
  transportHandle = opened.handle;
}
```

or equivalently:

```js
connect() {
  return gate.then(() => host.transport.open({ ... }));
}
```

After the first `await` / Promise continuation, `__activeDeviceConnect` has already been restored to `null`. The later `open` request is therefore untagged.

This reopens the timed-out-unregister leak:

```text
connect() starts and waits before calling transport.open
-> device invocation reaches its 10s timeout
-> direct device.unregister()
-> timed-out connect invocation is retired
-> disconnect runs with no transport handle and registration is removed
-> old JS connect continuation resumes later
-> host.transport.open() is sent without device ownership metadata
-> plugin generation is still current, so the open is allowed
-> transport becomes live after the device has already been removed
```

Because the `open` has no registration/invocation ownership metadata, `_retiredDeviceConnects` cannot reject it and `closeDeviceConnect()` cannot find it. Direct unregister also cannot use `closeAllForPlugin()` without breaking unrelated devices/transports in the same plugin generation.

The unload case is less exposed because the stale-generation fence rejects the later open globally, which is why the new unload regression passes even though its `host.transport.open()` happens after a JS-controlled gate. Direct unregister keeps the generation current, so it needs durable registration-scoped attribution.

## Agent worker prompt

Fix PR #775 only for this remaining connect-ownership blocker. Preserve the existing public device API if practical, stable IDs, permissions, limits, the 10-second caller-facing invocation timeout, per-connection cleanup, registration-scoped command fencing, unload generation fencing, close-only retiring transport exception, and manager-disposal behavior. Do not solve this with plugin-wide transport teardown.

### Make connect transport ownership survive asynchronous continuations

The ownership token for a device `connect` invocation must remain associated with **every `host.transport.open()` initiated by that connect operation**, including opens performed after one or more `await`/Promise continuations.

Requirements:

- Do not rely on a single mutable synchronous global such as `__activeDeviceConnect` unless the design also provides correct async-context propagation.
- A timed-out/retired connect must not be able to issue an unowned transport open later.
- Existing opens belonging to a retired connect must continue to be closed/rejected as today.
- Future opens belonging to that retired connect must be rejected before becoming orphan resources.
- Ownership must remain registration/invocation scoped; do not block or close unrelated transports belonging to another device or other plugin functionality.
- Support multiple plugin devices and potentially overlapping connect attempts without cross-attributing one device's transport to another.
- Keep the caller-facing 10-second device invocation timeout unchanged.
- Late invocation results must remain unable to resurrect a removed device or emit `connected` after teardown.

An implementation could use a durable invocation-scoped transport facade/token rather than ambient synchronous state, or another mechanism that explicitly carries `(registrationHandle, invocationId)` through the transport-open path. The exact implementation is flexible, but ownership must survive arbitrary async yielding inside the connect handler.

### Required regression test

Add a variant that differs from the current delayed-native-open test in one critical way: **do not call `host.transport.open()` until after the device invocation has already timed out and unregister has completed.**

For example:

1. Register a plugin sensor with a `connect()` handler that returns a Promise and waits on a JS-controlled gate **before** calling `host.transport.open()`.
2. Use a short `deviceInvocationTimeout` (20–50 ms) and wait for the connect invocation to time out.
3. Call direct `device.unregister()` and wait until the registration is removed and disconnect cleanup has completed.
4. Only then release the JS gate so the old connect continuation attempts `host.transport.open()`.
5. Assert the open is rejected/retired and `manager.liveTransportCount == 0` after the continuation settles.
6. Assert the device remains absent from `DeviceController` / `SensorController`, no late `connected` state appears, and disconnect cleanup is not invoked twice.
7. Ideally keep another device/transport in the same plugin alive to prove the retirement is registration-scoped rather than plugin-wide.

Keep the existing delayed-native-connector regression too; the two tests exercise different failure windows.

### Verification

Run and report:

- `dart format lib test`
- `flutter analyze`
- affected plugin-device / plugin-manager / transport tests
- full `flutter test`

Keep all previously added lifecycle regressions passing.

ODevStudio (changes requested):
Re-review: the previous async-yield orphan-transport case is now prevented, but the new fallback is too broad and introduces a cross-device / cross-feature regression within the same plugin generation.

### Blocker: retiring one timed-out device connect blocks unrelated unowned transport opens for the whole plugin generation

The current fix adds `_retiredDeviceConnectInvocations` and, in `_handleTransportMessage`, rejects an `open` when it is not tagged as `ownsDeviceConnect` and **any** retired connect invocation exists for the same `(pluginId, generation)`.

That closes the previous hole for a connect handler that does an `await` before `host.transport.open()`: the late open is untagged, sees a retired connect in the generation, and is rejected.

However, the rejection is generation-wide rather than registration/invocation-scoped. After direct unregister of one timed-out sensor connect, the retired invocation entry remains until the plugin generation is unloaded. During that window, every legitimate untagged `host.transport.open()` from the same plugin generation is also rejected with `Plugin device connect retired`.

Concrete examples that now break:

```text
sensor A connect times out
-> unregister sensor A
-> A's connect invocation is added to the retired set

later, same plugin generation:
-> unrelated plugin code calls host.transport.open(...)
   OR
-> sensor B does async connect() { await something; return host.transport.open(...); }

=> the open is untagged
=> manager sees A's retired invocation in the generation
=> valid open is rejected
```

This can persist for the rest of the plugin generation after a direct unregister because `_retiredDeviceConnectInvocations` is only swept generation-wide during plugin unload/disposal. So the fix prevents the resource leak by effectively disabling unowned transport opens for unrelated code after one timed-out device connect.

The exact `PluginTransportService.retireDeviceConnect(...)` / `closeDeviceConnect(...)` ownership model is appropriately registration/invocation scoped; the problem is the manager-level generation-wide fallback for untagged opens.

## Agent worker prompt

Fix PR #775 only for this remaining ownership-scope blocker. Keep the existing public API if practical, stable device IDs, permissions, limits, 10-second caller-facing invocation timeout, per-connection cleanup, registration-scoped command fencing, retiring-generation cleanup, unload stale-generation fencing, and manager-disposal behavior. Do not solve this by blocking or closing all transports in the plugin generation.

### Preserve async connect attribution without generation-wide overblocking

Required invariant:

> Retiring device connect invocation A must prevent only A's late transport opens/resources from surviving. It must not prevent unrelated transport opens from device B or non-device plugin code in the same generation.

Requirements:

- Remove or narrow the rule that rejects every unowned `open` whenever any retired connect exists for `(pluginId, generation)`.
- Ownership/retirement must remain scoped to `(pluginId, generation, registrationHandle, invocationId)` or an equivalent identity.
- A `connect()` that yields one or more times before calling `host.transport.open()` must still carry enough durable context for the later open to be attributed to that specific connect invocation.
- A retired/timed-out connect's later open must be rejected or deterministically closed.
- An unrelated plugin-level `host.transport.open()` in the same generation must continue to work.
- A second sensor's async `connect()` in the same plugin generation must continue to work and must not inherit the first sensor's retired state.
- Multiple overlapping device connect attempts must not be cross-attributed.
- Keep the 10-second user-visible device invocation timeout unchanged.
- Do not resurrect removed devices or emit a late `connected` state.

A durable invocation-scoped transport capability/facade, explicit token propagation, or equivalent async-context mechanism is preferable to a generation-wide ambient tombstone. The implementation is flexible as long as attribution survives arbitrary Promise/`await` continuations and remains scoped.

### Required regression coverage

Add a test that proves both sides of the invariant:

1. Register sensor A whose `connect()` waits on a JS gate **before** calling `host.transport.open()`.
2. Use a short `deviceInvocationTimeout`, allow A's connect to time out, then direct-unregister A and wait for removal/cleanup.
3. While A's retired-connect state still exists, perform an **unrelated plugin-level** `host.transport.open()` in the same plugin generation and assert it succeeds; close it normally.
4. Preferably also register sensor B whose `connect()` yields before `host.transport.open()` and assert B can connect successfully after A was retired.
5. Only then release A's old gate and let its stale continuation attempt `host.transport.open()`.
6. Assert A's late open is rejected/retired, `liveTransportCount` reflects only legitimate remaining transports, A stays absent, and no late `connected` state is emitted.
7. Keep the current delayed-native-open regression, because it covers a different timing window.

### CI / verification

The current head's PR checks are failing as of this review; fix the failing run as part of the update and report the cause.

Run and report:

- `dart format lib test`
- `flutter analyze`
- affected plugin-device / plugin-manager / transport tests
- full `flutter test`

Keep all previously added lifecycle regression tests passing.

ODevStudio (changes requested):
Re-review at `e2334ff9`: the previous generation-wide **rejection** is fixed and current PR checks are green. However, the replacement `retiredConnectCandidate` heuristic still has a blocker on normal multi-transport / multi-device usage.

### Blocker: generation-wide candidate marking can close legitimate persistent transports and can still miss the stale transport when another connect is pending

The current rule marks an unowned `open` as a retired-connect candidate when:

- some retired connect exists for `(pluginId, generation)`, and
- there is no currently pending device `connect` invocation in that generation.

When the retired invocation later settles, `closeRetiredConnectCandidates(pluginId, generation)` closes **all** records marked by that heuristic.

This is still not ownership-scoped, and there are two ordinary failure modes.

#### 1. Legitimate long-lived plugin transport is closed when stale connect A settles

A persistent WebSocket is a normal plugin transport use case, not an edge case:

```text
sensor A connect times out
-> direct-unregister A
-> A remains retired/unsettled

unrelated plugin code opens a persistent websocket
-> no other connect is pending
-> open is accepted but marked retiredConnectCandidate

later A's stale JS connect handler settles
-> closeRetiredConnectCandidates() runs
-> unrelated persistent websocket is closed
```

The new regression avoids this by explicitly closing the unrelated transport **before** releasing A's gate. That proves admission succeeds, but it does not prove the unrelated transport survives the retirement sweep. The implementation comment itself effectively concedes this behavior: unrelated opens survive only if they are no longer live when the stale attempt settles.

#### 2. A's stale transport can leak if sensor B is still connecting

The inverse timing is also broken:

```text
A is retired/unsettled
B starts an async connect and remains pending
A's old async continuation resumes and calls host.transport.open()
```

A's open is untagged (the original async-context problem still exists), but because **some** connect invocation (B) is pending, `retiredConnectCandidate` is false. A's stale transport is therefore admitted without ownership and without the candidate flag. When A then settles, the candidate sweep cannot find/close it, so direct unregister can still leave A's orphan transport live.

This is a realistic startup/reconnect scenario for a plugin registering multiple sensors: connect attempts can overlap naturally. The current tests deliberately settle/connect B before releasing A, so they do not exercise this ordering.

The core issue is that `pending connect exists somewhere in the generation` is not a valid proxy for ownership of an untagged transport open. The heuristic necessarily over-attributes some unrelated opens and under-attributes some stale opens.

## Agent worker prompt

Fix PR #775 only for this remaining connect-ownership blocker. Preserve the current public device behavior where practical, stable IDs, permissions, limits, 10-second caller-facing invocation timeout, per-connection cleanup/reconnect behavior, registration-scoped execute fencing, unload generation fencing, close-only retiring transport cleanup, and manager-disposal behavior.

### Required invariant

A transport created for device connect invocation A must be attributable to A across arbitrary async yields. Retiring A must close/reject A's transports only. It must neither close unrelated plugin transports nor depend on whether another device connect happens to be pending.

Requirements:

- Remove the generation-wide `retiredConnectCandidate` ownership heuristic (including the `no pending connect in generation` discriminator) as the final correctness mechanism.
- Use an invocation-scoped transport capability/facade/token or another explicit ownership mechanism that survives `await` / Promise continuations.
- A timed-out/unregistered connect A that calls transport open later must be rejected or deterministically closed.
- An unrelated persistent `host.transport` connection in the same generation must remain live when A later settles.
- A second sensor B may be connecting concurrently; A's stale open must still be attributed/retired even while B remains pending.
- B's own transport must not be claimed or closed by A.
- Do not solve this with `closeAllForPlugin()` or any generation-wide sweep of otherwise-unowned live transports.
- Late A results must not resurrect the device or emit a late `connected` state.

Because this device API is being introduced by this PR, an explicit invocation-bound transport object passed to `connect()` is acceptable if ambient async attribution cannot be made reliable in JSCore/QuickJS. The exact API is flexible; the ownership must be explicit and durable rather than inferred from generation timing.

### Required regressions

Add both tests; they exercise common behavior the current suite skips:

1. **Persistent unrelated transport survives A settling**
   - Let A time out and unregister while its JS connect remains gated.
   - Open an unrelated plugin-level WebSocket in the same generation and **keep it open**.
   - Release A and let its stale continuation/handler settle.
   - Assert A's stale transport is gone/rejected, but the unrelated WebSocket is still live and usable (for example, a send/round-trip still succeeds) until explicitly closed.

2. **A is retired correctly while B is still pending**
   - Retire/unregister A while A's connect is gated.
   - Start B's async connect and keep B pending on its own gate.
   - Release A **before B settles**, so A attempts its late transport open while B is still a pending connect.
   - Assert A's late transport cannot survive and `liveTransportCount` does not gain an orphan from A.
   - Then release B and assert B connects normally and its transport remains live.

Keep the existing delayed-native-open, stale-async-connect, reconnect cleanup, unload cleanup, registration-scoped command, and stale-generation tests passing.

### Verification

Current head CI is green; keep it green. Run/report:

- `dart format lib test`
- `flutter analyze`
- affected plugin-device / plugin-manager / transport tests
- full `flutter test`

ODevStudio (approved):
Re-review at `a9a52722`: approved.

The remaining transport-ownership blocker is resolved by making ownership explicit rather than inferring it from generation timing. Device `connect(transport)` now receives an invocation-bound transport facade; `transport.open()` carries the exact registration/invocation identity through arbitrary Promise/`await` continuations, and the manager validates that claim against the live connect attempt before admitting the open.

The previous generation-wide `retiredConnectCandidate` heuristic and sweep are removed, so unrelated persistent plugin transports are no longer candidates for retirement. A retired connect's later claimed open is rejected, including while another device connect remains pending.

The focused regressions now cover both cases from the previous review:

- a persistent unrelated plugin WebSocket remains live and usable after retired A settles;
- retired A cannot open a transport while B is still pending, and B subsequently connects normally.

The earlier lifecycle regressions remain represented in the suite: reconnect cleanup, reportDisconnected/unregister cleanup, registration-scoped pending-command fencing, delayed native opens, unload-time cleanup/transport close, stale-generation fencing, and manager disposal.

Current-head PR checks are green. I do not see another ordinary-path blocker in this change set.

PR #781, opened 2026-09-04
fix(ble): retry a write rejected for its write property

### LINKED ISSUE #780: [Bug Report] Eclair scale can not tare
## Description
Eclair scale can not tare

## System Info
**System Information:**
- App Version: 0.8.5 (build 2624)
- Commit: a08bc41e
- Branch: HEAD
- Platform: macos
- OS Version: Version 26.5.2 (Build 25F84)
- Dart Version: 3.12.2 (stable) (Tue Jun 9 01:11:39 2026 -0700) on "macos_arm64"

## Attachments
Logs and screenshots: https://gist.github.com/reaprime-bot/563e77e11b8b459000fcc3fb655f83a2

---
_Submitted via in-app feedback on 2026-09-04T11:26:08.502477_

---
**Contact:** `1`

allofmeng:
## Summary

What changed, and why?

- The Eclair command characteristic `4F9A45BA-8E1B-4E07-E157-0814D393B968` is write-with-response only on current firmware, but `AtomheartScale._safeWrite` issued every command as write-without-response. CoreBluetooth rejects the mismatch against its own GATT database before any radio traffic, so each tare failed in ~3 ms with `characteristicDoesNotSupportWriteWithoutResponse`.
- That code was unmapped in `_handleGattError`, whose final `throw e` logged nothing, and `ScaleHandler` turned the exception into a 500 body it never logged. #780 therefore arrived as a bare HTTP 500 with no cause anywhere in the user's log.
- `AtomheartScale` now issues its commands with response. The device contract is fixed at the caller, not papered over in the transport.
- `UniversalBleTransport.write` retries in one direction only: a write the caller asked to send unacknowledged that the platform rejects for its property is retried once with response. That adds an acknowledgement the caller did not ask for and never removes one it did. A rejected write-with-response is surfaced as-is, never downgraded — a write request and a write command are different ATT procedures with different error paths, and the transport does not treat them as interchangeable.
- `_handleGattError` logs before its unmapped rethrow; the scale tare and timer routes log the exception behind a 500; the Eclair readiness-gate timeout now reports how many notifications arrived and which frame was rejected.

The evidence chain, from the reporter's log:

```
11:20:42 ScaleController - scale connection update: connected
11:20:50 PUT [500] /api/v1/scale/tare   0:00:00.003445
11:21:08 PUT [500] /api/v1/scale/tare
11:21:16 PUT [500] /api/v1/scale/tare
```

Connected with weight frames parsing, yet tare 500s in 3 ms with zero log output. 3 ms means no radio round trip. Zero log output rules out every code `_handleGattError` maps (`characteristicNotFound`, device-gone, GATT 133, unknown error), all of which log at WARNING, leaving only the unmapped silent rethrow.

## Review follow-up

Both blocking items from the first review are addressed in `815430ed`:

1. **No silent downgrade.** The `withResponse -> withoutResponse` fallback is gone, along with the docs claim that the two modes are equivalent for short command frames. `BLETransport.write(..., withResponse: true)` is honoured as an explicit caller choice and its rejection is surfaced. The Eclair contract is fixed at the caller (`atomheart_scale.dart` now writes with response). The remaining upward retry only ever adds an acknowledgement.
2. **Format gate.** The five dart files this PR touches are formatted with `dart_style 3.1.13`, matching the formatter on CI's stable Flutter (3.47.2); running the gate's own command against them exits 0. Nothing else is reformatted, so the gate's `BASE_SHA...HEAD_SHA` file list stays limited to this PR's own files.

## Linked Issue

Fixes #780

## Verification

- Format gate reproduced locally with `dart_style 3.1.13` over the PR's dart files — 0 changed, exit 0.
- `flutter analyze` — 4 warnings, all pre-existing missing asset directories (`assets/plugins/dye2.reaplugin/`, `assets/plugins/shot-upload.reaplugin/`, `assets/plugins/dcamp.reaplugin/`, `assets/bundled_skins/`) referenced by `pubspec.yaml` and fetched at build time, not tracked in the repo.
- `flutter test` — 3828 passed, 42 failed. Every failure is in a suite that loads one of those absent asset directories: `test/plugins/shot_upload_reconciliation_test.dart` (25), `test/webui_support/webui_token_injection_test.dart` (8), `test/plugins/shot_upload_plugin_test.dart` (4), `test/plugins/bundled_plugin_permissions_test.dart` (4), `test/webui_storage_bundled_test.dart` (1), `test/hive_store_service_test.dart` (1). No BLE, scale, transport, or webserver test fails.
- Targeted suites green: `test/universal_ble_transport_recovery_test.dart`, `test/unit/models/atomheart_scale_reliability_test.dart`, `test/atomheart_scale_test.dart`, `test/unit/models/atomheart_scale_connect_failure_test.dart`, `test/controllers/de1_controller_test.dart`, `test/controllers/shot_sequencer_test.dart` — 104 passed.
- Write-property tests in `test/universal_ble_transport_recovery_test.dart`: without-response falls back to with-response; **a rejected write-with-response is never downgraded** (throws after exactly one attempt); a characteristic supporting neither still throws after exactly two attempts from a without-response caller; a supported write is issued exactly once with the property the caller asked for.
- Tests in `test/unit/models/atomheart_scale_reliability_test.dart`: tare and the three timer opcodes are all written with response; a silent subscription names `0 notifications`; a rejected frame is named in the connect failure with its bytes.
- No hardware testing. The root cause is derived from the reporter's log and from universal_ble's platform sources; a maintainer with an Eclair should confirm the tare before release.

## Impact

- Eclair tare and timer start/stop/reset now work instead of returning HTTP 500. Eclair commands become acknowledged writes, which is what the characteristic advertises.
- No behavior change for a device whose write already succeeds: the retry lives in the exception path, gated to two error codes and to a without-response caller, and the success path issues the same single write with the caller's property.
- The retry cannot duplicate a command. Darwin (`UniversalBlePlugin.swift:445`), Android (`UniversalBlePlugin.kt:899`), and Windows (`universal_ble_plugin.cpp:406`) all validate the property and return before dispatching to the radio, so a rejected command was never delivered; BlueZ never reports these codes and never enters the retry path.
- No caller depends on the old failure. Every device write passes a fixed property except Acaia, which selects by protocol (`acaia_scale.dart:163`) and is correct today; nothing catches a write failure to probe for capability.
- For the retry code path `_handleGattError` previously reached its bare rethrow, running neither `_probeAndDeclareIfDead` nor `_clearQueue`, so skipping it drops no side effect that used to run.
- New WARNING-level log lines on failure paths only. No API, spec, schema, or migration changes.
- `doc/AI_BLE_NOTES.md` records the write-property contract, that the two ATT write procedures are not interchangeable, why the one-way retry cannot duplicate a command, and why a readiness gate must name what it saw.

## Contributor Responsibility

https://claude.ai/code/session_011yWbCbjVvE1BBhizYRk35k


tadelv (changes requested):
Two required fixes before merge:

1. **Please do not silently downgrade a requested write-with-response to write-without-response at the transport layer.** `BLETransport.write(..., withResponse: true)` currently expresses an explicit caller choice. The new bidirectional fallback changes ATT semantics: a write request has a server response/error path, while a write command does not. The local property-rejection check makes the retry non-duplicating, but it does not make the two write procedures equivalent. For #780 the known Eclair command characteristic is write-with-response-only while `AtomheartScale._safeWrite` explicitly requests `withResponse: false`; please fix that device contract at the caller (or add an explicit/capability-driven negotiation policy) rather than globally turning a requested acknowledged write into an unacknowledged one. At minimum, remove the `withResponse -> withoutResponse` fallback and its test/docs claim that the two modes are equivalent for short command frames.

2. **Current CI fails the required Dart format check.** The `Analyze / test` job stops at `Dart format check`, so analyze/tests are skipped. `AGENTS.md` explicitly requires `dart format lib test` after meaningful changes and before considering the work done. Please format the changed files and get that gate green.

tadelv (changes requested):
The two original blockers are fixed: Eclair now uses write-with-response at the caller, the transport no longer downgrades acknowledged writes, and the current CI run is fully green (format, analyze, tests, Linux smoke).

One remaining required cleanup before merge: please revert the unrelated `pubspec.lock` changes. This PR currently downgrades several transitive packages (`intl` 0.20.3→0.20.2, `matcher` 0.12.20→0.12.19, `meta` 1.19.0→1.18.0, `test_api` 0.7.12→0.7.11, `vector_math` 2.4.2→2.2.0). Current `main` still has the newer lockfile values, and these dependency changes are unrelated to #780; they look like fallout from running `pub get` with the older local Flutter SDK mentioned in the follow-up. Please restore `pubspec.lock` to `main` and rerun CI.

tadelv (approved):

PR #782, opened 2026-09-04
fix(serial): deduplicate Android devices by ID

### LINKED ISSUE #767: [Bug] Duplicate deviceId in /api/v1/devices after reconnect churn (same machine listed disconnected + connected)
## Summary

`GET /api/v1/devices` (and the devices websocket) can return the **same deviceId twice** — once `disconnected`, once `connected`, both `available: true`. Observed after connect/reconnect churn where a machine that was already connected got re-adopted via the automatic quick-connect path.

Observed output (ids scrubbed):

```
Bengle        | usb-2e8a-a-<serial> | disconnected | machine
Bengle        | usb-2e8a-a-<serial> | connected    | machine
```

4 machine entries for 3 unique ids. Clients that key by `deviceId` (skins, inspectors) silently drop one of the pair, so the live machine can appear disconnected depending on ordering.

## Evidence (tablet logs)

Explicit connect of a serial machine, then auto re-adoption of the *same* machine ~3 minutes later while the first record was still tracked:

```
08:17:36.494 Serial - port connected (id=usb-2e8a-a-<serial>)     ← explicit connect, record A added
08:17:37.090 PUT /api/v1/devices/connect [200]
08:19:55.437 PUT /api/v1/devices/disconnect [200]                 ← other transport (F2:5C…) disconnected
08:20:06.518 Serial - port connected (id=usb-2e8a-a-<serial>)     ← auto quick-connect, record B added
08:20:07.006 ConnectionManager - Quick-connect: machine adopted (usb-2e8a-a-<serial>)
```

No `USB_DETACHED` or machine-disconnect between the two connects; from then on the devices list carried two records with the same id in different states.

## Root cause

`lib/src/services/serial/serial_service_android.dart` keeps a plain `List<Device> _devices` with three add sites, and only one of them dedupes:

- `_connectAttachedMachine` — **guarded**: skips if `_devices.any((d) => d.deviceId == stableId)` (line ~208).
- `tryQuickConnect` — **unguarded**: after `device.onConnect()` it unconditionally does `_devices.add(connected)` (line ~325), with no check for an existing record with the same `deviceId`.
- `_performScan` — safe: filters enumeration against ids already in `_devices` before `addAll`.

Record removal is by instance identity: each added device self-removes only through *its own* `connectionState` listener (`_devices.remove(connected)`), and `_transportForDeviceId` is keyed by `deviceId` — so when a second record for the same id is created, a superseded first record's transport can be disposed via the shared map slot without the record itself being removed, leaving a stale `disconnected` record beside the live one.

Downstream, `DeviceController.devices` flattens the per-service lists with **no dedupe**, and `buildAvailabilityDeviceList` (`lib/src/services/webserver/devices_handler.dart`) dedupes remembered-vs-live but assumes live ids are already unique — so the duplicate reaches the API and the devices websocket unchanged.

## Suggested fix directions

1. In `SerialServiceAndroid.tryQuickConnect`, before `_devices.add(connected)`, skip/remove any existing record with the same `deviceId` (mirror the `_connectAttachedMachine` guard).
2. Defensively collapse duplicate ids in `buildAvailabilityDeviceList` (or `DeviceController.devices`) so the public surface can never emit the same id twice with conflicting states.
3. Consider keying records by `deviceId` (map) instead of a list with instance-identity removal.

## Repro

1. Connect a serial Bengle machine while no machine slot is held in a way the auto-reconnect notices.
2. Disconnect the transport that currently owns the machine slot (e.g. a second connected machine), which triggers `ConnectionManager` auto quick-connect on the remembered serial machine.
3. `GET /api/v1/devices` → same machine id appears twice (one `connected`, one stale `disconnected`).

Environment: Android tablet, Decaid, Bengle over composite USB serial. Severity: medium — stale duplicate confuses clients keyed by `deviceId` and masks the real connection state.

ODevStudio:
## Summary

- Replace same-ID Android serial devices at the shared registry insertion point so quick-connect cannot retain a stale duplicate.
- Prevent superseded devices from clearing the replacement's physical ownership and transport when their stale disconnect listeners fire.
- Cover scan-to-quick-connect API inventory and stale-listener detach regressions, and document the registry invariant.

## Linked Issue

Fixes #767

## Verification

- `dart format lib/src/services/serial/serial_service_android.dart test/services/serial/usb_attach_probe_test.dart`
- `flutter test --no-pub test/services/serial/usb_attach_probe_test.dart test/unit/services/availability_device_list_test.dart test/devices_handler_test.dart` (59 passed)
- `flutter test --no-pub` (3,865 passed, 1 skipped)
- I could not run the Android hardware API smoke: `adb devices -l` returned no connected devices. Earlier Windows desktop and emulator attempts also hit host toolchain and Gradle loopback errors.

## Impact

- Android REST and WebSocket device inventory no longer exposes stale and connected serial instances with the same `deviceId`.
- A superseded instance can no longer clear the active replacement's detach ownership or transport.
- No API schema, migration, compatibility, or security impact.

## Contributor Responsibility

### DIFF
--- a/lib/src/services/serial/serial_service_android.dart
+++ b/lib/src/services/serial/serial_service_android.dart
@@ -262,7 +262,7 @@ class SerialServiceAndroid
         deviceName: machine.name,
       );
     }
-    _devices.add(machine);
+    _addDevices([machine]);
     _recordPhysicalOwnership(machine.deviceId, device);
     machine.connectionState.listen((state) {
       if (state == ConnectionState.disconnected) {
@@ -340,7 +340,7 @@ class SerialServiceAndroid
       try {
         await device.onConnect().timeout(const Duration(seconds: 10));
         final connected = device;
-        _devices.add(connected);
+        _addDevices([connected]);
         _recordPhysicalOwnership(connected.deviceId, d);
         connected.connectionState.listen((state) {
           if (state == ConnectionState.disconnected) {
@@ -393,6 +393,13 @@ class SerialServiceAndroid
     final physicalId = physical.deviceId;
     if (physicalId != null) {
       _logicalToPhysicalDeviceId[logicalDeviceId] = physicalId;
+    }
+  }
+
+  void _addDevices(Iterable<Device> devices) {
+    for (final device in devices) {
+      _devices.removeWhere((known) => known.deviceId == device.deviceId);
+      _devices.add(device);
     }
   }
 
@@ -509,7 +516,7 @@ class SerialServiceAndroid
         return found;
       }),
     );
-    _devices.addAll(results.expand((e) => e));
+    _addDevices(results.expand((e) => e));
     _machineSubject.add(_devices);
   }
 
--- a/lib/src/services/serial/serial_service_android.dart
+++ b/lib/src/services/serial/serial_service_android.dart
@@ -266,7 +266,7 @@ class SerialServiceAndroid
     _recordPhysicalOwnership(machine.deviceId, device);
     machine.connectionState.listen((state) {
       if (state == ConnectionState.disconnected) {
-        _devices.remove(machine);
+        if (!_devices.remove(machine)) return;
         _machineSubject.add(_devices);
         _logicalToPhysicalDeviceId.remove(machine.deviceId);
         final t = _transportForDeviceId.remove(machine.deviceId);
@@ -344,7 +344,7 @@ class SerialServiceAndroid
         _recordPhysicalOwnership(connected.deviceId, d);
         connected.connectionState.listen((state) {
           if (state == ConnectionState.disconnected) {
-            _devices.remove(connected);
+            if (!_devices.remove(connected)) return;
             _machineSubject.add(_devices);
             _logicalToPhysicalDeviceId.remove(connected.deviceId);
             final t = _transportForDeviceId.remove(connected.deviceId);
--- a/doc/DeviceManagement.md
+++ b/doc/DeviceManagement.md
@@ -893,6 +893,10 @@ attach recovery"). Explicit native, REST, and WebSocket scans call
 `connectingMachine` is published before the attempt, then phase falls
 through to `scanning` (existing scan path).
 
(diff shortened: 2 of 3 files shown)


lib/src/services/serial/serial_service_android.dart
```diff
@@ -396,6 +396,13 @@ class SerialServiceAndroid
     }
   }
 
+  void _addDevices(Iterable<Device> devices) {
+    for (final device in devices) {
+      _devices.removeWhere((known) => known.deviceId == device.deviceId);
```
tadelv, line 401: This removes the superseded instance from the registry, but it does not retire that instance's lifecycle. In particular, quick-connect/attach instances have a `connectionState` listener that later does `_transportForDeviceId.remove(deviceId)` and disposes the result. `_runDetection()` has already stored the replacement transport under that same ID before `_addDevices()` runs, so a disconnect from the old instance can remove/dispose the *new* transport. Even for scan-created stale instances, this path bypasses the existing `_releaseDisconnected` cleanup. Please make replacement explicitly clean up/neutralize the superseded instance without touching the replacement's transport/ownership, and add a regression test proving a stale instance disconnect after replacement cannot tear down the live one.

tadelv (changes requested):
The dedupe direction is sound, but replacement currently drops the superseded object from `_devices` without transferring/cleaning its lifecycle. Because serial ownership and transport cleanup are keyed by the same `deviceId`, an old instance can still affect the newly inserted instance after replacement. Please fix the replacement lifecycle and cover it with a regression test.

lib/src/services/serial/serial_service_android.dart
ODevStudio: Addressed in b697aba9. Both serial lifecycle listeners now stop unless they remove their own registered instance, so a superseded device cannot clear the replacement's ID-keyed ownership or transport. The new regression quick-connects A and B under the same ID, emits A's stale disconnect, then verifies the physical detach still resolves and disconnects B.

tadelv (approved):

Issue #338, opened 2026-06-16
[Bug]PUT /api/v1/workflow hangs forever when profile JSON fails enum parse


allofmeng:
WorkflowHandler._updateWorkflow debounces incoming PUTs by stashing a Completer<Response> and scheduling _applyPendingUpdate via Timer. The timer callback is fire-and-forget — its Future is never awaited and there is no try/catch:

// workflow_handler.dart:39-52
Future<Response> _updateWorkflow(Request req) async {
    final payload = await req.readAsString();
    final Map<String, dynamic> json = jsonDecode(payload);
    _pendingMerge = deepMergeJson(_pendingMerge, json);
    final completer = Completer<Response>();
    _pendingResponses.add(completer);
    _debounceTimer?.cancel();
    _debounceTimer = Timer(_debounceDuration, _applyPendingUpdate);
    return completer.future;
}

Future<void> _applyPendingUpdate() async {
    ...
    final updatedWorkflow = Workflow.fromJson(resultJson); // throws
    ...
    for (final completer in responses) {
      completer.complete(jsonOk(updatedWorkflow.toJson()));
    }
}

If Workflow.fromJson throws — e.g. ExitType.values.byName('weight') since profile.dart:129 declares enum ExitType { pressure, flow } — the throw escapes the async function silently. The completer is never completed; the HTTP request hangs until the connection times out. Any further PUTs also stall because _pendingMerge retains the bad payload until the next successful merge.

Reproduce:
curl -X PUT http://localhost:8080/api/v1/workflow \
  -H 'Content-Type: application/json' \
  -d '{"profile":{"version":"2","title":"x","steps":[{"name":"p","pump":"flow","transition":"fast","flow":2,"temperature":93,"sensor":"coffee","seconds":10,"volume":0,"exit":{"type":"weight","condition":"over","value":36}}]}}'
Request never returns. Subsequent PUTs also stall.

Suggested fix: wrap the merge/parse in try/catch, complete all queued completers with a 400 (Bad Request), and clear _pendingMerge:

Future<void> _applyPendingUpdate() async {
    final merge = _pendingMerge;
    final responses = List<Completer<Response>>.from(_pendingResponses);
    _pendingMerge = {};
    _pendingResponses.clear();

    try {
        final oldWorkflow = _controller.currentWorkflow;
        final currentJson = oldWorkflow.toJson();
        final resultJson = deepMergeJson(currentJson, merge);
[...]

PR #786, opened 2026-09-04
fix(shots): read back shots imported from de1app

### LINKED ISSUE #784: [bug] shot history can not be rendered in streamline.js
see attached logs, user reported that shot history couldn't be shown when tap nav button below in streamline.js 

[nohisotry_log.txt](https://github.com/user-attachments/files/31830692/nohisotry_log.txt)
[webview_console.log](https://github.com/user-attachments/files/31830693/webview_console.log)

allofmeng:
## Summary

What changed, and why?

- de1app `.shot` files record the shot, not the profile that produced it, so `TclShotParser` (`tcl_shot_parser.dart:48`) and the no-profile branch of `ShotV2JsonParser` (`shot_v2_json_parser.dart:80`) store a placeholder profile with an empty `steps` array. `Profile.fromJson` (`profile.dart:52`) refuses to read that back, so importing a de1app folder made every later `GET /api/v1/shots` throw `ArgumentError` and return 500, and streamline.js rendered no history at all.
- `DriftStorageService` mapped rows with a bare `rows.map(ShotMapper.fromRow)`, so the single unreadable row aborted the whole page — one 2021 import hid a history of otherwise fine shots.
- `ShotMapper.fromRow` now reads the stored workflow with `Workflow.fromRecordedJson`, which parses its profile with `Profile.fromRecordedJson`. A recorded profile is history, not something to brew from.
- `Workflow.fromJson` and `Profile.fromJson` stay strict, backing the profile library, the profile and workflow REST handlers, the stored current workflow, and DE1 upload.
- Reading was only half of it (review feedback). `PUT /api/v1/shots/<id>` merges the patch into `existingShot.toJson()` and reparses the result, and `ShotImporter` reads back shots from a backup export; both used the strict `ShotRecord.fromJson`, so a recovered de1app shot read back fine but threw again on any annotation edit or re-import. `ShotRecord.fromRecordedJson` mirrors the same split and is used on those two stored-shot paths; `ShotRecord.fromJson` is unchanged and stays strict everywhere else.
- `ShotMapper.fromRows` skips and logs a row it cannot map, so future corruption costs its own shot instead of the entire list. Single-shot reads (`getShot`, `getLatestShot`) still surface the error, because there the failing row is the answer.
- Two more strict paths, found by review (b7fd4718). `ShotExportSection.importJson` — the restore side of the app's own `/backup` archive endpoint, distinct from `ShotImporter` — still used strict `ShotRecord.fromJson`, so a recovered de1app shot could be exported but rejected on restore; switched to `fromRecordedJson`. Separately, `ShotExportSection.exportJson` treats a page shorter than `pageSize` as end-of-stream, but `ShotMapper.fromRows` silently drops unmappable rows, so a raw page containing one unreadable row came back short even with more rows waiting, truncating the backup. The `pageShots` wiring is now a standalone `pageShotsForExport` (`lib/main.dart`) that re-queries past a dropped row — advancing the cursor from the raw row, not the last successfully mapped shot — until it either fills the requested page or the table is genuinely exhausted.

The evidence, from the logs attached to the issue. `webview_console.log`:

```
[streamline.js] [WARN] Could not load shot history source: Error: HTTP error! status: 500
```

`nohisotry_log.txt`, 132 identical failures:

```
19:53:27 SEVERE ShotsHandler - Error getting paginated shots
### ArgumentError: Invalid argument(s): Profile must have a non-empty "steps" array
#0  new Profile.fromJson (profile.dart:52)
#1  new Workflow.fromJson (workflow.dart:68)
#2  ShotMapper.fromRow (shot_mapper.dart:12)
#9  DriftStorageService.getShotsPaginated (drift_storage_service.dart:95)
#10 ShotsHandler._getShots (shots_handler.dart:107)
```

The timeline pins the trigger to a de1app import:

```
19:38-19:52  GET /api/v1/shots  -> 200   (repeatedly, history works)
19:53:16     SafFolderCopier - Picked directory: de1plus ... Copied 6 files
19:53:27     GET /api/v1/shots  -> 500   (and every call after)
```

The poison row is named by the one single-shot failure: `GET /api/v1/shots/de1app-1626149813` -> 500, while `de1app-1785894912`, `-1785895243` and `-1785895651` still return 200. Unix `1626149813` is July 2021, an old TCL-format `.shot` file; the 2026 ones came through the v2 JSON parser's real-profile branch.

## Linked Issue

Fixes #784

## Verification

- `flutter test` — 3868 passed, 8 failed, out of 3877 (the new suite added 2 tests). The 8 are `test/webui_support/webui_token_injection_test.dart` failing with `SocketException: Failed to create server socket (OS Error: Address already in use, errno = 48), address = 0.0.0.0, port = 3000`, caused by a Decaid instance running on the dev machine holding port 3000. They fail identically on clean `origin/main` with none of this branch's changes, and are unrelated to shots.
- Regression suite `test/database/shot_history_imported_profile_test.dart` (11 tests), against an in-memory Drift database (the update tests drive the real `ShotsHandler`):
  - a stored profile with no steps reads back (reproduces `de1app-1626149813`);
  - a stored profile with no title reads back;
  - one step-less shot does not hide the rest of the history (the actual issue: the page returns all three shots instead of throwing);
  - `PUT /api/v1/shots/de1app-1626149813` with an annotation patch on a step-less shot returns 200, persists the note, and leaves the profile step-less (both update tests return 500 on the previous commit);
  - the same for a title-less imported shot, which reads back as `Unknown profile`;
  - a row that cannot be mapped at all is skipped, not fatal;
  - `getShot` still surfaces the failure for a single unreadable id;
  - an unreadable row does not truncate the rest of a paged backup — drives the real `pageShotsForExport` (via `import 'package:reaprime/main.dart' as app;`) against a real Drift DB with `pageSize: 2`, an unmappable row placed inside what would otherwise be the first page, and asserts all three good shots are exported. Confirmed this fails (truncates to one shot) against the pre-fix callback;
  - `Workflow.fromJson` still rejects a step-less profile while `Workflow.fromRecordedJson` accepts it;
  - `Profile.fromJson` still rejects an empty steps array and an empty title.
- `test/data_export/shot_export_section_test.dart` gained "imports a recorded shot with a step-less imported profile (gh#784)", driving `ShotExportSection.importJson` directly with a de1app-style step-less-profile record — exercises the `/backup` restore endpoint specifically, since `ShotImporter` is a separate code path.
- `test/webserver/workflow_handler_test.dart` caught a first attempt that made `Workflow.fromJson` lenient everywhere: `PUT /api/v1/workflow` with an empty profile title returned 200 instead of 400, breaking issue #338. Leniency is now reached only through `Workflow.fromRecordedJson`, that suite is green (47 tests), and the new suite pins the boundary.
- No hardware testing; this is a storage read path exercised entirely by the Drift tests.

## Impact

- Users who imported a de1app folder get their shot history back. `GET /api/v1/shots` returns 200 instead of 500, and streamline.js renders the list again. This recovers existing databases with no migration — the offending rows become readable rather than needing repair or deletion.
- A recorded profile with no title reads back as `Unknown profile`, and missing `tank_temperature` / `target_volume_count_start` read back as `0`. These only apply to profiles embedded in stored shots.
- No behavior change for the profile library or any write boundary: with `requireExecutable: true` every guard still fires before the new fallbacks, so `Profile.fromJson` is byte-for-byte the parser it was. `PUT /api/v1/workflow` still returns 400 for a step-less profile (issue #338), verified by the existing suite.
- The app's `/backup` archive round-trips a recovered de1app shot: export via `ShotExportSection.exportJson` no longer truncates around an unreadable row, and `ShotExportSection.importJson` can restore a step-less-profile shot from that archive. `ShotImporter` (a separate, non-`/backup` code path) already handled this from the prior commit.
- Recovered de1app shots are editable: annotations, notes and metadata can be written through `PUT /api/v1/shots/<id>` instead of returning 500.
- No API, spec, schema, or migration changes. No REST or WebSocket surface was touched, so `assets/api/rest_v1.yml` and `assets/api/websocket_v1.yml` are unchanged.
- `doc/Profiles.md` documents the strict/lenient split and when each parser applies. `doc/AI_STORAGE_NOTES.md` records why de1app imports carry a step-less profile, why one bad row used to hide the whole history, the single-shot-read exception, why every stored-shot JSON read path (row mapper, shot update, backup import/export) has to be lenient together, and why `pageShotsForExport` re-queries past a dropped row instead of trusting the mapped page length.
- The importers still store `steps: []`. That is the honest representation of a `.shot` file, and it now round-trips.

## Contributor Responsibility

https://claude.ai/code/session_01KYQbrfYh9W4xTy2tsEya3m

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -560,7 +560,7 @@ void main(List<String> args) async {
             cursorTimestamp: afterTimestamp,
             cursorId: afterId,
           );
-          return rows.map(ShotMapper.fromRow).toList();
+          return ShotMapper.fromRows(rows);
         },
         pageSteams: (limit, {afterTimestamp, afterCreatedAt, afterId}) async {
           final rows = await appDatabase.steamDao.getSteamsForExport(
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -44,6 +44,7 @@ import 'package:reaprime/src/plugins/plugin_source_service.dart';
 import 'package:reaprime/src/services/android_updater.dart';
 import 'package:reaprime/src/services/wifi/wifi_scale_discovery_service.dart';
 import 'package:reaprime/src/services/database/database.dart' hide Workflow;
+import 'package:reaprime/src/models/data/shot_record.dart' as domain;
 import 'package:reaprime/src/services/database/mappers/shot_mapper.dart';
 import 'package:reaprime/src/services/database/mappers/steam_mapper.dart';
 import 'package:reaprime/src/services/database/mappers/bean_mapper.dart';
@@ -149,6 +150,34 @@ Set<SimulatedDevicesTypes> _parseSimulateFlag(String value) {
       .map((e) => SimulatedDevicesTypesFromString.fromString(e))
       .whereType<SimulatedDevicesTypes>()
       .toSet();
+}
+
+/// Pages shots for backup export, skipping unreadable rows without letting
+/// them shrink the page below [limit] while more rows remain -- otherwise the
+/// exporter's `page.length < pageSize` end-of-stream check would truncate a
+/// backup at the first unreadable row (gh#784).
+Future<List<domain.ShotRecord>> pageShotsForExport(
+  AppDatabase appDatabase,
+  int limit, {
+  DateTime? afterTimestamp,
+  DateTime? afterCreatedAt,
+  String? afterId,
+}) async {
+  final result = <domain.ShotRecord>[];
+  DateTime? cursorTimestamp = afterTimestamp;
+  String? cursorId = afterId;
+  while (result.length < limit) {
+    final rows = await appDatabase.shotDao.getShotsForExport(
+      limit: limit - result.length,
+      cursorTimestamp: cursorTimestamp,
+      cursorId: cursorId,
+    );
+    if (rows.isEmpty) break;
+    cursorTimestamp = rows.last.timestamp;
+    cursorId = rows.last.id;
+    result.addAll(ShotMapper.fromRows(rows));
+  }
+  return result;
 }
 
 Future<void> _printStoragePaths() async {
@@ -554,14 +583,14 @@ void main(List<String> args) async {
       grinderStorage: grinderStorage,
       connectionManager: connectionManager,
       backupSources: BackupDataSources(
-        pageShots: (limit, {afterTimestamp, afterCreatedAt, afterId}) async {
-          final rows = await appDatabase.shotDao.getShotsForExport(
-            limit: limit,
-            cursorTimestamp: afterTimestamp,
-            cursorId: afterId,
-          );
-          return ShotMapper.fromRows(rows);
-        },
+        pageShots: (limit, {afterTimestamp, afterCreatedAt, afterId}) =>
+            pageShotsForExport(
+              appDatabase,
+              limit,
+              afterTimestamp: afterTimestamp,
+              afterCreatedAt: afterCreatedAt,
+              afterId: afterId,
+            ),
         pageSteams: (limit, {afterTimestamp, afterCreatedAt, afterId}) async {
           final rows = await appDatabase.steamDao.getSteamsForExport(
             limit: limit,
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -152,10 +152,6 @@ Set<SimulatedDevicesTypes> _parseSimulateFlag(String value) {
       .toSet();
 }
 
-/// Pages shots for backup export, skipping unreadable rows without letting
-/// them shrink the page below [limit] while more rows remain -- otherwise the
-/// exporter's `page.length < pageSize` end-of-stream check would truncate a
-/// backup at the first unreadable row (gh#784).
 Future<List<domain.ShotRecord>> pageShotsForExport(
   AppDatabase appDatabase,
   int limit, {
--- a/lib/src/services/database/mappers/shot_mapper.dart
+++ b/lib/src/services/database/mappers/shot_mapper.dart
@@ -1,6 +1,7 @@
 import 'dart:convert';
 
 import 'package:drift/drift.dart';
+import 'package:logging/logging.dart';
 import 'package:reaprime/src/models/data/shot_annotations.dart';
 import 'package:reaprime/src/models/data/shot_record.dart' as domain;
 import 'package:reaprime/src/models/data/shot_snapshot.dart';
@@ -8,8 +9,24 @@ import 'package:reaprime/src/models/data/workflow.dart' as domain_workflow;
 import 'package:reaprime/src/services/database/database.dart' as db;
 
 class ShotMapper {
+  static final Logger _log = Logger('ShotMapper');
+
+  static List<domain.ShotRecord> fromRows(Iterable<db.ShotRecord> rows) {
+    final records = <domain.ShotRecord>[];
+    for (final row in rows) {
+      try {
+        records.add(fromRow(row));
+      } catch (e) {
+        _log.warning('Skipping unreadable shot ${row.id}', e);
+      }
+    }
+    return records;
+  }
+
   static domain.ShotRecord fromRow(db.ShotRecord row) {
-    final workflow = domain_workflow.Workflow.fromJson(row.workflowJson);
+    final workflow = domain_workflow.Workflow.fromRecordedJson(
+      row.workflowJson,
+    );
 
     ShotAnnotations? annotations;
     if (row.annotationsJson != null) {
--- a/lib/src/util/shot_importer.dart
+++ b/lib/src/util/shot_importer.dart
@@ -22,7 +22,7 @@ class ShotImporter {
           'Expected JSON object in array, got ${item.runtimeType}',
         );
       }
-      final shot = ShotRecord.fromJson(item);
+      final shot = ShotRecord.fromRecordedJson(item);
       await storage.storeShot(shot);
       count++;
     }
@@ -37,7 +37,7 @@ class ShotImporter {
       throw FormatException('Expected JSON object, got ${json.runtimeType}');
     }
 
-    final shot = ShotRecord.fromJson(json);
+    final shot = ShotRecord.fromRecordedJson(json);
     await storage.storeShot(shot);
   }
 }
--- a/doc/AI_STORAGE_NOTES.md
+++ b/doc/AI_STORAGE_NOTES.md
@@ -81,6 +81,31 @@ Content-based hash IDs for deduplication. `ProfileController` manages the profil
 - `ProfileStorageService` interface with `DriftProfileStorageService` implementation.
 - ID-changing updates use `ProfileStorageService.replace`, which inserts the replacement and deletes the original in one Drift transaction. Target-ID collisions throw `ArgumentError` and leave both records unchanged.
 
+### Recorded Shot Profiles Are Not Executable Profiles
+
+A shot's stored `workflowJson.profile` is a historical record, not something the
+app brews from, so `ShotMapper.fromRow` reads it with
+`Workflow.fromRecordedJson`, which parses the profile via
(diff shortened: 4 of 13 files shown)


lib/src/services/database/mappers/shot_mapper.dart
```diff
@@ -1,15 +1,32 @@
...
+      } catch (e) {
+        _log.warning('Skipping unreadable shot ${row.id}', e);
+      }
+    }
+    return records;
+  }
+
   static domain.ShotRecord fromRow(db.ShotRecord row) {
-    final workflow = domain_workflow.Workflow.fromJson(row.workflowJson);
+    final workflow = domain_workflow.Workflow.fromRecordedJson(
```
tadelv, line 27: This fixes the DB read path, but `PUT /api/v1/shots/<id>` still cannot update one of these recovered de1app shots. `ShotsHandler._updateShot` starts from `existingShot.toJson()` and then calls `ShotRecord.fromJson(merged)`; `ShotRecord.fromJson` still uses strict `Workflow.fromJson`, so a profile with `steps: []` throws again even when the patch only changes annotations/metadata. Please make the stored-shot update round-trip use the recorded-workflow semantics without weakening the executable profile/workflow boundaries, and add a regression test that updates a step-less imported shot.

tadelv (changes requested):
One required fix: the new recorded-workflow read path makes imported step-less shots visible, but the existing shot update path still reparses those same records through the strict executable workflow parser. That leaves the recovered shots uneditable. Please preserve strictness at executable/write boundaries while making stored-shot updates round-trip recorded workflows, and add a regression test for updating an imported step-less shot.

lib/src/services/database/mappers/shot_mapper.dart
allofmeng: Fixed in a9156b11.

`ShotRecord.fromRecordedJson` now mirrors the existing `Workflow` / `Profile` split, and `ShotsHandler._updateShot` uses it for both the `contentChanged` comparison and the merged record it stores. `ShotRecord.fromJson` is unchanged and still strict everywhere else, so no executable profile or workflow boundary moved: `PUT /api/v1/workflow` still returns 400 for a step-less profile (issue #338), and `test/webserver/workflow_handler_test.dart` stays green.

`ShotImporter` had the same gap in the other direction: a backup export of a recovered de1app shot could be written but not read back, so its two `ShotRecord.fromJson` calls moved to `fromRecordedJson` as well.

Regression tests added to `test/database/shot_history_imported_profile_test.dart`, driving the real `ShotsHandler` over an in-memory Drift database:
- `PUT /api/v1/shots/de1app-1626149813` with an annotation patch on a `steps: []` shot returns 200, persists the note, and leaves the profile step-less;
- the same for a title-less imported shot, which reads back as `Unknown profile`.

Both fail with 500 on the previous commit and pass now. `flutter analyze` clean; `flutter test` 3866 passed with the same 8 pre-existing `webui_token_injection_test.dart` port-3000 failures reported earlier.

lib/src/util/shot_importer.dart
```diff
@@ -22,7 +22,7 @@ class ShotImporter {
           'Expected JSON object in array, got ${item.runtimeType}',
         );
       }
-      final shot = ShotRecord.fromJson(item);
+      final shot = ShotRecord.fromRecordedJson(item);
```
tadelv, line 25: This fixes `ShotImporter`, but that is not the importer used by the current backup archive path. `ShotExportSection.importJson` still does `ShotRecord.fromJson(...)`, so a `shots.json` backup containing one of these recovered `steps: []` records can be exported successfully and then rejected on restore. Please switch that stored-shot backup import path to recorded parsing as well and add a backup round-trip regression for a step-less de1app shot.

lib/main.dart
```diff
@@ -560,7 +560,7 @@ void main(List<String> args) async {
             cursorTimestamp: afterTimestamp,
             cursorId: afterId,
           );
-          return rows.map(ShotMapper.fromRow).toList();
+          return ShotMapper.fromRows(rows);
```
tadelv: `ShotExportSection.exportJson` treats `page.length < pageSize` as EOF. `getShotsForExport` returns a raw page of up to `pageSize` rows, but `ShotMapper.fromRows` can drop an unreadable row, making this callback return fewer items; the exporter then stops and silently omits every older shot. Since the known de1app rows are now readable via `fromRow`, please either keep export mapping non-filtering or preserve the raw-page cursor/size semantics when skipping rows. Add coverage so one unreadable row cannot truncate the remainder of a backup.

tadelv (changes requested):
Two required fixes remain in the backup path. The recorded-shot parser split is sound and the previous `PUT /api/v1/shots/<id>` blocker is fixed, but the current backup restore path still reparses shots through the strict parser, and the new skip-on-map export callback can truncate a paged backup. Please address both and add regression coverage for backup export/import.

lib/src/util/shot_importer.dart
allofmeng: Fixed in b7fd4718.

`ShotExportSection.importJson` (the restore side of `/backup`, not `ShotImporter`) was the remaining strict path -- switched its `ShotRecord.fromJson` to `ShotRecord.fromRecordedJson`, matching `ShotImporter` and the shot-update path. `ShotRecord.fromJson` stays strict everywhere else.

Regression: `test/data_export/shot_export_section_test.dart` — "imports a recorded shot with a step-less imported profile (gh#784)", driving `ShotExportSection.importJson` directly with a de1app-style step-less-profile record.

lib/main.dart
allofmeng: Fixed in b7fd4718.

Extracted the `pageShots` wiring into a standalone `pageShotsForExport` (`lib/main.dart`). It re-queries past a row `ShotMapper.fromRows` drops, advancing the cursor from the raw row (not the last successfully mapped one), until it either fills the requested page or the underlying table is genuinely exhausted. `ShotExportSection.exportJson`'s `page.length < pageSize` end-of-stream check is unchanged -- it now only sees short pages that are real EOF.

Regression: `test/database/shot_history_imported_profile_test.dart` — "an unreadable row does not truncate the rest of a paged backup" (`backup export (gh#784)` group). It drives the real `pageShotsForExport` (via `import 'package:reaprime/main.dart' as app;`) against a real Drift DB with `pageSize: 2` and an unmappable row placed inside what would otherwise be the first page, and asserts all three good shots are exported. I confirmed this test fails (truncates to one shot) against the pre-fix version of the callback.

lib/main.dart
```diff
@@ -151,6 +152,34 @@ Set<SimulatedDevicesTypes> _parseSimulateFlag(String value) {
       .toSet();
 }
 
+/// Pages shots for backup export, skipping unreadable rows without letting
```
tadelv: `AGENTS.md` explicitly says not to add explanatory comments to new/substantially rewritten code and to keep rationale/debugging history in the matching `doc/AI_*_NOTES.md`. This block is exactly that rationale, and the same explanation is already captured in `doc/AI_STORAGE_NOTES.md`. Please remove this doc comment; `pageShotsForExport` plus its regression test are clear enough without it.

tadelv (changes requested):
The two substantive blockers from the previous round are fixed: `/backup` restore now uses recorded-shot parsing, and backup export now preserves page semantics across skipped rows with regression coverage. CI on b7fd4718 is green. One required repository-guidance cleanup remains in `lib/main.dart`; otherwise this looks ready.

lib/main.dart
allofmeng: Fixed in f1f3d4ba. Removed the doc comment -- doc/AI_STORAGE_NOTES.md already carries the rationale.

tadelv (approved):
Re-reviewed at f1f3d4ba. The final AGENTS.md cleanup is fixed, the previous functional blockers remain addressed with regression coverage, and current PR checks are green. No remaining merge-blocking issues found.

PR #790, opened 2026-09-05
fix(webui): give each skin a stable origin, so its browser storage survives

ChampionDesigns:
## Summary

A skin's browser storage was wiped at every app start, because its **origin changed every time**.

`_serveFresh` bound the skin server on port 0 and refused to reuse a port. Browser storage is
keyed by origin, so every setting a skin had written was gone after a restart. `:3000` is only an
entry point that 307-redirects to whichever ephemeral port was chosen that run.

Measured on the tablet, 28 Aug 2026: the same skin was served at 41825, then 39055, then 45939
across three restarts, and both keys written before a restart were absent after it. **Every skin is
affected, not one.**

Base: `main`. Independent.

### Design notes

A skin now gets a **stable port derived from its identity** by FNV-1a, persisted in
`shared_preferences` under `webUISkinPorts`.

- Range 24800 + 4096, with 3000, 4001 and 8080 reserved, so a skin can never land on the entry
  point, the `:644` listener or the API.
- Three bind attempts with a 150 ms retry, then eight temporary-port attempts as a fallback. **A
  taken port degrades to today's behaviour rather than failing the boot.**
- `loadSkinPortAssignments` / `saveSkinPortAssignments` are injected, so `WebUIService` keeps no
  ambient dependency and the tests drive it directly.

The KV layer was never affected — `store/<skin>/*` lives in the app, not the browser — which is
why this looked like a skin bug for so long.

## Linked Issue

N/A
## Verification

- `flutter analyze` — clean.
- `flutter test` — **full suite 3906 passed / 1 skipped**, run against current `main` on
  5 Sep 2026.
- `flutter test` — `test/webui_support` 41 passed, including the new
  `webui_stable_origin_test.dart` and `skin_port_assignments_test.dart`.
- **Verified on hardware.** Measured on the tablet before the fix: the same skin served at 41825,
  then 39055, then 45939 across three restarts, with its stored keys gone each time. The fix ships
  in the Decaid-Canary build Ben runs, and skin settings now survive a restart.

## Impact

- **User-visible**: skin settings stop disappearing on restart. This is the fix for the screensaver
  bug, and it applies to every skin.
- **Compatibility**: additive. A skin with no stored assignment gets one on first serve.
- **Migration**: none. Settings lost before this change are already gone; nothing is recoverable.
- **Security**: none. The port is derived from the skin's own identity, not from user input.

## Contributor Responsibility

### DIFF
--- a/doc/Skins.md
+++ b/doc/Skins.md
@@ -64,6 +64,29 @@ backgrounded. After ten minutes in the background, it unloads the page and
 reloads the selected skin when the app returns. Skin state that must survive
 this reload should be persisted through the Decaid API or browser storage.
 
+### Skin Origins and Browser Storage
+
+Each installed skin is served from its own **stable origin** — a port derived
+from the skin's identity and remembered across restarts. Browser storage is
+keyed by origin, so `localStorage`, `sessionStorage` and IndexedDB written by a
+skin are still there the next time the app starts.
+
+Ports are assigned from 24800 upward. 3000, 4001 and 8080 are reserved, so a
+skin never lands on the entry point, the local listener or the REST API.
+`localhost:3000` remains the entry point and redirects to the skin's own
+origin.
+
+**The fallback, and what it costs.** If a skin's assigned port cannot be bound
+— another process holds it — Decaid retries briefly, then falls back to a
+**temporary origin** for that run. The skin loads and works normally, but
+because the origin differs, **browser storage written under the stable origin
+is not visible**, and anything written during that run is not visible after it.
+Nothing is deleted; it is simply keyed to a different origin.
+
+A skin that must not lose state across such a run should persist it through the
+Decaid API rather than browser storage. The KV endpoints are unaffected by
+origin, because that data lives in the app.
+
 ### Offline Operation
 
 Decaid serves installed skins from `localhost`, so Wi-Fi and internet access
--- a/doc/Skins.md
+++ b/doc/Skins.md
@@ -67,9 +67,13 @@ this reload should be persisted through the Decaid API or browser storage.
 ### Skin Origins and Browser Storage
 
 Each installed skin is served from its own **stable origin** — a port derived
-from the skin's identity and remembered across restarts. Browser storage is
-keyed by origin, so `localStorage`, `sessionStorage` and IndexedDB written by a
-skin are still there the next time the app starts.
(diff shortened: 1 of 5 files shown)


tadelv (changes requested):
Required docs are missing. This PR changes embedded-skin behavior by making the served origin stable across app restarts, but `doc/Skins.md` is unchanged. Both `AGENTS.md` (Pre-Commit / Pre-PR checklist) and `CONTRIBUTING.md` (Guardrails → Spec & Docs) explicitly require `doc/Skins.md` to be updated when skin behavior changes. Please document the stable-origin/browser-storage behavior, including the temporary-origin fallback when the assigned port cannot be bound.

doc/Skins.md
```diff
@@ -64,6 +64,29 @@ backgrounded. After ten minutes in the background, it unloads the page and
 reloads the selected skin when the app returns. Skin state that must survive
 this reload should be persisted through the Decaid API or browser storage.
 
+### Skin Origins and Browser Storage
+
+Each installed skin is served from its own **stable origin** — a port derived
+from the skin's identity and remembered across restarts. Browser storage is
+keyed by origin, so `localStorage`, `sessionStorage` and IndexedDB written by a
+skin are still there the next time the app starts.
+
```
tadelv: `sessionStorage` should not be included in the restart-persistence guarantee. It is page-session scoped, unlike `localStorage` and IndexedDB. Please document only the persistent storage mechanisms here.

tadelv (changes requested):
The original documentation blocker is addressed, but the new text has one correctness issue that should be fixed before merge: it says `sessionStorage` survives until the next app start. `sessionStorage` is scoped to a page session and should not be documented as restart-persistent. Please limit the persistence guarantee here to persistent browser storage such as `localStorage` and IndexedDB (and adjust the later generic "browser storage" wording if needed so it does not imply `sessionStorage` survives app restarts).

tadelv (approved):
The requested documentation fixes are resolved. `doc/Skins.md` now correctly limits restart persistence to persistent browser storage (`localStorage` and IndexedDB), explicitly notes that `sessionStorage` does not survive page unload/app restart, and still documents the temporary-origin fallback. Current PR-head CI is green. I found no remaining merge-blocking issues.

PR #791, opened 2026-09-05
fix(api): refuse an explicit null targetYield on PUT /api/v1/workflow

ChampionDesigns:
## Summary

You were right, and the repository backs you up, so this branch now does the
small thing instead of the behaviour change.

`BengleSawBridge` read the target as `targetYield ?? 0.0`, and this branch made
an absent target skip the firmware write. Three things make that wrong:

1. **It reverses an accepted design.**
   `doc/plans/archive/bengle-saw-mmr/2026-05-18-bengle-saw-mmr.md` specifies the
   connect-side re-apply as `currentWorkflow.context?.targetYield ?? 0.0`, and
   states `0.0` = SAW off, mirroring the cup warmer's `0.0` = mat off. The
   earlier version of this PR described that as an oversight. It is not.
2. **It split the app in two.** `De1StateManager` still passes
   `context?.targetYield ?? 0` to `ShotSequencer`, and `ShotSequencer`
   suppresses its own weight stop when the machine stops autonomously. So the
   app believed there was no weight target while the firmware still held one,
   and nothing corrected it. The machine would stop at a weight no current
   workflow asked for.
3. **"Leave the machine alone" needs another owner, and there is none.** The
   same design doc chose `targetYield` as the single source of truth
   deliberately, "rather than inventing a `POST /machine/saw` endpoint", so the
   bridge is the only writer.

So the write guard is reverted and the ambiguity is closed at the edge instead,
which is what you asked for.

### What this PR does now

`PUT /api/v1/workflow` refuses an explicit null on `context.targetYield` with
`400`, through `rejectExplicitNulls` — the same helper the same function already
applies to `steamSettings` and its four fields. Three lines of production code.

The PUT is a deep merge, so omitting the field still keeps the current value.
Only an explicit null is refused, and `0` still means stop-at-weight off.

`assets/api/rest_v1.yml` and `doc/Api.md` now state that null and `0` both mean
the feature is off, and that the field is the single source of truth for
stop-at-weight. That sentence is the ambiguity this branch existed to remove.

### On "make targetYield a mandatory field"

I did not mark it `required`. `WorkflowContext` documents "All fields are
nullable for flexible usage — from minimal (dose only) to full", and a required
field would break every partial PUT and every stored workflow without a yield.
Refusing an explicit null gives the strictness at the edge without that cost.
Say so if you did mean `required` and I will change it.

## Linked Issue

N/A

## Verification

- `flutter analyze` — clean.
- `flutter test` — full suite **3894 passed / 1 skipped / 0 failed**.
- `dart format --output=none --set-exit-if-changed` — clean on every changed
  file, run with Dart 3.13.3, the version CI resolves on `stable`.
- The refusal test is a real regression test: with the three production lines
  removed it returns `200` instead of `400`.
- New coverage in `test/webserver/workflow_handler_test.dart`: an explicit `0`
  still lands as `0`, an explicit `null` is refused with `400` and leaves the
  document unchanged, and a PUT that omits the field keeps the current value.

## Impact

- **User-visible**: none. The bridge behaves exactly as it did before this
  branch, which is what the accepted design specifies.
- **API**: `PUT /api/v1/workflow` returns `400` for an explicit
  `context.targetYield: null`, where it previously accepted it and dropped the
  field. Omitting the field is unchanged, and `0` is unchanged.
- **Compatibility**: a client that never sends an explicit null is unaffected.
- **Security**: none.

### Removed from this branch

`GET /api/v1/machine/stopAtWeight` moved to its own PR. It is unrelated to the
null question, and it was added here without a `rest_v1.yml` or `doc/Api.md`
entry, which `CONTRIBUTING.md` makes a hard gate. It has both there. The six
bridge tests that pinned the skip behaviour went with it.

## Contributor Responsibility


assets/api/rest_v1.yml
```diff
@@ -7685,7 +7685,15 @@ components:
           type: number
           format: double
           nullable: true
-          description: Target output weight in grams (beverage)
+          description: |
```
tadelv, line 7792: Blocking contract mismatch: this description says `PUT /api/v1/workflow` rejects an explicit null, but `WorkflowPatch.context` still references `WorkflowContext`, where `targetYield` is declared `nullable: true`. Since `rest_v1.yml` is authoritative, OpenAPI clients/codegen still see `{ "context": { "targetYield": null } }` as valid even though the handler returns 400. Please model the PUT semantics in the schema itself—e.g. a patch-specific context schema analogous to `SteamSettingsPatch`, with `targetYield` optional but non-nullable—while retaining nullable `WorkflowContext.targetYield` for returned/stored workflows if needed. The wording should also distinguish workflow-model nullability from PUT-patch nullability.

tadelv (changes requested):
This is very close, and the implementation direction now matches the accepted SAW design. I found two final contract/guidance issues that should be fixed before merge:

1. The authoritative OpenAPI request schema still permits the exact explicit-null payload the handler now rejects. Please make the PUT patch schema express the new rule, while preserving nullable/optional workflow data where that remains valid.
2. `CONTRIBUTING.md` requires an end-to-end regression scenario for API surface changes. This PR changes `PUT /api/v1/workflow` from accepting an explicit `context.targetYield: null` to returning 400. Please extend the existing Bengle integrated-scale/workflow scenario (or an equivalent scenario) to assert explicit null -> 400 and explicit 0 -> accepted.

The focused handler tests are otherwise appropriate, CI is green, and I did not find another merge-blocking implementation issue.

tadelv (approved):
Re-reviewed the updated PR. The previous OpenAPI contract mismatch is resolved with a patch-specific `WorkflowContextPatch`, stored/returned `WorkflowContext` remains nullable, and the API regression scenario now covers explicit-null rejection and explicit `0`. The added whole-context/null and type guards are consistent with the same contract and close direct bypasses rather than expanding the feature. I found no remaining merge-blocking issues. CI for the latest head is still running, so merge should remain gated on that completing successfully.

PR #792, opened 2026-09-05
fix(update): a failed update check must say so, not look like "up to date"

ChampionDesigns:
## Summary

A failed update check was indistinguishable from "you are up to date". Three ways it could fail
silently, all fixed here.

- `android_updater.dart` logged a warning and returned `null` when the release fetch failed. It now
  throws a typed `UpdateCheckException` carrying the reason.
- `update_check_service.dart` called `checkForUpdate()` from two places with no guard, so a failure
  escaped as an unhandled error. Both sites now catch and log; the error state was already emitted
  by then.
- `UpdateHandler` answered nothing useful on a build that cannot check. It now replies with an
  explicit "not supported on this build" plus the releases URL, gated on the new
  `UpdateCheckService.canCheck`.

`canCheck` is `!isMacOS && !externallyManaged`. The second clause was added in review: upstream's
externally managed builds (App Store / TestFlight) no-op inside `checkForUpdate()` and emit an
`idle` frame, which reads on the wire as "checked, you are up to date" — the same defect as above,
reached by a different route.

Base: `main`. Independent.

## Linked Issue

N/A
## Verification

- `flutter analyze` — clean.
- `dart format` — clean on every changed `.dart` file, run on Dart 3.13.3, the version CI uses.
- `flutter test` — the focused files pass: `update_check_service`, `android_updater_download`,
  `update_handler` and `app_update_state`.
- **Red proof.** With `&& !externallyManaged` reverted out of `canCheck`, exactly two tests fail
  and nothing else: the service test, and the handler test with `Bad state: No element` — no reply
  at all, the reviewed defect demonstrated at the handler tier. Restoring the one line returns
  them to green.

**End-to-end, run on this head.** `.agents/skills/decent-app/scenarios/update-check-capability.md`
splits the two `canCheck` clauses, because either alone is enough to refuse and one recipe cannot
test both.

*Part A — the ownership clause, Linux host, `--dart-define APP_STORE=true`:*

```json
{"error":"App update checks are not supported on this build",
 "url":"https://github.com/decentespresso/decaid/releases"}
```

`GET /api/v1/update` unchanged: `{"phase":"idle","latestVersion":null,"installable":false}`.

*Control, same host, ordinary build, same command:*

```json
{"phase":"idle","currentVersion":"0.8.6-beta.1"}
{"phase":"checking","currentVersion":"0.8.6-beta.1"}
{"phase":"idle","currentVersion":"0.8.6-beta.1"}
```

A state frame, not a refusal — same box, same command, opposite outcome, so the refusal is the
ownership clause and not the host.

*Part B — the platform clause* requires a macOS host and was **not run**. It covers pre-existing
`!isMacOS` behaviour that this PR does not change, and it is pinned at the unit tier.

- **On hardware, scoped.** The first three fixes above ship in the Decaid-Canary build on Ben's
  machine — they reached that fork on 30 Aug 2026 and the tablet has run them since v0.8.31
  (1 Sep). The `externallyManaged` clause added in review is **not** on that build; its evidence
  is the scenario run and the tests above.

## Impact

- **User-visible**: a failed check now says it failed. Previously it looked like "no update
  available", which is the same thing a user sees when genuinely up to date.
- **API**: `UpdateCheckException` is new and thrown where `null` was returned. A caller that
  ignored the null result now sees an exception — the two call sites in this repo are both updated.
- **Compatibility**: `canCheck` is additive. A build that cannot check now receives an explicit
  refusal where it previously received an `idle` state frame or, on macOS, nothing at all.
- **Documentation**: the `check` command's AsyncAPI description and `doc/Api.md` carry the
  ownership case, and a new end-to-end scenario covers the refusals.
- **Security**: none.

## Contributor Responsibility


lib/src/util/github_archive.dart
```diff
@@ -21,7 +21,7 @@ class GitHubRelease {
 
 const _headers = {
   'Accept': 'application/vnd.github.v3+json',
-  'User-Agent': 'Decaid',
+  'User-Agent': 'Decide',
```
tadelv: This unrelated hunk changes the shared GitHub `User-Agent` from the project name `Decaid` to `Decide`. `AGENTS.md` explicitly defines the user-facing/project naming as **Decaid**, and this PR has no reason to alter the archive helper's identity. Please revert this line to `Decaid`.

lib/src/services/webserver/update_handler.dart
```diff
@@ -58,6 +58,13 @@ class UpdateHandler {
 
     switch (command) {
       case 'check':
+        if (!_service.canCheck) {
```
tadelv, line 61: This changes the wire contract of `/ws/v1/update`: on macOS, `{"command":"check"}` now returns a transient `{error, url}` instead of behaving as the existing no-op check. The repo's `AGENTS.md` / `doc/AI_API_NOTES.md` hard rule requires `assets/api/websocket_v1.yml` to be updated in the same commit as endpoint behavior changes. The current spec still describes `check` as an unconditional GitHub re-poll and only explicitly documents `install` as unsupported on other platforms. Please update the AsyncAPI description/contract to match this behavior (and run the prescribed WebSocket smoke check).

tadelv (changes requested):
Two required fixes before merge: keep the shared GitHub User-Agent as `Decaid`, and update the authoritative WebSocket spec for the new macOS `check` command behavior. The implementation/tests otherwise look coherent, and CI is green.

lib/src/util/github_archive.dart
ChampionDesigns: Reverted — the shared `User-Agent` is `Decaid` again. That hunk was unrelated to this PR and
should not have been in it.

lib/src/services/webserver/update_handler.dart
ChampionDesigns: Done — `assets/api/websocket_v1.yml` now documents the macOS `check` reply, so the spec matches
what the endpoint actually returns rather than describing an unconditional re-poll.

tadelv (changes requested):
The two code/spec findings from my previous pass are fixed: the unrelated User-Agent change is gone, and `assets/api/websocket_v1.yml` now matches the macOS `check` behavior. I found no new implementation blocker in the updated diff, and current-head CI is green.

One required repository-guidance item remains from the existing WebSocket review thread: there is still no evidence of the prescribed runtime smoke check for the changed `/ws/v1/update` behavior. `AGENTS.md` requires API/spec changes to be exercised via `scripts/sb-dev.sh` + `websocat`/`curl`, and `.agents/skills/decent-app/verification.md` explicitly says runtime-facing changes need the relevant `sb-dev` smoke before completion. The follow-up commit records analyze/tests/format/spec parsing, but not that endpoint smoke. Please run the affected `/ws/v1/update` flow against a running app and add the result to the PR verification. Once that is recorded, I don't see another required change.

lib/src/services/update_check_service.dart
```diff
@@ -60,6 +60,8 @@ class UpdateCheckService {
 
   bool get canInstall => _isAndroid;
 
+  bool get canCheck => !_isMacOS;
```
tadelv: After the upstream merge this capability predicate is incomplete. `checkForUpdate()` now immediately no-ops when `externallyManaged`, and App Store/TestFlight builds are built with `APP_STORE=true`; on those builds this still returns `true`, so `/ws/v1/update` accepts `{"command":"check"}` and then silently settles/returns instead of sending the explicit unsupported reply this PR introduces. Please make `canCheck` false for externally managed builds as well (for example `!_isMacOS && !externallyManaged`), update the AsyncAPI description to cover that ownership case, and add service/handler tests for it.

tadelv (changes requested):
The requested runtime smoke is now sufficient, and I’ve resolved the prior WebSocket/spec thread. Current-head CI is green.

One new blocker appeared when this branch merged the newer `main`: upstream now has `externallyManaged` app builds whose `checkForUpdate()` intentionally returns without polling. `canCheck` still only excludes macOS, so an App Store/TestFlight build reports itself as check-capable; `/ws/v1/update` accepts `{"command":"check"}`, calls `requestCheck()`, and produces no command-level refusal — effectively recreating the silent unsupported-check behavior this PR is fixing.

Please fold `externallyManaged` into the capability gate, update the AsyncAPI wording accordingly, and add service/handler coverage for that case. I don’t see another blocker beyond this upstream-integration fix.

.agents/skills/decent-app/scenarios/update-check-capability.md
```diff
@@ -0,0 +1,90 @@
...
+built with `--dart-define APP_STORE=true`) and macOS builds both refuse the
+command with a direct `{"error", "url"}` reply instead of polling GitHub.
+
+The refusal is the point: a check that cannot run must say so, rather than
+settle back to `idle` and read as "up to date".
+
+## Preconditions
+
+```bash
+scripts/sb-dev.sh start --connect-machine MockDe1 --connect-scale MockScale \
```
tadelv, line 31: This scenario is supposed to pin the new `externallyManaged` capability gate, but without an explicit non-macOS target it does not necessarily test that clause. On a macOS host, `canCheck` is already false because of `_isMacOS`, so deleting `&& !externallyManaged` would still make this recipe pass. Please make this precondition require a non-macOS target (for example Linux where available), or split platform-ownership and externally-managed cases so each gate can fail independently. Then run the managed-build case on this head and add the result to the PR verification; the earlier runtime smoke used an ordinary build before this fix.

tadelv (changes requested):
The externally-managed capability bug from my previous pass is fixed correctly: `canCheck` now matches `checkForUpdate()`, the AsyncAPI and `doc/Api.md` cover the ownership case, the service/handler tests pin it, and current-head CI is green against current `main`. I resolved that old thread.

One verification blocker remains in the new follow-up. The newly added E2E scenario is meant to catch regression of the `externallyManaged` clause, but its precondition does not require a non-macOS target. On macOS the scenario still passes if `!externallyManaged` is removed, because `!_isMacOS` independently refuses the command. That makes the persistent regression recipe unable to distinguish the new ownership gate from the older platform gate on the most obvious developer host. The latest PR comment also reports focused tests but not a run of this new scenario; the earlier `sb-dev` smoke predates the externally-managed fix and used an ordinary Linux build.

Please make the scenario explicitly exercise `APP_STORE=true` on a non-macOS target (or split the macOS and externally-managed cases so each clause is independently testable), then run that managed-build scenario on this head and record the result. I found no additional production-code blocker beyond that verification issue.

tadelv (approved):
Re-reviewed the latest head adversarially. The externally-managed capability gate, API/spec/docs updates, unit/handler coverage, and the split E2E regression scenario are now coherent. Part A explicitly isolates `externallyManaged` on a non-macOS host and has been run on this head with the same-host ordinary-build control; the previous verification thread is therefore resolved. Current-head CI is green. `main` has advanced since that run, but the intervening commits do not touch this PR's update/API/scenario surface and the PR remains mergeable. I found no remaining merge blocker.

PR #794, opened 2026-09-05
fix(api): PUT /machine/ledStrip echoes the palette the machine stored

ChampionDesigns:
## Summary

`PUT /api/v1/machine/ledStrip` answered a bare `{"status":"accepted"}`, so a client had no way to
learn how its colours were actually stored. The firmware **canonicalises** what it is given — a
replicated-byte spelling like `FFFF22220000` comes back as `FF0022000000` — and a following `GET`
returned that canonical form. So the PUT response and the GET disagreed about the same write.

The handler now reads the stored state back and returns it. **A PUT body and the following GET body
are byte-identical.**

Base: `main`. Independent.

### Design notes

`MockReplayDe1.setLedStrip` stored whatever it was handed, so the replay mock could hold a spelling
the machine never would. It now canonicalises too. `Color16.quantized()` and
`ZoneLedState.quantized()` hoist that arithmetic onto the model, where both the mock and any future
caller reach one copy instead of keeping their own inline.

Same class of defect as a mock that does not clamp: **a mock more permissive than the hardware hides
the bug it exists to expose.**

## Linked Issue

N/A
## Verification

- `flutter analyze` — clean.
- `flutter test` — **full suite 3897 passed / 1 skipped**, run against current `main` on
  5 Sep 2026.
- `flutter test` — `de1handler_led_strip_test` 18 passed, up from upstream's 12, plus the replay
  tests: 37 across both files.
- The four new cases pin what upstream did not cover: replicated-byte input stored and echoed
  canonically, the audited F-044 write on `frontStrip.awake`, canonical input echoed
  byte-identically, and a reset after a PUT returning the same canonical body.

## Impact

- **API**: the `PUT /machine/ledStrip` 200 body changes from `{"status":"accepted"}` to the stored
  palette. A client that ignored the body is unaffected; a client that parsed it for
  `status == accepted` must be updated.
- **Compatibility**: this is the one breaking-shaped change in the set, and it is the point of the
  PR. The old body carried no information.
- **User-visible**: none directly.
- **Security**: none.

## Contributor Responsibility

### DIFF
--- a/lib/src/models/device/impl/replay/mock_replay_de1.dart
+++ b/lib/src/models/device/impl/replay/mock_replay_de1.dart
@@ -298,7 +298,16 @@ class MockReplayDe1 implements BengleInterface, SimulatedDevice {
   @override
   Future<LedStripState> getLedStripState() async => _led.value;
   @override
-  Future<void> setLedStrip(LedStripState state) async => _led.add(state);
+  Future<void> setLedStrip(LedStripState state) async => _led.add(
+    // The firmware canonicalises what it is given, so a replay mock that
+    // stored the raw spelling would answer a GET with a value the machine
+    // could never hold.
+    LedStripState(
+      frontStrip: state.frontStrip.quantized(),
+      backStrip: state.backStrip.quantized(),
+      frontSwitch: state.frontSwitch,
+    ),
+  );
   @override
   Future<void> commitLedStrip() async {}
   @override
--- a/lib/src/models/device/impl/replay/mock_replay_de1.dart
+++ b/lib/src/models/device/impl/replay/mock_replay_de1.dart
@@ -291,23 +291,15 @@ class MockReplayDe1 implements BengleInterface, SimulatedDevice {
   }) async {}
 
   final BehaviorSubject<LedStripState> _led = BehaviorSubject.seeded(
-    const LedStripState(),
+    const LedStripState().canonical(),
   );
   @override
   Stream<LedStripState> get ledStripState => _led.stream;
   @override
   Future<LedStripState> getLedStripState() async => _led.value;
   @override
-  Future<void> setLedStrip(LedStripState state) async => _led.add(
-    // The firmware canonicalises what it is given, so a replay mock that
-    // stored the raw spelling would answer a GET with a value the machine
-    // could never hold.
-    LedStripState(
-      frontStrip: state.frontStrip.quantized(),
-      backStrip: state.backStrip.quantized(),
-      frontSwitch: state.frontSwitch,
-    ),
-  );
+  Future<void> setLedStrip(LedStripState state) async =>
+      _led.add(state.canonical());
   @override
   Future<void> commitLedStrip() async {}
   @override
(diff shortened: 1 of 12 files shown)


lib/src/services/webserver/de1handler.dart
```diff
@@ -244,7 +244,8 @@ class De1Handler {
         final gate = _bengleFirmwareGate(de1, 'ledStrip');
         if (gate != null) return gate;
         await (de1 as BengleInterface).setLedStrip(state);
-        return jsonOk({'status': 'accepted'});
+        final stored = await de1.getLedStripState();
+        return jsonOk(stored?.toJson() ?? {'status': 'accepted'});
```
tadelv, line 248: This changes the `PUT /api/v1/machine/ledStrip` 200 body from the existing acknowledgement to `LedStripState`, but the PR does not update `assets/api/rest_v1.yml` or `doc/Api.md`. Current `AGENTS.md` / `doc/AI_API_NOTES.md` require the REST spec in the same commit for endpoint changes, and this user-facing response change also needs the API reference updated. Please document the new 200 schema/semantics, including that the response is the canonical stored palette and the compatibility impact.

lib/src/models/device/impl/replay/mock_replay_de1.dart
```diff
@@ -298,7 +298,16 @@ class MockReplayDe1 implements BengleInterface, SimulatedDevice {
...
   @override
-  Future<void> setLedStrip(LedStripState state) async => _led.add(state);
+  Future<void> setLedStrip(LedStripState state) async => _led.add(
+    // The firmware canonicalises what it is given, so a replay mock that
+    // stored the raw spelling would answer a GET with a value the machine
+    // could never hold.
+    LedStripState(
+      frontStrip: state.frontStrip.quantized(),
+      backStrip: state.backStrip.quantized(),
+      frontSwitch: state.frontSwitch,
```
tadelv: Real Bengle writes ignore `frontSwitch` and derive it from the quantized `frontStrip` (including the firmware default switch palette when the strip is black). Preserving `state.frontSwitch` here means the replay mock can still store a state the machine cannot produce—the exact class of defect this change is intended to eliminate. Please derive it with the same semantics and add a replay assertion for this behavior. Also remove the new explanatory comment above this block; current `AGENTS.md` requires new code to be self-explanatory and hardware rationale to live in the matching AI notes instead.

tadelv (changes requested):
Requesting changes for two merge-blocking issues: the REST response contract is changed without updating the authoritative API spec/docs, and the replay mock still does not reproduce the Bengle `frontSwitch` write semantics. CI is green and I did not find other non-edge-case blockers.

tadelv (approved):
Re-reviewed the updated head. The two previous blockers are resolved: the PUT response contract is documented in the REST spec/API reference, and LED canonicalization/switch derivation is now shared across the real capability and both mocks with replay coverage. CI is green. I found no remaining non-edge-case blockers.

PR #795, opened 2026-09-05
fix(profiles): accept a limiter with no range, and refuse one with no value

ChampionDesigns:
## Summary

`StepLimiter.fromJson` defaulted a missing `value`, which turns a garbage limiter into a **silent
OFF on the machine**. It now refuses visibly — a `FormatException`, which the REST boundary reports
as a 400.

`range` becomes genuinely optional. Absent or null means a **hard cap at `value`**, encoded as
range 0 — which is the DE1 wire's own encoding, and what a client arming a cap actually sends.

Base: `main`. Independent.

### Design notes

`value` is required and `range` is not, deliberately: the same policy `TransitionType.hold` and
`ExitType.power` already follow above it in the same file. A limiter with no value is not a limiter.

A rangeless limiter hashes identically to one with an explicit `range: 0`, so profile identity is
unchanged for anything already stored.

**Two tests were deliberately left out.** They carried the limiter on a `ProfileStepPower`, which
is added by the profile-modes PR and does not exist on `main`. One duplicated the flow-step test
beside it; the other tested the **power step's** mandatory-limiter rule, not the limiter parser.
They should be re-added once both PRs land.

## Linked Issue

N/A
## Verification

- `flutter analyze` — clean.
- `flutter test` — **full suite 3914 passed / 1 skipped**, run against current `main` on
  5 Sep 2026.
- `flutter test` — `step_limiter_test` 18 passed; `test/models` and `test/services/webserver`
  506 passed.
- Fixtures `test/fixtures/f048/e01_post_body.json` and `e01_put_body.json` pin the REST shape.

## Impact

- **API**: a malformed limiter is now a 400 where it was silently accepted and disabled the
  limiter. `assets/api/rest_v1.yml` documents the rule.
- **Compatibility**: a client sending `{"value": x}` with no range now gets a hard cap, which is
  what it meant. A client sending a limiter with no `value` now gets a 400 — that request never did
  what it looked like.
- **User-visible**: a profile with a malformed limiter is refused at upload instead of running with
  the limiter off.
- **Security**: none.

## Contributor Responsibility


assets/api/rest_v1.yml
```diff
@@ -6109,7 +6108,12 @@ components:
         range:
           type: number
           format: double
-          description: Acceptable range around the limiter value
+          default: 0
+          description: >-
```
tadelv, line 6185: The description says explicit `null` is accepted, and the parser/tests do accept it, but this file is OpenAPI 3.0.3. Removing `range` from `required` only permits omission; it does not permit JSON `null` for a `type: number` property. Please add `nullable: true` (or stop documenting/accepting null) so the published contract matches runtime behavior.

lib/src/models/data/profile.dart
```diff
@@ -150,10 +150,26 @@ class StepLimiter extends Equatable {
   const StepLimiter({required this.value, required this.range});
 
   factory StepLimiter.fromJson(Map<String, dynamic> json) {
-    return StepLimiter(
-      value: parseDouble(json["value"]),
-      range: parseDouble(json["range"]),
-    );
+    // `value` is required: defaulting it would turn a garbage limiter into a
```
tadelv: Current `AGENTS.md` explicitly says not to add explanatory comments to new/substantially rewritten code and to put rationale, hardware constraints, and debugging history in the matching `doc/AI_*_NOTES.md`. Please keep the validation readable through names/control flow and move/remove this rationale; the same guidance applies to the long explanatory block immediately below and the new test-file header.

tadelv (changes requested):
Two required fixes before merge:

1. The OpenAPI contract says an explicit `range: null` is accepted, but in OpenAPI 3.0.3 an optional numeric property is still non-nullable unless `nullable: true` is set. Please make the schema match the implemented/request-tested behavior.
2. Current `AGENTS.md` says not to add explanatory comments to new/substantially rewritten code and to keep rationale/hardware/debug history in the matching `doc/AI_*_NOTES.md`. Please remove/move the new explanatory/debug-history comments (including the test-file header where applicable) while keeping the code/tests self-explanatory.

The behavioral direction otherwise looks sound: value-only limiters canonicalize to range 0, valueless limiters fail visibly, POST/PUT behavior is aligned, hash compatibility is covered, and CI is green.

doc/AI_API_NOTES.md
```diff
@@ -148,6 +148,27 @@ All four machine sockets: `/ws/v1/machine/snapshot`, `/ws/v1/machine/shotSetting
...
+wire's own encoding of "no band" and is what the mock's hard-clamp branch already
+implements. Real libraries carry steps whose limiter has no range, and a client
+arming one sends `{"value": x}` with no range at all. A response never carries a
+null range, because an omitted one is stored as 0.
+
+`rest_v1.yml` marks `range` `nullable: true`. In OpenAPI 3.0.3 dropping a
+property from `required` permits omission only; without `nullable` the published
+contract would still refuse the explicit null the parser accepts.
+
+`POST /api/v1/profile` used to answer such a body with a 500 while `PUT` answered
```
tadelv: This endpoint path is singular here, but the actual route is `POST /api/v1/profiles`. Please correct the note so the API guidance does not point readers at a nonexistent endpoint.

tadelv (changes requested):
The two previous blockers are resolved: the OpenAPI schema now permits explicit null via `nullable: true`, and the implementation/test rationale was moved out of code into `doc/AI_API_NOTES.md`. Current CI is green.

One small documentation fix remains before merge: the new API note says `POST /api/v1/profile`, but the implemented endpoint is `POST /api/v1/profiles`. Please correct that path. I did not find any other merge-blocking issues in the updated diff.

PR #796, opened 2026-09-05
docs(plugins): the http endpoint contract, and the query pass-through pinned

ChampionDesigns:
## Summary

Two things about the plugin HTTP endpoint contract that were true but undocumented, plus the tests
that pin them.

- `doc/Plugins.md` gains a **Serving HTTP Endpoints** section: how an `api` entry of
  `"type": "http"` exposes a plugin at `/api/v1/plugins/:id/:endpoint`, and what the `request`
  object carries.
- It names the `handleHttpRequest` alias. A method of that name on the object `createPlugin`
  returns works exactly like `__httpRequestHandler`, because the loader aliases it at load
  (`plugin_manager.dart:1978`). That has always been true and was never written down.
- `plugins_handler_test.dart` gains nine assertions pinning the query-string pass-through.

Base: `main`. Independent. Documentation and tests only — no production change.

## Linked Issue

N/A
## Verification

**Update, 6 Sep 2026 — the `return` contract, as you asked.** You were right that
`return=/skin/history` was not a usable contract: a plugin page runs on the API origin, so that
path resolves there rather than on the skin server, and accepting an absolute URL instead would be
an open redirect.

`doc/Plugins.md` now gives `return` its own section and states the rule: require a **skin-local
path** — one leading slash, no `//`, no scheme, nothing else accepted; read the skin origin from
`GET /api/v1/webui/server/status`, which already answers `{serving, path, port, ip}` and is already
in `rest_v1.yml`; then join the two with `new URL(path, skinOrigin)`, which cannot escape that
origin once the path is known to be skin-local. The worked example returns `null` when the skin is
not being served or the value was not skin-local, so a page shows its own way out instead of
navigating somewhere wrong.

The section also says to read the origin on **every** use, because Decaid assigns the skin server
its port and a remembered origin can point at a port nothing is listening on.

The general guidance above it now separates the two cases: most parameters are REST lookup keys,
and `return` is named as the exception, with a cross-reference from the example.

Documentation only — no code and no behaviour change.

- `flutter analyze` — clean.
- `flutter test` — **full suite 3893 passed / 1 skipped**, run against current `main` on
  5 Sep 2026.
- `flutter test` — `plugins_handler_test` 13 passed, including the nine new query assertions.
- The aliasing claim was checked against `lib/src/plugins/plugin_manager.dart` on `main`, not
  assumed from the canary.

## Impact

- **Documentation**: `doc/Plugins.md` gains a section. No behaviour changes.
- **API**: none. The endpoint and the alias both already existed.
- **Compatibility**: none.
- **Security**: none.

## Contributor Responsibility


doc/Plugins.md
```diff
@@ -602,6 +602,71 @@ commands owned by the retiring generation, even if `onUnload()` fails. Late
...
+const shotId = new URLSearchParams(location.search).get("shotId");
+```
+
+Prefer this for a page that fetches its data over the REST API; it keeps the
+value out of the generated HTML.
+
+Skins are served from a different browser origin than plugin pages, so a skin
+cannot write a plugin page's `sessionStorage` or `localStorage`. A query
+parameter on a top-level navigation is how a skin hands a plugin page its
+subject; accept a `return` parameter for the way back.
```
tadelv, line 875: The preceding text correctly says plugin pages and skins are on different origins. That makes the example `return=/skin/history` unsafe/incomplete as a return contract: if a plugin page on the API origin assigns that value directly to `location`, it resolves on the API origin rather than the skin server, and accepting an arbitrary absolute `return` would create an open-redirect footgun. Please make this explicit and usable: constrain `return` to a skin-local path and reconstruct the known skin origin (or otherwise validate/allowlist the destination), and adjust the final “every parameter” guidance so `return` is handled as a navigation target rather than a REST lookup key.

tadelv (changes requested):
One required documentation fix: the proposed `return` navigation pattern needs to define a working, constrained cross-origin return target. The HTTP/query contract and tests otherwise match the current implementation and repository guidance.

doc/Plugins.md
ChampionDesigns: You were right — `return=/skin/history` was not a usable contract. A plugin page runs on the API
origin, so that path resolves there rather than on the skin server, and accepting an absolute URL
instead would have been an open redirect.

`return` now has its own section, and it is defined rather than implied:

1. Require a skin-local path — one leading slash, no `//`, no scheme. Nothing else is accepted.
2. Read the skin origin from `GET /api/v1/webui/server/status`, which already answers
   `{serving, path, port, ip}` and is already in `rest_v1.yml`. The page never learns the origin
   from the parameter.
3. Join them with `new URL(path, skinOrigin)`, which cannot escape that origin once the path is
   known to be skin-local.

The worked example returns `null` when the skin is not being served or the value was not
skin-local, so a page shows its own way out instead of navigating somewhere wrong. The section
also says to read the origin on every use, since a remembered one can point at a port nothing is
listening on.

The general guidance above it now separates the two cases: most parameters are REST lookup keys,
and `return` is named as the exception.

Documentation only. Pushed as 575bf7a4.

tadelv: The direction is right, but the worked example still does not produce a reliable/safe skin origin.

On current `main`, `WebUIService._serveFresh()` binds the backing skin server to `0.0.0.0`, and on Android `serverIP()` returns `_server.address.address`. So `/api/v1/webui/server/status` reports `ip: 0.0.0.0` there. That is the bind address, not the browser origin; the WebUI tests explicitly treat the live skin origin as e.g. `http://localhost:${service.port}` and the fixed entry point preserves the request host when redirecting to the rotated backing port.

Please reconstruct from the plugin page's already-known host plus the reported live `port`, rather than treating `status.ip` as an origin. For example, derive a URL from `location.origin`/`location.href`, replace only its port, then resolve the return path against that.

There is one related security detail worth fixing in the same example: `raw.startsWith('/') && !raw.startsWith('//')` does not by itself guarantee that `new URL(raw, skinOrigin)` stays on `skinOrigin` for HTTP URLs, because the WHATWG parser treats `\\` as a slash in special URLs. A value such as `/\\example.invalid` can therefore be interpreted as an authority change. The robust contract is to parse the target and require `target.origin === skinOrigin.origin` before returning it (in addition to requiring the intended path form).

With those two adjustments, this resolves my original concern cleanly.

tadelv (changes requested):
The original return-target ambiguity is mostly addressed, but the new worked example still needs one correction before merge: do not use `server/status.ip` as the browser origin. On Android that value is the `0.0.0.0` bind address, while the skin origin is based on the host the browser actually used plus the current backing port. Preserve the current page host, replace the port from `/api/v1/webui/server/status`, and validate the parsed target's `origin` after `new URL(...)` so alternate slash parsing cannot escape it. I left the concrete details on the existing thread. Everything else in this PR still looks consistent with current `main`, `AGENTS.md`, and the existing plugin HTTP implementation/tests.

tadelv (approved):
Previous review concern is resolved. The return-path guidance now preserves the browser host, substitutes only the live skin port, and verifies the parsed target stays on the reconstructed skin origin. Query pass-through tests and documentation remain consistent with the implementation and repository guidance. Current PR checks pass. GitHub still reports a merge conflict with current main, so the branch needs to be updated/rebased before merge, but I found no remaining blocker in the PR's changes themselves.

tadelv (approved):
Previous review concern is resolved. The return-path guidance now preserves the browser host, substitutes only the live skin port, and verifies the parsed target stays on the reconstructed skin origin. Query pass-through tests and documentation remain consistent with the implementation and repository guidance. Current PR checks pass. GitHub still reports a merge conflict with current main, so the branch needs to be updated/rebased before merge, but I found no remaining blocker in the PR's changes themselves.

PR #799, opened 2026-09-05
fix(webserver): refuse to boot when a fixed port is already bound

ChampionDesigns:
## Summary

The app could boot **with no API at all** and say nothing.

`startWebServer` binds two fixed ports: 8080 for the REST and WebSocket API, and 4001 for the API
docs. `main()` wrapped the call in a bare catch that logged "failed to start web server" and
carried on.

**That failure is close to invisible.** The WebUI binds an ephemeral port and is unaffected, so the
skin still loads and the app looks normal — only every REST call and every WebSocket behind it is
gone. `BootTiming.mark('webserver_up')` fires either way. The user sees an app that opened
correctly and then misbehaves, with the reason only in the log.

Two Decaid-family apps installed on one device share those fixed ports, so the common cause is
simply that the other one is already running.

Base: `main`. Independent.

### Design notes

- `serveOrReportPortInUse` raises a typed `WebServerPortInUse` for `EADDRINUSE` and **rethrows
  every other failure unchanged**, so a permission denial keeps its own error path and is never
  reported to the user as a port clash.
- `isAddressInUse` matches the OS code per platform — 98 on Linux and Android, 48 on macOS and
  iOS, 10048 on Windows — because Dart reports the raw code and does not normalise it.
- `main()` stops on `WebServerPortInUse` and shows a screen naming the port, offering "Check again"
  and "Close this app".

## Linked Issue

N/A
## Verification

- `flutter analyze` — clean.
- `flutter test` — **full suite 3904 passed** against current `main`, including
  `port_binding_test` and `webserver_port_conflict_app_test`.

## Impact

- **User-visible**: the app now refuses to boot and says which port is taken, instead of running
  without an API. That is a deliberate behaviour change.
- **Compatibility**: none for a normal single-app install; the path is only reached when a bind
  fails.
- **API**: none.
- **Security**: none. A permission failure is not swallowed as a port clash.

## Contributor Responsibility

### DIFF
--- a/doc/AI_API_NOTES.md
+++ b/doc/AI_API_NOTES.md
@@ -128,6 +128,61 @@ All four machine sockets: `/ws/v1/machine/snapshot`, `/ws/v1/machine/shotSetting
 
 **Full gateway mode is explicitly exempt**, checked via `settingsController.gatewayMode == GatewayMode.full` alongside the shot-state check — not left implicit. In `full` mode the skin owns the shot and `De1StateManager` normally never starts its own `ShotSequencer` for it, so `currentShotState` would usually stay idle anyway; but the launcher/home-screen path in `De1StateManager._handleEspressoState` starts an app-owned `ShotSequencer` "regardless of mode" whenever the app's own home screen is foregrounded, even under `full` gateway mode. Relying on ShotSequencer-absence alone would make the lockout accidentally engage in that edge case, which is out of scope for the setting's intent (it exists to protect app-tracked shots, not skin-owned ones) — hence the explicit gateway-mode check rather than an inferred one.
 
+## Fixed Port Already Bound
+
+### Problem
+
+`startWebServer` binds two fixed ports: 8080 for the REST and WebSocket API, and
+4001 for the API docs. `main()` wrapped the call in a bare `catch` that logged
+`failed to start web server` and carried on, so a bind failure left the app
+running with no API at all.
+
+That failure was close to invisible. The WebUI binds an ephemeral port and is
+unaffected, so the skin still loaded and the app looked normal; only every REST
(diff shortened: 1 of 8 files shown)


tadelv (changes requested):
There is one required fix before this should merge:

`startWebServer()` binds the real REST/WS server on 8080 first and only then calls `startApiDocsServer()` on 4001. This PR converts both binds to `WebServerPortInUse`, and `main()` treats either one as fatal. That means a conflict on **4001 only** now aborts normal app startup even though the actual API on 8080 has already started successfully. It also leaves that successfully-created 8080 `HttpServer` alive while the conflict UI is shown, because the server reference is local to `startWebServer()` and never gets closed on the later 4001 exception.

That broadens the fix beyond the stated failure mode and can create a confusing self-inflicted port conflict. Please make only failure to bind the required 8080 REST/WS API fatal, or otherwise make startup transactional and explicitly clean up 8080 if 4001 is intentionally required too. Given that 4001 serves API docs, I would strongly prefer keeping a docs-port failure non-fatal and logging it. Please add a regression test for the 4001-conflict case so the core API/app remains usable (or, if both ports are deliberately mandatory, that 8080 is rolled back cleanly).

Aside from this, the typed EADDRINUSE handling, error propagation for non-address-in-use failures, focused tests, and rationale placement are consistent with the repository guidance.

tadelv (approved):
Re-checked the update. The requested blocker is resolved: port 4001/API-docs conflicts are now handled locally as non-fatal, while only failure to bind the required 8080 REST/WS API reaches the fatal startup path. The added regression coverage exercises both the occupied and free docs-port cases. The implementation and rationale placement remain consistent with AGENTS.md, and current CI passes format, analyze, full tests, and Linux smoke build. No further blocking issues found.

PR #800, opened 2026-09-05
report pausedSteam and puffing as themselves, not as idle

ChampionDesigns:
## Summary

`mapDe1SubToMachineSubstate` collapsed five raw DE1 substates onto `MachineSubstate.idle`. Two of
them are steam phases a caller has a real reason to tell apart:

```
pausedSteam (0x12)  the valve is shut and the session is not over
puffing     (0x14)  the purge after steam stops on time or on milk temperature
```

Reported as `idle`, neither could be shown, and a client could not distinguish a pause from a purge
from a machine doing nothing at all. **The purge matters most:** it can hold for a while, and a
screen that cannot name it leaves someone watching a machine that looks stopped and is not.

`noState`, `userNotPresent` and `refill` still map to `idle` — they carry no phase a caller can act
on.

Base: `main`. Independent. Two files, 15 lines.

### What this could have broken, and does not

`shot_sequencer` ends a shot when the substate reads `pouringDone` or `idle`, so a steam session
arriving as `idle` used to satisfy it. **It never reaches that branch**: the sequencer leaves `idle`
only on `espresso` + `preparingForShot`, and aborts in `preheating` if the state is not espresso,
so steam never enters `ShotState.pouring`. Every other reader compares against a specific substate
rather than switching exhaustively.

## Linked Issue

N/A
## Verification

- `flutter analyze` — clean.
- `flutter test` — **full suite 3891 passed** against current `main`.
- The sequencer reasoning above was checked against the code, not assumed.

## Impact

- **API**: `MachineSubstate` gains two values that were previously reported as `idle`. A client
  switching exhaustively on the enum must handle them.
- **Compatibility**: a client comparing against specific substates is unaffected. One that treats
  "not a substate I know" as idle keeps today's behaviour.
- **User-visible**: a skin can now name the steam pause and the purge.
- **Security**: none.

## Contributor Responsibility

### DIFF
--- a/lib/src/models/device/impl/de1/de1.utils.dart
+++ b/lib/src/models/device/impl/de1/de1.utils.dart
@@ -70,9 +70,15 @@ MachineSubstate mapDe1SubToMachineSubstate(De1SubState de1SubState) {
     case De1SubState.noState:
     case De1SubState.userNotPresent:
     case De1SubState.refill:
-    case De1SubState.pausedSteam:
-    case De1SubState.puffing:
       return MachineSubstate.idle;
+
+    // Both are steam phases with nothing pouring, and both were reported as `idle`.
+    // A caller could not tell a pause from a purge from a machine doing nothing, so
+    // neither could be shown, and the purge in particular can last a while.
+    case De1SubState.pausedSteam:
+      return MachineSubstate.pausedSteam;
+    case De1SubState.puffing:
+      return MachineSubstate.puffing;
 
     case De1SubState.heatWaterTank:
     case De1SubState.heatWaterHeater:
--- a/lib/src/models/device/impl/de1/de1.utils.dart
+++ b/lib/src/models/device/impl/de1/de1.utils.dart
@@ -72,9 +72,6 @@ MachineSubstate mapDe1SubToMachineSubstate(De1SubState de1SubState) {
     case De1SubState.refill:
       return MachineSubstate.idle;
 
-    // Both are steam phases with nothing pouring, and both were reported as `idle`.
-    // A caller could not tell a pause from a purge from a machine doing nothing, so
-    // neither could be shown, and the purge in particular can last a while.
     case De1SubState.pausedSteam:
       return MachineSubstate.pausedSteam;
     case De1SubState.puffing:
--- a/lib/src/models/device/machine.dart
+++ b/lib/src/models/device/machine.dart
@@ -171,6 +171,14 @@ enum MachineSubstate {
   cleanSoaking,
   cleaningSteam,
 
+  /// The steam valve is shut but the session is not over: the machine is between
+  /// pours and will resume where it left off.
+  pausedSteam,
+
+  /// The short purge the machine makes after steam stops on time or on milk
+  /// temperature. It can hold this state for a while, so a screen should say so.
+  puffing,
+
   errorNaN,
   errorInf,
   errorGeneric,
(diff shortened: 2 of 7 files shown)


lib/src/models/device/machine.dart
```diff
@@ -171,6 +171,14 @@ enum MachineSubstate {
   cleanSoaking,
   cleaningSteam,
 
+  /// The steam valve is shut but the session is not over: the machine is between
+  /// pours and will resume where it left off.
+  pausedSteam,
```
tadelv, line 175: `MachineSnapshot.toJson()` serializes `state.substate.name`, so adding these enum members creates new values on both `GET /api/v1/machine/state` and `ws/v1/machine/snapshot`. `AGENTS.md`/`CONTRIBUTING.md` require API changes to update the authoritative specs in the same PR. Please add `pausedSteam` and `puffing` to the `MachineSubstate` schemas in both `assets/api/rest_v1.yml` and `assets/api/websocket_v1.yml`, and update `doc/Api.md` as required for the changed API contract.

lib/src/models/device/impl/de1/de1.utils.dart
```diff
@@ -70,9 +70,15 @@ MachineSubstate mapDe1SubToMachineSubstate(De1SubState de1SubState) {
     case De1SubState.noState:
     case De1SubState.userNotPresent:
     case De1SubState.refill:
+      return MachineSubstate.idle;
+
+    // Both are steam phases with nothing pouring, and both were reported as `idle`.
+    // A caller could not tell a pause from a purge from a machine doing nothing, so
+    // neither could be shown, and the purge in particular can last a while.
     case De1SubState.pausedSteam:
+      return MachineSubstate.pausedSteam;
```
tadelv, line 76: This is new externally observable behavior, but the PR adds no regression test. `CONTRIBUTING.md` makes this a hard gate: new behavior needs a test and bug fixes need regression coverage. Please add focused tests proving `pausedSteam -> MachineSubstate.pausedSteam` and `puffing -> MachineSubstate.puffing`; because these values are exposed by the machine snapshot API, cover the serialized contract as appropriate as well.

lib/src/models/device/impl/de1/de1.utils.dart
```diff
@@ -70,9 +70,15 @@ MachineSubstate mapDe1SubToMachineSubstate(De1SubState de1SubState) {
     case De1SubState.noState:
     case De1SubState.userNotPresent:
     case De1SubState.refill:
+      return MachineSubstate.idle;
+
+    // Both are steam phases with nothing pouring, and both were reported as `idle`.
```
tadelv: `AGENTS.md` explicitly says not to add explanatory comments to new/substantially rewritten code and to keep rationale out of production comments. This block restates the motivation rather than explaining a non-obvious invariant. Please remove it (the mapping is clear from the case names), keeping any longer rationale in the PR/design documentation instead.

tadelv (changes requested):
Requesting changes for repository-guidance hard gates. The core mapping looks reasonable, but this PR changes the public MachineSubstate contract without updating the authoritative API specs/docs or adding the required regression coverage. There is also one AGENTS.md code-style violation in the new explanatory comment block.

tadelv (approved):
Re-reviewed current head. The prior blockers are addressed: the REST and WebSocket MachineSubstate schemas and API/skin docs now include pausedSteam/puffing; focused regression/serialization coverage was added; and the explanatory rationale block in de1.utils.dart was removed. CI is green on this head. I don't see any remaining merge-blocking issues.

PR #806, opened 2026-09-05
fix(firmware): poll for verify completion when the notification never arrives

ChampionDesigns:
## Summary

Some firmwares never emit the post-erase / post-verify notification, leaving the firmware-update flow
waiting forever on a push that will not arrive.

Base: `main`. Independent.

The verify flow now actively re-reads the firmware-map register rather than waiting to be notified.
Adds `readFwMapRequestFresh` — a read that bypasses the cached value, since the whole point is to
observe a change the cache cannot see — plus the poll itself, with a bounded timeout so a genuinely
stuck erase still fails rather than hanging.

### Design notes worth reviewing

**Serial and BLE take different paths on purpose.** On serial, `_serialRead` would hand back the last
*pushed* `[I]` frame, which never updates once the firmware stops emitting the notify. Re-sending
`<+I>` reprovokes a fresh frame. No matching `<-I>` is sent: dropping the continuous subscription
mid-update would blind the notify path a stock DE1 still relies on.

**The BLE poll deliberately avoids `read()`.** It goes through `_bleRead`, so it does not inherit
`read()`'s timeout disconnect-and-reconnect recovery. Tearing the link down and re-establishing it
mid-update would corrupt the in-flight firmware write. A failed poll read throws to the caller, which
logs and retries; the outer stage timeout still bounds the overall wait.

**Both the single read and the loop are bounded.** One stalled read cannot hang the poll, and a
genuinely stuck erase still fails.

### Note for reviewers

This combines what were two commits in the port's own history (`87819b21` + `3435af28`). The first
references a timeout constant the second defines, so neither ever compiled on its own in either line
and the split was not meaningful. They are presented here as the single working change.

### Note on `test/helpers/fake_serial_transport.dart`

This PR and the Bengle puck-estimator PR both add this helper, byte-identical. Each PR is independent
and each needs a serial fake, so neither can rely on the other landing first. Git merges the two
identical additions without a conflict. If you would rather it lived in one place, say which PR should
own it and I will rebase the other onto it.

### Brought current with upstream, 5 Sep 2026

`upstream/main` was **merged into** this lane rather than rebased onto it, so the three-dot diff
GitHub shows is still exactly the lane's own work while the merge base moves forward. The merge
resolutions were checked individually — `git rerere` is enabled in the working repo and replayed
two of them wrongly elsewhere in this round, so none was taken on trust.

## Linked Issue

N/A
## Verification

- `flutter analyze` — clean.
- `flutter test` — **full suite 3897 passed / 1 skipped**, run against current `main` on 5 Sep 2026.
- `firmware_verify_poll_test.dart` covers the BLE poll path, the fresh-read path and the timeout.
- `firmware_verify_poll_serial_test.dart` covers the serial reprovoke: `<+I>` goes out, no `<-I>` follows,
  and the next fresh `[I]` frame resolves the read.
- **Verified against real firmware.** The poll path runs in the Decaid-Canary build Ben uses for
  firmware updates, on the machine whose firmware prompted this fix.

## Impact

- **API**: no wire or schema change. `doc/Api.md` now says the verify step polls rather than waits for
  a notification, because the old wording implied a push.
- **Compatibility**: none. The NDJSON event order and the `done` condition are unchanged.
- **User-visible**: a firmware update that used to hang at the verify step now completes, or fails with
  a timeout instead of hanging.
- **Security**: none.

## Update — rebased onto `main` 28 Aug 2026

The lane was rebased from `94873593` onto `82f941d7`. One file conflicted,
`unified_de1_transport.dart`, and it was an add/add at the same point: upstream added
`_defaultShotSettingsFrame` (the serial shot-settings work, `8995a314` / `bb392d9c` / `ae1f7c1b`)
and this lane adds `_firmwareMapPollReadTimeout`. Both are kept. Nothing was chosen over anything.

`dart format` and `flutter analyze` are clean on the rebased lane.

## Contributor Responsibility


tadelv (changes requested):
Two required cleanup items before merge:

1. **AGENTS.md code-style violation:** this PR adds extensive explanatory/rationale comments throughout the new production and test code (notably the firmware poll loop, `readFwMapRequestFresh`, timeout changes, fake transports, and regression tests). `AGENTS.md` explicitly says not to add explanatory comments to new/substantially rewritten code and to put rationale, hardware constraints, and debugging history in the matching `doc/AI_*_NOTES.md`. Please reduce the inline comments to only what is needed to understand non-obvious mechanics from the code itself, and move the BLE/serial firmware-update rationale into `doc/AI_BLE_NOTES.md` (or the most appropriate existing AI notes file).

2. **`doc/Api.md` describes a different mechanism than the implementation:** it currently says the app polls the firmware-map register *rather than waiting for a notification*, but the implementation intentionally races the existing notify path against polling. Please describe that as a bounded polling fallback/race, or keep the API doc focused on externally observable behavior and move the transport rationale to the AI notes.

I did not find a functional merge blocker in the actual poll/notify design. The regression coverage is appropriate and the PR checks are green.

tadelv (changes requested):
The two previous review items are resolved. The rationale is now in `doc/AI_BLE_NOTES.md`, the inline comments are reduced to local non-obvious mechanics, and `doc/Api.md` now accurately describes the notify/poll race. I do not see a new functional blocker in the firmware-update change itself, and the current head's PR checks are green.

One remaining merge blocker: GitHub currently reports this head as `dirty` / non-mergeable against current `main`. `main` has advanced since this branch's `2b60dd7e` base through overlapping BLE-notes/discovery-test work, so please bring the branch current again, resolve the conflict while preserving both sets of changes, and rerun the normal verification. No other changes requested from this review.

tadelv (changes requested):
The previous merge/conflict blocker is resolved, and the follow-up that removes the shadowing `PluginBleFixtureTransport` counters is correct. I do not see a functional blocker in the firmware poll implementation.

One remaining required documentation fix in the new `doc/AI_BLE_NOTES.md` section:

- It says "neither transport's normal read path returns a current register value here", but on BLE the normal public `read()` path does perform a fresh GATT read; the reason it is unsuitable during firmware update is its timeout recovery can disconnect/reconnect. The paragraph immediately below already describes this correctly.
- More importantly, the terminal-frame sentence says "Erase and verify reach terminal on error `ff ff ff`". That is not what the implementation does: erase completion requires `ff ff ff`, while during verification `ff ff ff` is the non-terminal/pending value; verification becomes terminal when the error tuple is anything else, with `ff ff fd` being success. The regression test with `ff ff 01` also relies on that behavior.

Please correct those two statements so the AI notes match `_isEraseComplete` / `_isTerminalVerificationResponse`. No code changes requested.

tadelv (approved):

PR #812, opened 2026-09-07
fix(ble): prevent stale discovery cache from disconnecting active links

### LINKED ISSUE #810: BLE discovery stale-cache teardown can disconnect active/reconnecting devices across platforms
## Summary

`0.8.6-beta.1` (`bc717af2`) has a reproducible BLE connection/reconnect regression that is not present in `0.8.5` (`a08bc41e`).

The failure has been reproduced on iOS, but investigation shows that the underlying regression is in shared BLE discovery/lifecycle code and is **not iOS-specific**.

The strongest regression point remains #773. Its stale discovery-cache recovery added a destructive `disconnect()` to `UniversalBleDiscoveryService._deviceScanned()` when a fresh advertisement disagrees with the cached `Device` state.

That fixes an important #451/#598/#638-class problem — stale cached devices must not suppress fresh advertisements indefinitely — but the implementation does not establish that the cached instance still owns the native BLE connection before tearing it down.

The result is a race where discovery can disconnect a connection that `ConnectionManager` / `EarlyConnectWatcher` is currently establishing or has just established.

### Bad: `0.8.6-beta.1`

Observed on iOS:

1. DE1 connects while sleeping.
2. Solicited MMR traffic works.
3. The machine snapshot stream is silent for 10 seconds while the native BLE link still reports connected.
4. The existing snapshot-staleness watchdog begins a clean reconnect.
5. During recovery, discovery sees the preferred DE1.
6. `EarlyConnectWatcher` begins connecting it.
7. A concurrent `_deviceScanned()` stale-cache decision calls `disconnect()` for the same BLE device ID.
8. BLE operations subsequently hit 10-second timeouts and connection/reconnect churn repeats.

### Good: `0.8.5`

With the same machine/iPhone:

- the machine connects while sleeping;
- remains connected;
- later sleep → idle transitions continue normally.

The snapshot-staleness watchdog predates the good build, so it is a detector/recovery trigger rather than the regression itself.

## Regression range

Known good:

```text
0.8.5
a08bc41e
```

Known bad:

```text
0.8.6-beta.1
bc717af2
```

#773 is the strongest suspect in this range and introduced the relevant stale-cache replacement behavior in `UniversalBleDiscoveryService`.

Do not assume that another change in the range is impossible, but the fix should first prove or disprove this path with regression tests.

---

## Root problem

`UniversalBleDiscoveryService._deviceScanned()` currently mixes two different responsibilities:

1. deciding whether a cached Dart `Device` should continue suppressing a fresh advertisement;
2. deciding whether the native BLE link for that device ID should be destroyed.

Those decisions are not equivalent.

A stale cache entry may safely need replacement while another part of Decaid is already establishing a newer connection to the same native BLE device.

### Current unsafe sequence

Conceptually the code does:

```text
fresh advertisement
    ↓
read cached Device.connectionState
    ↓
possibly query UniversalBle.getConnectionState(deviceId)
    ↓
decide cached Device is stale
    ↓
remove cache entry
    ↓
cancel discovery listener
    ↓
await existing.disconnect()
```

There is no connection-owner/generation validation around the destructive final step.

Several awaits separate the initial observation from `disconnect()`, so the ownership situation can change before teardown occurs.

---

### 1. `connecting + native disconnected` is incorrectly considered stale

A Decaid `Device` enters `ConnectionState.connecting` before its native BLE connection has necessarily started or reached the OS `connecting` state.

Therefore this is a valid transitional state:

```text
Decaid Device: connecting
native BLE:    disconnected
```

It must **not** by itself mean that the cached device is stale.

This is particularly important because platform implementations expose connection progress differently.

### 2. The native probe is asymmetric

The native-state safety check currently only runs when the cached Dart state is:

```text
connected
connecting
```

For:

```text
discovered
disconnected
```

the code proceeds toward cache replacement and `existing.disconnect()` without proving that the native link is also inactive.

That means stale Dart state can destroy a healthy or newly-starting native connection.

### 3. Timeout is converted into permission to destroy state

This pattern is unsafe for destructive lifecycle decisions:

```dart
connectionState.first.timeout(
  const Duration(seconds: 2),
  onTimeout: () => ConnectionState.disconnected,
)
```

A timeout means **unknown**, not disconnected.

Failure or uncertainty while determining ownership must fail safe: preserve the link / defer cleanup rather than issue a native disconnect.

### 4. The stale decision has a TOCTOU race

Even if both Dart and native state were correct when sampled:

```text
check ownership
    ↓
await
    ↓
await
    ↓
connection starts / ownership changes
    ↓
disconnect
```

The original observations no longer prove that teardown is safe.

Rechecking `_devices[deviceId]` alone is also insufficient: `EarlyConnectWatcher` can start connecting the same cached `Device` instance between the stale-state sample and the disconnect.

Any destructive teardown needs an ownership/generation fence that remains valid when the native disconnect actually executes.

### 5. `BleLifecycleGate` serializes operations but does not establish ownership

`BleLifecycleGate` is currently a per-device FIFO.

That prevents native lifecycle operations from literally executing simultaneously, but it can produce this legal sequence:

```text
new connect()
    ↓
connect succeeds
    ↓
stale old disconnect() was queued behind it
    ↓
disconnect executes
    ↓
new connection dies
```

Serialization therefore does not fix this race.

Any solution using the lifecycle gate must distinguish **which connection generation/owner** is allowed to perform teardown.

### 6. Transport connection generations are instance-local

`UniversalBleTransport._connectionGeneration` protects asynchronous work inside one transport instance.

It does not establish ownership between:

```text
old cached transport
new/reconnecting transport
```

for the same native device ID.

Do not assume the existing transport generation is sufficient to fence stale discovery cleanup.

### 7. Cache disconnect listeners are not instance-fenced

Both normal discovery and quick-connect install listeners that effectively do:

```dart
if (state == ConnectionState.disconnected) {
  _devices.remove(deviceId);
}
```

They do not verify:

```dart
identical(_devices[deviceId], deviceThatEmittedTheEvent)
```

A disconnect from an older device generation must never remove a newer cache entry.

This is especially relevant in quick-connect, where the new `_devices[key]` entry is assigned before the previous `_connections[key]` subscription has finished being cancelled.

Fence cache mutation by device-instance identity or an equivalent cache generation.

---

## Why this is cross-platform

The implementation defect is shared. iOS is currently the confirmed hardware reproducer.

### iOS / macOS

Both use the same Darwin/CoreBluetooth `universal_ble` backend.

`disconnect(deviceId)` resolves the `CBPeripheral` for that ID and calls `cancelPeripheralConnection()`.

A stale Dart object therefore does not have an isolated native connection that can safely be disconnected independently of a newer Dart object using the same peripheral.

macOS must be considered affected until proven otherwise.

### Android

Android does expose `STATE_CONNECTING`, but Decaid/`universal_ble` deliberately has asynchronous lifecycle windows, including pre-connect settling and connect/disconnect spacing.

A valid Decaid `connecting` state can therefore precede a native `connecting` state.

A stale disconnect can also be delayed and execute after connection progress has changed.

The fix must not regress the Android background scale-watch and hard-power-cycle recovery that motivated #451/#773.

### Linux / BlueZ

This is an especially clear unsafe case.

The Linux backend:

- tracks an in-flight connection separately in `_connectingDevices`;
- `getConnectionState()` only returns `connected` or `disconnected`;
- while a connection is in progress, `device.connected` may still be false;
- `disconnect()` explicitly looks in `_connectingDevices` and disconnects that device.

Therefore:

```text
valid BlueZ connect in progress
→ getConnectionState() == disconnected
→ stale-cache cleanup
→ disconnect()
→ in-flight connect cancelled
```

must be covered by regression tests even if Linux hardware reproduction is not immediately available.

### Windows

The Windows backend also does not expose a useful `connecting` state through `getConnectionState()`.

`GetConnectionState()` returns disconnected until the device has been inserted into `connected_devices_`, while `ConnectAsync()` performs asynchronous work before that insertion.

Therefore `native == disconnected` is not sufficient evidence that no Windows connection attempt owns the device.

The fix must model this state in tests.

---

## Required design invariant

For a given normalized BLE device ID:

> Discovery/cache maintenance must never issue a native disconnect that can affect a connection generation it does not own.

A stale cached `Device` must not be able to tear down:

- the same instance after another component has started connecting it;
- a replacement instance;
- a newer reconnect generation;
- an OS/native connection attempt whose platform state still reports disconnected.

Treat **cache ownership** and **native connection ownership** as separate concepts.

### Preferred architectural direction

`UniversalBleDiscoveryService` should primarily own discovery/cache state.

Native connection teardown should preferably remain with the component that owns connection lifecycle (`ConnectionManager` / transport lifecycle), rather than being triggered opportunistically by an advertisement callback.

If native teardown remains in stale-cache replacement, it must be protected by a per-device ownership/generation mechanism whose validity is checked at the point the destructive operation executes — not only before an await.

Do not fix this with:

- an iOS-only conditional;
- extra arbitrary delays;
- extra connect retries;
- longer BLE timeouts;
- suppressing the snapshot watchdog;
- simply reverting stale-cache replacement from #773.

Those approaches either hide the race or reintroduce #451/#598/#638.

---

## Preserve the original stale-cache recovery

#773 addressed a real problem.

After hard power loss, a cached `Device` can remain logically connected even though the native link is gone. A fresh advertisement must eventually allow Decaid to create/use a fresh device instance.

The fix therefore needs both properties:

```text
genuinely stale cache
→ fresh advertisement is not suppressed indefinitely
→ recovery can continue
```

and:

```text
connection/reconnection in progress
→ duplicate/fresh advertisement
→ no destructive disconnect of that connection
```

Cache eviction/replacement does not automatically imply permission to issue a native disconnect.

---

### A. Connecting state is protected

Set up:

```text
cached Device state = connecting
native state        = disconnected
fresh advertisement arrives
```

Assert:

- cached/current connection attempt is not disconnected;
- no destructive stale cleanup runs;
- connection may continue normally.

This represents a valid pre-native-connect window on every platform and is especially important for Linux/Windows semantics.

### B. Stale decision racing a connect

Deterministically pause stale-cache processing after it observes an apparently stale state.

Then:

1. begin connecting that device;
2. allow stale cleanup to continue.

Assert that stale cleanup cannot disconnect the in-flight/new connection.

The test must model the shared device-ID lifecycle rather than using two completely independent fake transports where `disconnect()` cannot affect `connect()`.

### C. Disconnect queued behind connect

Model:

```text
connect(deviceId) in lifecycle gate
stale cleanup requests disconnect(deviceId)
connect completes
```

Assert that the stale cleanup does **not** turn this into:

```text
connect succeeds
disconnect executes
```

against the newly owned native link.

### D. Genuine stale cache still recovers

Preserve the #773/#451 behavior:

```text
cached Device says connected
native link is genuinely gone
no connection/reconnection owns device
fresh advertisement arrives
```

Assert:

- the stale cache does not suppress the advertisement indefinitely;
- a fresh usable device can be produced;
- no unrelated/new connection is disconnected.

Update the existing test that currently requires:

```text
staleTransport.disconnectCalls >= 1
```

That assertion encodes the implementation responsible for this regression. Test the required recovery outcome instead of requiring destructive teardown.

### E. Unknown/timeout is non-destructive

Simulate:

- cached state lookup timing out;
- native-state lookup failing/timing out where applicable.

Assert that uncertainty does not authorize a native disconnect.

### F. Old cache listener cannot evict new generation

Create device generation A, then replace it with generation B.

Deliver a delayed `disconnected` event from A.

Assert:

```text
_devices[id] == generation B
```

Do the equivalent around the quick-connect replacement path.

### G. Linux-style connection semantics

Model a backend where:

```text
connect has started
getConnectionState() == disconnected
```

until connection completion.

Assert that advertisements/cache maintenance cannot cancel that connect.

### H. Windows-style connection semantics

Model a connection attempt that is not represented in the backend's connected-device map until late in asynchronous setup.

Assert that a transient native `disconnected` result cannot cause discovery to invalidate or tear down the attempt.

---

## Existing test harness concern

The current stale-cache regression test uses `_TrackingFakeBleTransport` and checks that the stale transport's `disconnect()` was called.

That fake does not reproduce the important native property:

> two Dart device/transport instances with the same BLE device ID ultimately act on the same OS connection.

At least one regression test for this issue must therefore use a fake lifecycle/native layer where stale `disconnect(deviceId)` can actually interfere with the concurrent `connect(deviceId)`.

Without that coupling, the original regression can pass unnoticed.

---

### Required: iOS

Re-run the original A/B scenario on the affected iPhone/DE1 combination.

Verify:

- DE1 can connect while sleeping;
- solicited traffic works;
- machine snapshot subscription becomes live;
- sleep → idle remains functional;
- forced snapshot-staleness recovery reconnects cleanly;
- no repeating 10-second BLE-operation timeout loop;
- duplicate advertisements during connect/reconnect do not trigger destructive cache cleanup.

Reliability should be at least equivalent to `0.8.5`.

### Required: Android regression check

Because #773 primarily hardened Android background scale-watch behavior, verify that the fix does not undo it.

At minimum retest the #598/#638-style sequence:

```text
DE1 + preferred BLE scale connected
→ machine sleep
→ scale hard power off
→ scale power on
→ machine wake
```

The fresh scale advertisement must still recover from a genuinely stale cache without restarting Decaid.

### Other platforms

macOS, Windows, and Linux hardware reproduction is useful but should not block fixing a race that can be demonstrated from their current backend semantics.

Their lifecycle behavior must nevertheless be represented by deterministic tests.

---

## Acceptance criteria

- [ ] iOS connection/reconnect reliability is restored to at least `0.8.5`.
- [ ] The fix is platform-neutral; there is no iOS-only workaround for shared lifecycle code.
- [ ] A cached `connecting` device is never declared stale solely because native state currently reports disconnected/disconnecting.
- [ ] Timeout/unknown state does not authorize destructive teardown.
- [ ] A stale-cache decision cannot disconnect a connection that starts or changes ownership while that decision is in flight.
- [ ] Serialization alone is not treated as ownership; stale cleanup queued behind a newer connect cannot disconnect that connection afterward.
- [ ] Older device/cache generations cannot remove newer `_devices` entries when delayed disconnect events arrive.
- [ ] Genuine stale-cache recovery from #451/#598/#638 remains functional.
- [ ] The current test requiring stale-cache `disconnectCalls >= 1` is replaced with an outcome/ownership-based assertion.
- [ ] Tests cover Darwin-style, Android-style, Linux-style, and Windows-style connection-state timing where relevant without platform conditionals in production logic.
- [ ] The snapshot-staleness watchdog remains enabled and can still trigger a clean recovery.
- [ ] The repeating 10-second BLE timeout/reconnect loop is no longer reproducible.
- [ ] `dart format lib test`, relevant targeted tests, `flutter analyze`, and the full `flutter test` suite pass.

## Agent workflow

Follow `AGENTS.md`.

This is a non-trivial BLE lifecycle fix:

1. inspect the relevant BLE/discovery/ConnectionManager guidance;
2. create the required implementation plan under `doc/plans/`;
3. add failing regression tests before changing production behavior;
4. preserve the stale-cache recovery invariant from #451/#773;
5. implement the smallest ownership-safe fix;
6. run targeted tests, `flutter analyze`, then the full test suite;
7. include verification evidence in the PR.

The PR should close #810 and explain explicitly how the implementation prevents a stale discovery/cache generation from tearing down the current native connection generation.

tadelv:
## Summary

What changed, and why?

- Separate BLE discovery-cache ownership from native connection ownership so a stale cache generation cannot disconnect an active or reconnecting device with the same native ID.
- Preserve unknown and transitional cache states, recheck Dart/native ownership after awaits, and identity-fence delayed disconnect listeners.
- Preserve genuine stale-cache recovery by replacing confirmed stale cache entries without issuing native teardown from discovery.
- Close the final-native-probe race: revalidate Dart state and cache identity after the last native probe, with no subsequent await before cache removal; preserve discovered state at the intermediate recheck too.
- Add deterministic regressions for shared-native connect/disconnect behavior, Linux/Windows-style pre-native-connect timing, timeout/unknown handling, and normal/quick-connect listener generations.

## Linked Issue

Fixes #810

## Verification

- `dart format lib test`
- `flutter test test/services/universal_ble_discovery_service_test.dart` — 43 passed
- Focused lifecycle/transport tests — 33 passed
- `flutter test` — 3,912 passed, 1 skipped
- iPhone hardware spot check: Bengle connected successfully with this fix
- New final-probe and intermediate-discovered regressions failed before the fix (an unwanted second transport was created), then passed. Final-probe coverage includes connecting, discovered, and disconnecting transitions.
- Follow-up coding delegated to a Pi subagent using `deepseek/deepseek-v4-flash` with medium thinking; parent reviewed correctness/scope with Karpathy guidelines and complexity with Ponytail review, and independently reran all 43 discovery tests.
- Full iOS sleep/reconnect and Android scale hard-power-cycle matrices remain outstanding and are explicitly deferred to a separate hardware session. This update does not establish hardware acceptance or merge readiness.

## Impact

- Restores platform-neutral BLE connection/reconnect safety without disabling stale-cache recovery or the snapshot watchdog.
- No API/spec, database, migration, or compatibility change.
- Updated `doc/DeviceManagement.md` and `doc/AI_BLE_NOTES.md` with the ownership invariant.

## Contributor Responsibility


ODevStudio (changes requested):
The overall direction is correct: discovery no longer performs native teardown, timeout/unknown states fail safe, and disconnect listeners are instance-fenced. CI is green, and the added regression coverage is substantial.

Blocking — the final native probe reopens the TOCTOU window

In _deviceScanned(), the stale-connected path now does roughly:

Read cached Dart state.
Probe native state.
Verify cache identity.
Re-read Dart state.
Await a second native-state probe.
Evict the cached instance.

The problem is step 5. If the same cached instance begins connecting while that second _nativeLinkState() is pending, and the backend still reports disconnected, execution proceeds directly to _evictCachedDevice(). There is no final Dart-state check after that await.

That leaves essentially the same class of TOCTOU hole the PR is intended to close. It does not issue a native disconnect anymore, which is an important improvement, but it can still remove an actively connecting device from discovery's cache and replace it with a second generation. That violates the issue's stated invariant that a cached device must not be discarded when its connection state changes while stale-state evidence is being gathered.

There is a related omission in the second Dart-state check:

if (latestState == null ||
    latestState == ConnectionState.connecting ||
    latestState == ConnectionState.disconnecting) {
  return;
}

ConnectionState.discovered is protected by the initial check but not by this recheck, even though the PR's own design says transitional/discovered state should be preserved.

Suggested fix: after the second native probe reports inactive, perform one final cache-identity + Dart-state check. There must be no further await between that final state validation and removing _devices[deviceId]. Preserve null, discovered, connecting, and disconnecting; permit eviction only if the same cache instance remains genuinely stale.

I would also add a deterministic test specifically for the missing window:

cached Dart state starts connected;
first native probe returns disconnected;
second Dart read returns connected;
block the second native probe;
emit Dart connecting;
release the second probe with disconnected;
assert the original cached instance remains and no replacement transport is created.

The existing “connection becoming active during stale observation” regression changes state while the first native probe is blocked, so it does not cover this final-await race.

Merge readiness

Separately, I would not consider the issue acceptance criteria complete yet. The PR reports only an iPhone/Bengle spot check and explicitly says the full iOS sleep/reconnect and Android scale hard-power-cycle matrices remain outstanding. Those field checks are part of #810's stated verification requirements.

Bottom line: good architectural correction, but I would request changes for the remaining final-probe race, then require the stated hardware verification before merge.

ODevStudio (changes requested):
Verdict: code looks ready; hardware acceptance is still incomplete. I would not approve/merge #810 as fully satisfied yet.

The earlier blocking code-review finding has been fixed on the current head 33fbfff8. The commit adds the missing post-second-native-probe Dart-state/cache-identity validation, protects discovered at the intermediate check, and adds deterministic tests for transitions to connecting, discovered, and disconnecting in precisely that final-await window. The production implementation also removes discovery-owned native disconnect() and identity-fences cache listeners, which is consistent with #810’s ownership invariant.

I did not find another blocking code defect in the changed implementation. Current CI is also green: formatting, flutter analyze, flutter test, and the Linux build smoke job all succeeded on this head.

The linked hardware comment is useful evidence, but it accurately describes itself as partial. It establishes successful iOS DE1 sleep/wake without the repeating 10-second loop, plus successful Android HDS reconnection after both graceful shutdown and a separate abrupt power loss.

The remaining blocking acceptance gaps are:

iOS: forced snapshot-staleness/watchdog recovery was not exercised. The controlled comparison with 0.8.5 was also not done, so “reliability at least equivalent to 0.8.5” has not really been established.
Android: #810 specifically requires DE1 sleep → scale hard power off → scale power on → machine wake. The session split this into two tests: graceful scale shutdown during DE1 sleep, and abrupt power loss while the DE1 was otherwise connected. That does not exercise the required combined sequence.
The comment itself explicitly states that it does not establish completion of the full hardware acceptance matrix.

The formal GitHub state also still contains the earlier CHANGES_REQUESTED review; its code objection is now obsolete, but its hardware-readiness objection remains valid.

Issue #809, opened 2026-09-07
Plugin drivers: host-owned BLE binding + plugin-backed Scale support


tadelv:
# Plugin drivers: host-owned BLE binding and plugin-backed Scale support

## Context

Discussion #749 sets the driver boundary: Decaid owns native capabilities and device lifecycle; JavaScript plugins own device protocols and device-specific behavior.

PR #759 added permission-gated outbound WebSocket/TCP/TLS transports. PR #775 added manifest-declared plugin sensors with generation-owned registrations in the existing device inventory and Sensor APIs.

Implement BLE-backed plugin drivers and Scale registration without redesigning the driver system.

## Outcomes

Deliver both pieces under this issue and validate them together, while allowing the work to land in focused implementation slices:

1. Add a device-type-independent, host-owned BLE binding for manifest-declared plugin drivers. Reuse Decaid's existing discovery, scan arbitration, connection lifecycle, queueing, recovery, and `UniversalBleTransport` infrastructure.
2. Adapt plugin drivers to Decaid's existing `Scale` abstraction, controllers, and APIs. The Scale adapter must remain independent of transport choice.

BLE drivers share Decaid's native BLE machinery for discovery, physical connection lifecycle, GATT access, cleanup, and reconnect policy. Each plugin driver owns its device protocol: service and characteristic choices, framing, parsing, checksums, handshakes, protocol state, commands, and readiness.

A BLE sensor enters the existing Sensor surface. A BLE scale enters the existing Scale surface. Keep Bookoo-specific and other device-specific protocol logic out of the native BLE bridge.

Use Bookoo Mini Scale as the reference BLE Scale unless implementation discovers a concrete blocker. Keep the native Bookoo implementation active throughout the proof so plugin-first arbitration and native fallback are exercised against the same hardware.

---

### 1. Driver declaration and BLE permission

BLE access is a separate permission from declaring a domain driver:

```json
{
  "permissions": ["transport.ble"],
  "drivers": [
    {
      "id": "bookoo",
      "type": "scale",
      "capabilities": [
        "battery",
        "tare",
        "timerControl",
        "disconnectToSleep"
      ],
      "ble": {
        "match": {
          "name": { "contains": "bookoo" },
          "serviceUuids": ["0ffe"]
        }
      }
    }
  ]
}
```

Keep existing plugin-created driver registrations compatible. The following limits apply specifically to the new host-bound BLE-driver feature:

- at most **1 BLE-backed driver declaration per plugin** in v1;
[...]

PR #813, opened 2026-09-07
WIP: host-owned BLE drivers and plugin Scale support (#809)

ODevStudio:
## Summary

Draft implementation checkpoint, not ready to merge and not completion of #809.

- Add BLE permission/declaration validation, tri-state matching, generation-owned registry and physical claims. These primitives are not yet integrated with discovery.
- Add session retirement/revocation primitives, bounded cleanup authority, GATT resource limits, subscription identity, and confirmed native teardown/unsubscribe support. The public BLE JS bridge is not yet wired.
- Add public non-BLE Scale registration through the real JS bridge, session-bound publications and failure reports, capability/handler validation, nullable battery, and first-sample controller handoff. Preserve existing Sensor registration.
- Fix immediate disconnect racing Scale initialization; add tests and work-in-progress documentation.

## Linked Issue

Refs #809. Related #809. The issue is open and labeled `ready-for-agent`; this draft must not close it.

## Verification

- Starting commit: `bc717af2089798b32dfd21ffc1227a808d4b38d2`; issue revision reviewed: 2026-09-07T12:02:20Z.
- `dart format lib test`: exit 0; unrelated formatter-only edits removed. The added direct-bridge test was subsequently formatted separately.
- `flutter analyze --no-pub`: exit 0, no issues found on the draft source tree.
- `flutter test --no-pub`: exit 0, 3927 passed, 1 skipped (2 minutes 27 seconds). Tested source is committed as `05e65d93`; raw local log: `.build/issue-809-draft-tests.log`.
- Real JS tests use the local QuickJS DLL; TLS tests use openssl on a process-local PATH. No system PATH changes.
- No API smoke test or Bookoo hardware test performed. Hardware availability, model, firmware, platform comparisons, and measured timing are unverified.

### Requirement-to-Test Map

| Contract area | Current evidence | Remaining acceptance |
| --- | --- | --- |
| A: retirement versus revocation | `plugin_ble_session_test.dart`, `universal_ble_transport_recovery_test.dart` | Complete bridge interleavings, link-loss/reconnect isolation, shutdown and leak coverage |
| B: uncertain ownership | `plugin_ble_matcher_test.dart`, `plugin_ble_registry_test.dart` | Discovery integration, startup gate, scan deadline diagnostics, remembered/native bypass audit |
| C: public non-BLE Scale | `plugin_manager_scale_test.dart`, `plugin_scale_test.dart`; existing Sensor bridge regressions | REST/WS proof, reload matrix, automatic optional calls and deliberate sleep policy |
| D: measurement timing | No acceptance evidence yet | Controlled identical native/full-JS traces, freshness/latency/control decisions; provenance amendment if required |
| BLE Sensor and Bookoo | No end-to-end evidence yet | Public bindDriver/GATT bridge, JS Bookoo reference/fixtures, coexistence, persistence and hardware comparison |

None of the four review findings is claimed fully resolved by this checkpoint.

## Impact

- Manifest/API schema accepts Scale capabilities, `transport.ble`, and bounded BLE matchers. Permission declarations alone grant no GATT access.
- Non-BLE Scale connect contexts capture session authority. Persistent Scale registrations cannot publish to a later session; unsupported optional operations fail explicitly.
- `ScaleSnapshot.batteryLevel` is nullable; existing serialized controller schemas already permit unknown battery. Native Bookoo remains available and unchanged.
- Native BLE transport gains unsubscribe and confirmed teardown. Unresolved native operations retain per-device exclusion; hardware behavior remains unverified.
- No new endpoints, migrations, dependencies, or estimator tuning. Publication-ingress timestamps remain provisional until the timing gate passes.
- Remaining work: BLE discovery/factory/startup wiring, stable native error mapping, API smoke coverage, optional automatic command handling, sleep/recovery policy, timing gate, Bookoo reference, preference/restart/resource isolation, and complete device-management documentation.
- The active design and requirement map remain in `doc/plans/issue-809-design.md` because implementation is unfinished. Archive rationale and remove completed implementation steps before marking ready.

## Contributor Responsibility

Left unchecked for the human submitter's review and attestation. This draft does not satisfy the merge-readiness gates.

### DIFF
--- a/lib/src/plugins/plugin_ble_matcher.dart
+++ b/lib/src/plugins/plugin_ble_matcher.dart
@@ -1,0 +1,102 @@
+import 'package:reaprime/src/models/device/ble_service_identifier.dart';
+
+enum PluginBleMatch { match, noMatch, indeterminate }
+
+String normalizePluginBleUuid(String value) {
+  final lower = value.toLowerCase();
+  if (RegExp(r'^[0-9a-f]{8}$').hasMatch(lower)) {
+    return '$lower-0000-1000-8000-00805f9b34fb';
+  }
+  try {
+    return BleServiceIdentifier.parse(lower).long;
+  } on ArgumentError {
+    throw const FormatException('Invalid BLE UUID');
+  }
+}
+
+class PluginBleMatcher {
+  final String? nameMode;
+  final String? nameValue;
+  final List<String>? serviceUuids;
+
+  const PluginBleMatcher._(this.nameMode, this.nameValue, this.serviceUuids);
+
+  factory PluginBleMatcher.fromJson(dynamic json) {
+    if (json is! Map ||
+        json.isEmpty ||
+        json.keys.any((key) => key != 'name' && key != 'serviceUuids')) {
+      throw const FormatException('Invalid BLE matcher');
+    }
+    String? nameMode;
+    String? nameValue;
+    if (json.containsKey('name')) {
+      final name = json['name'];
+      if (name is! Map ||
+          name.length != 1 ||
+          !const ['exact', 'prefix', 'contains'].contains(name.keys.single) ||
+          name.values.single is! String ||
+          (name.values.single as String).isEmpty ||
+          (name.values.single as String).length > 248) {
+        throw const FormatException('Invalid BLE name predicate');
+      }
+      nameMode = name.keys.single as String;
+      nameValue = name.values.single as String;
+    }
+    List<String>? services;
+    if (json.containsKey('serviceUuids')) {
+      final values = json['serviceUuids'];
+      if (values is! List ||
+          values.isEmpty ||
+          values.length > 64 ||
+          values.any((value) => value is! String)) {
+        throw const FormatException('Invalid BLE service UUID list');
+      }
+      services = List.unmodifiable(
+        values.cast<String>().map(normalizePluginBleUuid),
+      );
+    }
+    return PluginBleMatcher._(nameMode, nameValue, services);
+  }
+
+  PluginBleMatch evaluate({
+    String? name,
+    Iterable<String>? serviceUuids,
+    bool servicesComplete = true,
+  }) {
+    var unresolved = false;
+    if (nameMode != null) {
+      if (name == null) {
+        unresolved = true;
+      } else {
+        final lower = name.toLowerCase();
+        final expected = nameValue!.toLowerCase();
+        final matches = switch (nameMode) {
+          'exact' => lower == expected,
+          'prefix' => lower.startsWith(expected),
+          _ => lower.contains(expected),
+        };
+        if (!matches) return PluginBleMatch.noMatch;
+      }
+    }
+    final requiredServices = this.serviceUuids;
+    if (requiredServices != null) {
+      if (serviceUuids == null) {
+        unresolved = true;
+      } else {
+        final matches = serviceUuids
+            .map(normalizePluginBleUuid)
+            .any(requiredServices.contains);
+        if (!matches) {
+          if (servicesComplete) return PluginBleMatch.noMatch;
+          unresolved = true;
+        }
+      }
+    }
+    return unresolved ? PluginBleMatch.indeterminate : PluginBleMatch.match;
+  }
+
+  Map<String, dynamic> toJson() => {
+    if (nameMode != null) 'name': {nameMode!: nameValue},
+    if (serviceUuids != null) 'serviceUuids': serviceUuids,
+  };
+}
--- a/lib/src/plugins/plugin_ble_matcher.dart
+++ b/lib/src/plugins/plugin_ble_matcher.dart
@@ -60,12 +60,14 @@ class PluginBleMatcher {
 
   PluginBleMatch evaluate({
     String? name,
+    bool nameComplete = false,
     Iterable<String>? serviceUuids,
     bool servicesComplete = true,
   }) {
     var unresolved = false;
     if (nameMode != null) {
       if (name == null) {
+        if (nameComplete) return PluginBleMatch.noMatch;
         unresolved = true;
       } else {
         final lower = name.toLowerCase();
--- a/lib/src/plugins/plugin_ble_registry.dart
+++ b/lib/src/plugins/plugin_ble_registry.dart
@@ -10,6 +10,7 @@ enum BleEvidenceSource { advertisement, system }
 
 class BleAdvertisementEvidence {
   final String? name;
+  final bool nameComplete;
   final List<String>? serviceUuids;
   final bool servicesComplete;
   final BleEvidenceSource source;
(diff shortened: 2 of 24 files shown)


lib/src/plugins/plugin_ble_matcher.dart
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  PluginBleMatch evaluate({
+    String? name,
+    Iterable<String>? serviceUuids,
+    bool servicesComplete = true,
+  }) {
+    var unresolved = false;
+    if (nameMode != null) {
+      if (name == null) {
```
tadelv, line 69: Complete absence of a name needs to be distinguishable from unavailable/incomplete name metadata. Right now `name == null` is always `indeterminate`, but #809's tri-state contract says a complete observation proving a required field is absent is `noMatch`. A name-only plugin matcher would therefore keep a genuinely nameless complete advertisement pending and suppress native fallback until the deadline. Please carry name completeness into the evidence/evaluation path (or equivalent) and add coverage for complete-name-absent => `noMatch` versus unavailable-name => `indeterminate`.

lib/src/plugins/plugin_ble_registry.dart
```diff
@@ -0,0 +1,215 @@
...
+
+  void record(
+    String physicalId,
+    int generation,
+    BleAdvertisementEvidence evidence,
+  ) {
+    if (generation != _generation) return;
+    final key = normalizeBleDeviceId(physicalId);
+    final previous = _observations[key];
+    if (previous != null &&
```
tadelv, line 57: This source-precedence rule protects *every* advertisement from a later system observation, even when the advertisement is incomplete. The approved design only protects complete advertisement evidence from being erased; it also says complete observations replace older evidence. If an incomplete advertisement arrives first and a later system result contains the complete required fields, this cache will retain the incomplete observation and can leave ownership pending unnecessarily. Please make replacement completeness-aware without merging observations, and cover incomplete-advertisement -> complete-system (plus the reverse order) explicitly.

tadelv (changes requested):
Two required fixes in the tri-state BLE ownership evidence model. Both can keep ownership artificially `pending` even when the available evidence should allow a definite decision under #809. The rest of this draft looks directionally consistent with the repository guidance for its current WIP stage; the documented smoke/hardware/timing/discovery gaps remain merge-readiness work rather than additional code-review findings here.

lib/src/plugins/plugin_ble_matcher.dart
ODevStudio: Fixed in 55193d20. Evidence now carries nameComplete into matcher evaluation: complete absent name returns noMatch; unavailable name remains indeterminate. Added matcher and registry ownership regression coverage.

lib/src/plugins/plugin_ble_registry.dart
ODevStudio: Fixed in 55193d20. Cache replacement now prefers complete observations regardless of source, then uses observation time between equally complete records. It never merges fields. Tests cover incomplete advertisement and complete system in both orders, the converse, known absence, and freshness. Full suite: 3936 passed, 1 skipped; analysis and formatting clean. PR remains draft.

lib/src/plugins/plugin_manager.dart
```diff
@@ -1299,16 +1348,36 @@ class PluginManager {
     final retiredConnectInvocations = <String>[];
     if (operation == PluginDeviceOperation.disconnect) {
       final key = (pluginId, generation, registrationHandle);
-      retiredConnectInvocations.addAll(
-        _timedOutDeviceConnects.remove(key) ?? const <String>{},
-      );
+      retiredConnectInvocations.addAll({
+        ...?_timedOutDeviceConnects.remove(key),
+        if (payload['session'] is String)
+          ..._transportService.deviceConnectInvocations(
```
tadelv, line 1354: This still recovers a completed Scale connect invocation only from *currently live* transport records. After a successful connect, `_deviceConnectAttempts` is removed; if the plugin opened no transport, or opened one and closed it before disconnect, `deviceConnectInvocations()` returns nothing and no retirement fence is installed. The old captured `context.transport` still carries that invocation ID, so a delayed callback can call `open()` after disconnect/reconnect and regain transport authority. That violates the session-bound authority guarantee. Please retain the Scale connect invocation ID by session/registration until retirement independently of live records, and add a regression where connect completes with zero live transports at disconnect and the old context's later `transport.open()` is rejected.

test/plugins/plugin_manager_scale_test.dart
```diff
@@ -0,0 +1,423 @@
...
+            .then((_) {
+              disconnectSettled = true;
+            });
+        await firstFinished;
+        var reconnected = false;
+        final replacement = scale.onConnect().then((_) {
+          reconnected = true;
+        });
+        await Future<void>.delayed(
+          scale.invocationTimeout + const Duration(milliseconds: 200),
```
tadelv: These regressions now spend production-scale wall time to prove an ordering property: two cases each wait through the manager's 10 s cleanup window, and the current PR CI is failing `Gate slow test suites` (20,000 ms active-time cap) even though `flutter test` itself passes. This also conflicts with `AI_TESTING_NOTES`: use injected short deadlines for behavior after a delay unless the exact production duration is what the test is validating. Here the contract is that reconnect stays blocked until manager retirement completes, not that five/ten real seconds elapse. Please keep the same ordering assertions with short injected timing so the suite passes the CI gate.

tadelv (changes requested):
The previous ownership-evidence findings are fixed. Two required issues remain: completed Scale connect authority is still not retained independently of live transport records, allowing a stale connect context to regain transport access in a normal no-live-handle case; and the new production-duration regression tests currently fail the repository's slow-suite CI gate. Please address both before approval.

lib/src/plugins/plugin_manager.dart
ODevStudio: This path is already rejected by the manager: hasDeviceConnectClaim && !ownsDeviceConnect returns "Plugin device connect retired" before calling the transport service. Removing the completed invocation revokes its authority; absence of a transport retirement fence does not restore it. Added a real-bridge regression in the latest commit: a permitted plugin saves its first connect context, completes disconnect/reconnect with zero transports and zero attempt/fence counts, and its old open receives that exact retirement error. No production change or retained session bookkeeping was needed. Full suite: 3941 passed, 1 skipped; runtime gate, analysis and formatting clean.

test/plugins/plugin_manager_scale_test.dart
ODevStudio: Fixed in 2f554158. The fixtures inject a 250 ms adapter deadline and 1 s manager deadline, and check at 300 ms that reconnect remains blocked. Production defaults are unchanged; the ordering assertions and real WebSocket checks remain intact. CI passed on 2f554158, and the full local suite including the subsequent zero-transport regression also passes the 20000 ms gate.

tadelv (approved):
Re-reviewed the updates since ae8e6f7. The CI-runtime blocker is fixed with injected short deadlines while preserving the ordering assertions, and the zero-live-transport stale-open concern is covered by the manager's existing connect-claim revocation path plus a real-bridge regression. Current PR checks pass at f0818b2. I found no new required fixes in this pass; the explicitly documented #809 integration/hardware/timing gaps remain future work for this WIP series rather than blockers in the code reviewed here.

tadelv (approved):

PR #821, opened 2026-09-08
Plugin Scale API and sleep-policy checkpoint

ODevStudio:
## Summary

- Complete the next Scale integration checkpoint on top of merged #820. The branch now contains current `main` (`c957e2b1`) and this PR targets `main`, so the review diff is limited to the #821 checkpoint.
- Observe automatic timer failures, preserve explicit Scale command error codes through a domain-facing error contract, and pause host recovery for disconnect-to-sleep until machine wake.
- Add real QuickJS non-BLE Scale inventory/selection/HTTP/WebSocket/reload coverage and host sleep-policy regressions. No BLE permission is needed for the fixture.

## Linked Issue

Refs #809. Builds on merged #820. Does not close the issue.

## Verification

- Previous full `flutter test --no-pub --machine`: 3,995 passed, one skipped. CI active-time gate passed. Local evidence: `.build/scale-integration-full.json`.
- Previous `flutter analyze --no-pub`: no issues. CI-compatible formatting and `git diff --check` passed.
- Final-head CI is rerunning after the restack and review fixes.
- Real HTTP/WebSocket clients exercise production handlers with a real JS registration and test-owned in-memory device. The Scale is now selected through `ConnectionManager`, reaches `ScaleController`, and persists its stable public ID through normal preference policy. This is not a full application UI or hardware run.
- Test-first regressions reproduced unhandled automatic timer failures and missing REST permission-independent optional-operation codes.

## Impact

- Existing Scale routes expose stable operation `code` values through a Scale-domain exception contract; `PluginScale` adapts plugin failures into that contract, including `unsupported_operation`. Existing HTTP status/message and native timer behavior are preserved.
- Display-off sleep of a plugin with `disconnectToSleep` pauses recovery until machine wake. Protocol failures outside deliberate sleep retain normal recovery.
- Documentation/spec updated. Measurement timing, Bookoo protocol/persistence proof, full-app smoke and hardware acceptance remain later work. Keep draft.

## Contributor Responsibility

- [x] I have reviewed and understand all changes in this PR and take responsibility for their correctness, security, behavior, licensing, and provenance, including any AI-assisted or AI-generated work.

PR #820, opened 2026-09-08
Plugin BLE binding and Sensor discovery checkpoint

ODevStudio:
## Summary

- Add `host.devices.bindDriver()` and host-owned BLE bindings with synchronous factories, session-scoped GATT/publication, and bounded teardown.
- Wire plugin-first ownership into the existing scanner, system results, watch, startup readiness, and native connection admission. Pending/conflict cannot select native; unconfirmed teardown retains physical exclusion.
- Prove a BLE Sensor through real JavaScript, existing DeviceController/SensorController selection, HTTP inventory/commands, and WebSocket snapshots with a fake BLE edge. Reuse the Scale adapter without claiming Bookoo or Scale timing acceptance.

This is the review-ready BLE binding and Sensor checkpoint after merged #813, not completion of #809. Its merge boundary is host-owned BLE arbitration/lifecycle and Sensor integration through existing controllers and APIs, verified with a fake BLE edge.

## Linked Issue

Refs #809. Keep the issue open.

## Verification

- Formatter: 818 files, zero changes using the CI-compatible dart_style 3.1.13 formatter.
- `flutter analyze --no-pub`: no issues.
- Full `flutter test --no-pub --machine`: 3,992 passing tests, one skip. The 20-second per-suite active-time gate passes.
- Regression coverage includes failed initial plugin loading, both evidence orders, conflict/pending deadline, native candidate fencing, failed handshake without native fallback, repeated permanently hung connects, throwing/hung cleanup, unconfirmed native teardown, capacity/isolation, stale/foreign authority, and terminal delivery on link loss/revocation.
- Real UniversalBleTransport bridge test verifies notification -> disabled -> notification, UUID aliases, old logical unsubscribe isolation, missing attributes, cancellation, and acknowledged-write errors without downgrade.
- Windows debug build (`lib/main.dart`, simulate=1) passed using CMake 3.28 and Visual Studio 2022 in a separate build directory. Flutter's default VS2019 toolchain failed on native dependencies; no global toolchain settings or dependency sources were changed.
- Desktop runtime smoke: an isolated Flutter entrypoint using the production manager, discovery, controllers, and Sensor routes with fake BLE/in-memory settings passed. HTTP inventory reported the connected Sensor; command returned `{"status":"ok","result":{"humidity":52}}`; a fresh WebSocket notification returned `{"humidity":57}`. Runtime exited after teardown. No real BLE/USB devices or saved user data were used. This is not full application UI/hardware acceptance.
[...]

PR #822, opened 2026-09-08
Plugin Scale measurement-timing checkpoint

### LINKED ISSUE #809: Plugin drivers: host-owned BLE binding + plugin-backed Scale support

ODevStudio:
## Summary

- Add the measurement-timing checkpoint directly on current `main` after merged #821.
- Preserve BLE notification ingress time through optional opaque sample provenance, without accepting JS timestamps or changing estimator tuning.
- Bound/session-fence provenance and reject duplicates, foreign/expired/out-of-order samples; protect Scale timestamp ordering across publications.
- Preserve #821's Scale-domain command error mapping while narrowing recoverable `stale_sample` handling to host-originated provenance failures from the active BLE callback invocation.

## Linked Issue

Refs #809, amendment D of the approved assignment. Builds on merged #821 and #820. Does not close #809.

## Verification

- Test-first full-JS trace reproduced first sample 100 ms becoming 400 ms in a four-notification burst before the fix.
- Native/full-JS ScaleController comparisons cover 100 ms cadence, one/four-event batches, zero/150 ms asynchronous delays, repeated backlog recovery, exact timestamps/order and flow equality within 1e-9. Tolerances were recorded before implementation; estimators are unchanged.
- Recorded accepted controller samples drive the existing ShotSequencer with equal stop sample indices; three-second-old delivery cannot trigger SAW. This replay is not proof of real-time hardware reaction during a JS stall.
- Regression coverage requires expired host provenance to remain recoverable while a plugin-created error that also uses `code: "stale_sample"` remains fatal.
- Previous pre-retarget full suite: 4,002 passed, one skipped; CI active-time gate passed. Evidence: `.build/timing-full.json`. Previous analysis, formatting and diff checks passed.
- Final-head `PR checks` run #1013 passed on `f41b662eb61e17c2017297fb5f65d017d45a805c`: Dart format, Flutter analyze, Flutter test, slow-suite gate, contribution policy, and Linux build smoke are green.
- No physical-device or full application UI run is required in this checkpoint.

## Impact

- BLE callback gains optional second `sample` argument; `publish(snapshot, sample)` resolves its host timestamp. Existing one-argument code remains compatible, with publication-ingress fallback for synthetic/non-BLE samples.
- Tokens expire at two seconds, are limited to 256 per session, and are invalidated on retirement/clock rollback. Invalid tokens produce `stale_sample`; no native/JS timestamps supplied by plugins are trusted.
- Only the host-originated stale provenance rejection from the active callback invocation is treated as a recoverable dropped sample. Arbitrary plugin callback failures, including plugin-created `stale_sample` errors, remain fatal.
- Bookoo reference/persistence proof and hardware/full-app acceptance remain separate later gates.

## Contributor Responsibility

### DIFF
--- a/doc/AI_BLE_NOTES.md
+++ b/doc/AI_BLE_NOTES.md
@@ -445,6 +445,12 @@ cancellation or fencing.
 
 ## Plugin BLE Binding (#809 Checkpoint)
 
+Notification provenance is captured before JS dispatch. An optional opaque token
+round-trips through the callback and publication; host validation supplies the
+Scale timestamp. Four-event/100 ms trace tests reproduced collapsed timestamps
+without it. Keep the two-second expiry aligned with shot freshness, retain bounded
+session ownership, and measure arrival delay separately from timestamp quality.
+
 `PluginBleBinding` reserves the physical ID before transport creation and owns a
 fresh `PluginBleSession` per connect. Factory metadata has no mutable publication
 target: all publication and GATT closures capture a session capability. Do not
--- a/doc/Plugins.md
+++ b/doc/Plugins.md
@@ -1,6 +1,30 @@
 
 # Decaid Plugin Development Guide
 
+## BLE Scale Sample Time
+
+BLE notification callbacks receive `(base64Data, sample)`. The optional opaque
+`sample` token identifies a host-timestamped notification. After decoding, use
+`await session.publish({weight: grams}, sample)` to retain its ingress time even
+when JavaScript dispatch/publication is delayed. No timestamp supplied by JS is
+trusted. Existing one-argument callbacks/publications remain compatible.
+
+Tokens are binding/session-owned, single-use, ordered, and expire after two
+seconds (the existing shot freshness window). At most 256 tokens are retained;
+the oldest token is discarded when that bound is reached, without dropping the
+notification itself. A foreign, duplicate, out-of-order, evicted, or expired token
+returns `stale_sample`. That rejection remains visible to plugin code, but if it
+escapes a notification callback the host drops that sample and continues the
+subscription. Other callback failures remain fatal. Retirement invalidates all
+tokens. A backward clock change retires the BLE session rather than leaving
+publication blocked behind its previous timestamp. Reconnect establishes a fresh
+timestamp sequence and follows normal Scale recovery policy.
+
+Omitting the token retains publication-ingress time for non-BLE or synthetic
+measurements. Do not use that fallback to disguise delayed notification samples.
+Accurate sample timestamps do not reduce delivery latency or recover stop commands
+missed while JavaScript was stalled.
+
 Plugin Scale commands report stable error codes through the existing Scale REST
 routes, including `unsupported_operation` for undeclared tare or timer support.
 Automatic shot timer failures are logged without aborting the shot. A Scale with
--- a/doc/plans/archive/issue-809-scale-timing/timing-gate.md
+++ b/doc/plans/archive/issue-809-scale-timing/timing-gate.md
@@ -1,0 +1,36 @@
+# Scale Timing Gate
+
+Keep the existing ScaleController estimators and tuning unchanged. Use 100 ms
+cadence, as in the native estimator fixtures, with identical 0.1 g increments.
+Require exact sample order and timestamps, and display/control flow equality
+within 1e-9 (identical inputs into the same deterministic estimators).
+
+Exercise immediate delivery, dispatch held for a four-notification burst, and
+publication delayed asynchronously after callback dispatch. Compare every accepted
+sample, not just final flow. Record delivery age separately from sample time;
+correct timestamps do not recover real-time decisions missed during a stall.
+
+The shot sequencer's existing freshness window is two seconds. Provenance must
+expire at that boundary, reject clock rollback/future timestamps, be single-use,
+and reject foreign/session/out-of-order tokens. Bound retained tokens to 256 per
+session. Retirement clears them. Non-BLE/synthetic publications retain host
+publication-ingress time. JS must never supply authoritative timestamps.
+
+Ingress timestamps failed: the first 100 ms sample in a four-event batch was
+published at 400 ms. Added only an optional opaque notification token as
+the second callback argument and second publish argument. Validate it host-side
+before forwarding to the shared Scale adapter. Keep protocol decoding in JS.
+
+The regression now compares the full JS path with native ScaleController output
+for batches of one/four and asynchronous delays of zero/150 ms over four recovery
+cycles. Recorded accepted samples drive the existing ShotSequencer: the stop
+sample agrees, while samples delivered three seconds late cannot trigger SAW.
+This replay proves timestamp/control-input equivalence, not that stalled delivery
+can stop hardware on time. Live device latency remains a hardware gate.
+
+Live QuickJS callbacks also remain suspended while the injected host clock advances
+three seconds. Expired publication rejects with `stale_sample` without retiring
+the subscription; a fresh notification still publishes. Unrelated callback
+failures remain fatal. Clock rollback at either capture or publication retires
+the BLE session, and reconnect accepts the earlier timestamp in a new connection.
+Capture happens before notification encoding.
--- a/lib/src/plugins/plugin_ble_binding.dart
+++ b/lib/src/plugins/plugin_ble_binding.dart
@@ -166,9 +166,19 @@ class PluginBleBinding {
     return session.call(authority, operation, args);
   }
 
-  void publish(Map<String, dynamic> snapshot, String? domainSession) {
+  void publish(
+    Map<String, dynamic> snapshot,
+    String? domainSession, {
+    String? sample,
+  }) {
     _checkPublication(domainSession);
-    device.publish(snapshot, session: domainSession);
+    final timestamp = sample == null ? null : _session!.consumeSample(sample);
+    final target = device;
+    if (target is PluginScale) {
+      target.publish(snapshot, session: domainSession, timestamp: timestamp);
+    } else {
+      target.publish(snapshot, session: domainSession);
+    }
   }
 
   void reportDisconnected(String? domainSession) {
--- a/lib/src/plugins/plugin_ble_bridge.dart
+++ b/lib/src/plugins/plugin_ble_bridge.dart
@@ -19,11 +19,19 @@ const __bindBleDriver = (driverId, factory) => {
     const handle = payload.registrationHandle;
     const sessions = new Map();
     const cleanups = new Set();
+    const recoverableSampleErrors = new WeakMap();
     const stale = () => Object.assign(new Error('BLE session retired'), {code: 'stale_session'});
     const context = (payload, cleanup) => {
       const authority = payload.gattSession;
       const callbacks = new Map();
-      const record = { callbacks, disconnected: false, delivered: false, onDisconnect: null };
+      const record = {
+        callbacks,
+        disconnected: false,
+        delivered: false,
+        onDisconnect: null,
+        callbackEpoch: 0,
+        activeCallbackEpoch: 0
+      };
       if (!cleanup) sessions.set(authority, record);
       else cleanups.add(record);
       const call = (operation, args = {}) => record.disconnected
@@ -67,8 +75,18 @@ const __bindBleDriver = (driverId, factory) => {
       if (cleanup) return Object.freeze({gatt});
       return Object.freeze({
         gatt,
-        publish: snapshot => record.disconnected ? Promise.reject(stale()) : __deviceCall('blePublish', {
-          registrationHandle: handle, session: payload.session, snapshot
+        publish: (snapshot, sample) => record.disconnected ? Promise.reject(stale()) : __deviceCall('blePublish', {
+          registrationHandle: handle, session: payload.session, snapshot, sample
+        }).catch(error => {
+          if (sample != null && error && error.code === 'stale_sample' &&
+              (typeof error === 'object' || typeof error === 'function') &&
+              record.activeCallbackEpoch !== 0) {
+            recoverableSampleErrors.set(error, {
+              record,
+              epoch: record.activeCallbackEpoch
+            });
+          }
+          throw error;
         }),
         reportDisconnected: () => record.disconnected ? Promise.reject(stale()) : __deviceCall('bleDisconnected', {
           registrationHandle: handle, session: payload.session
@@ -93,7 +111,23 @@ const __bindBleDriver = (driverId, factory) => {
         return;
       }
       const callback = record.callbacks.get(event.listener);
-      if (callback && !record.disconnected) await callback(event.data);
+      if (callback && !record.disconnected) {
+        const epoch = ++record.callbackEpoch;
+        record.activeCallbackEpoch = epoch;
+        try {
+          await callback(event.data, event.sample);
+        } catch (error) {
+          const provenance = error &&
+            (typeof error === 'object' || typeof error === 'function')
+              ? recoverableSampleErrors.get(error) : null;
+          if (!provenance || provenance.record !== record || provenance.epoch !== epoch) {
+            throw error;
+          }
+          recoverableSampleErrors.delete(error);
+        } finally {
+          if (record.activeCallbackEpoch === epoch) record.activeCallbackEpoch = 0;
+        }
+      }
     };
     __deviceSetHandlers(handle, {
       pluginId, generation: pluginGeneration, bridgeToken: pluginBridgeToken,
--- a/lib/src/plugins/plugin_ble_service.dart
+++ b/lib/src/plugins/plugin_ble_service.dart
@@ -193,8 +193,13 @@ class PluginBleService {
     int generation,
     String handle,
     Map<String, dynamic> snapshot,
-    String? session,
-  ) => _binding(pluginId, generation, handle).publish(snapshot, session);
+    String? session, {
+    String? sample,
+  }) => _binding(
+    pluginId,
+    generation,
+    handle,
+  ).publish(snapshot, session, sample: sample);
 
   void reportDisconnected(
     String pluginId,
--- a/lib/src/plugins/plugin_ble_session.dart
+++ b/lib/src/plugins/plugin_ble_session.dart
@@ -9,6 +9,7 @@ import 'package:reaprime/src/models/device/transport/ble_transport.dart';
 import 'package:uuid/uuid.dart';
 
 import 'plugin_ble_matcher.dart';
+import 'plugin_sample_provenance.dart';
 
 enum PluginBleSessionState {
   connecting,
@@ -53,6 +54,9 @@ class PluginBleSession {
   final Map<(String, String), _BleSubscription> _subscriptions = {};
   final Set<Completer<Object?>> _pending = {};
   final Queue<(Map<String, dynamic>, int)> _notifications = Queue();
+  late final PluginSampleProvenance _samples = PluginSampleProvenance(
+    onClockRollback: revoke,
+  );
   final Completer<void> _closed = Completer<void>();
   final Completer<void> _cleanupInterrupted = Completer<void>();
   StreamSubscription<ConnectionState>? _connectionSubscription;
@@ -81,6 +85,13 @@ class PluginBleSession {
   int get subscriptionCount => _subscriptions.length;
   int get pendingOperationCount => _pending.length;
   int get queuedEventCount => _notifications.length;
+  DateTime consumeSample(String token) {
+    if (!acceptsPublications) {
+      throw const PluginBleException('stale_session', 'BLE session retired');
+    }
+    return _samples.consume(token);
+  }
+
   bool get acceptsPublications =>
       authorized() &&
       runtimeAlive() &&
@@ -350,6 +361,11 @@ class PluginBleSession {
   }
 
   void _enqueue(_BleSubscription subscription, Uint8List bytes) {
+    final sample = _samples.capture();
+    if (!acceptsPublications) {
+      _samples.clear();
+      return;
+    }
     if (_notifications.length >= maxQueuedEvents ||
         _queuedBytes + bytes.length > maxQueuedBytes) {
       _log.warning(
@@ -365,6 +381,7 @@ class PluginBleSession {
         'subscription': subscription.id,
         if (subscription.listener != null) 'listener': subscription.listener,
         'data': base64Encode(bytes),
+        'sample': sample,
       },
       bytes.length,
     ));
@@ -406,6 +423,7 @@ class PluginBleSession {
         _state == PluginBleSessionState.initializing ||
         _state == PluginBleSessionState.ready;
     _state = PluginBleSessionState.retiring;
+    _samples.clear();
     _notifications.clear();
     _queuedBytes = 0;
     _rejectPending();
--- a/lib/src/plugins/plugin_manager.dart
+++ b/lib/src/plugins/plugin_manager.dart
@@ -1292,7 +1292,8 @@ class PluginManager {
           _replyDevice(requestId, bridgeToken, result: {'value': value});
         case 'blePublish':
           final snapshot = data['snapshot'];
-          if (snapshot is! Map) {
+          final sample = data['sample'];
+          if (snapshot is! Map || (sample != null && sample is! String)) {
             throw const PluginBleException(
               'invalid_argument',
               'Invalid BLE snapshot',
@@ -1304,6 +1305,7 @@ class PluginManager {
             registrationHandle,
             Map<String, dynamic>.from(snapshot),
             data['session'] as String?,
+            sample: sample as String?,
           );
           _replyDevice(requestId, bridgeToken, result: const {});
         case 'bleDisconnected':
--- a/lib/src/plugins/plugin_sample_provenance.dart
+++ b/lib/src/plugins/plugin_sample_provenance.dart
@@ -1,0 +1,51 @@
+import 'package:clock/clock.dart';
+import 'package:uuid/uuid.dart';
+
+import 'plugin_device_contract.dart';
+
+class PluginSampleProvenance {
+  final void Function() onClockRollback;
+  PluginSampleProvenance({required this.onClockRollback});
+  static const capacity = 256;
+  static const lifetime = Duration(seconds: 2);
+  final Map<String, (int, DateTime)> _samples = {};
+  int _sequence = 0;
+  int _lastConsumed = 0;
+  DateTime? _lastClock;
+
+  int get count => _samples.length;
+
+  DateTime _now() {
+    final now = clock.now();
+    if (_lastClock != null && now.isBefore(_lastClock!)) {
+      clear();
+      onClockRollback();
+    }
+    _lastClock = now;
+    _samples.removeWhere((_, sample) => now.difference(sample.$2) >= lifetime);
+    return now;
+  }
+
+  String capture() {
+    final now = _now();
+    if (_samples.length == capacity) _samples.remove(_samples.keys.first);
+    final token = const Uuid().v4();
+    _samples[token] = (++_sequence, now);
+    return token;
+  }
+
+  DateTime consume(String token) {
+    _now();
+    final sample = _samples.remove(token);
+    if (sample == null || sample.$1 <= _lastConsumed) {
+      throw const PluginDeviceException(
+        'Invalid or expired sample provenance',
+        code: 'stale_sample',
+      );
+    }
+    _lastConsumed = sample.$1;
+    return sample.$2;
+  }
+
+  void clear() => _samples.clear();
+}
--- a/lib/src/plugins/plugin_scale.dart
+++ b/lib/src/plugins/plugin_scale.dart
@@ -16,6 +16,7 @@ class PluginScale extends PluginProtocolDevice
   final List<ScaleSnapshot> _handoff = [];
   Completer<void> _firstWeight = Completer<void>();
   bool _active = false;
+  DateTime? _lastTimestamp;
 
   PluginScale({
     required super.deviceId,
@@ -39,6 +40,7 @@ class PluginScale extends PluginProtocolDevice
     _firstWeight = Completer<void>();
     _handoff.clear();
     _active = false;
+    _lastTimestamp = null;
   }
 
   @override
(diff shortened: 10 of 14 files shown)


tadelv (changes requested):
The measurement-timing implementation itself looks sound in this pass. The host remains the timestamp authority, provenance is session-local/bounded/single-use, delayed callbacks preserve ingress timestamps, stale provenance is recoverable only when it is the exact host error escaping the active callback, plugin-created `stale_sample` failures remain fatal, and `PluginScale` prevents timestamp regression before samples reach the existing estimators. The native-vs-full-JS timing/flow comparisons and ShotSequencer freshness replay cover the first-order contract without changing estimator tuning.

I have one remaining merge blocker, which is now actionable because #821 has landed:

**Restack/retarget #822 onto current `main` and run final-head CI.** #821 is merged and `main` is now `1619bb1`, but #822 still targets `odev/issue-809-scale-integration` at `69bb3f5e`. The PR workflow only runs for PRs targeting `main`, so the current head has no final-head workflow run despite the body saying CI is pending. Please retarget/rebase this checkpoint onto current `main`, confirm the resulting review diff remains just the timing checkpoint, and get the normal format/analyze/full-test/slow-suite checks green on that final head.

I did not find another merge-blocking issue in the provenance lifecycle, backlog/expiry behavior, callback error fencing, Scale timestamp ordering, or the estimator/ShotSequencer integration. No physical-device run is required for this checkpoint; the later Bookoo/hardware gate can remain separate as documented.

tadelv (approved):
Re-reviewed latest head `f41b662e`. The remaining blocker is resolved: #822 is now a single focused timing commit directly on current `main`, the diff is back to the expected 14-file checkpoint, and final-head PR checks run #1013 is green. Dart format, Flutter analyze, full Flutter test, slow-suite gate, contribution policy, and Linux build/AppImage smoke all pass. I did not find a new merge-blocking regression in the provenance lifecycle, callback `stale_sample` fencing, Scale timestamp ordering, or native-vs-plugin timing/ShotSequencer coverage. Approved.

PR #827, opened 2026-09-09
feat(api): read the firmware stop-at-weight target

ChampionDesigns:
## Summary

`GET /api/v1/machine/capabilities` already reports `stopAtWeight`, but no
endpoint let a client read what the machine actually holds. The app writes that
target from `WorkflowContext.targetYield`, its single source of truth, so until
now a client could only infer the firmware value from the workflow it last sent.

`GET /api/v1/machine/stopAtWeight` returns `{"grams": n}`. `0` means
stop-at-weight is off. It sits behind the existing `_bengleFirmwareGate` and
answers `404` on a plain DE1.

`getStopAtWeightTarget` already exists on `BengleInterface`, and so does the
gate, so the handler is eight lines.

### History

This started inside #791, which is about a different question — whether an
absent `targetYield` should write `0` — and it had no spec entry there. It is
split out and specified.

## Linked Issue

N/A

## Verification

- `flutter analyze` — clean.
- `flutter test` — full suite **3968 passed / 1 skipped / 0 failed**.
- `flutter test test/services/webserver/de1handler_stop_at_weight_test.dart` —
  4 passed: the target in grams, `0` when the feature is off, `404` on a plain
  DE1, and a check that the spec documents the route.
- `dart format --output=none --set-exit-if-changed` — clean on both changed
  files, run with Dart 3.13.3, the version CI resolves on `stable`.

## Impact

- **User-visible**: none. The endpoint is read-only and additive.
- **API**: adds `GET /api/v1/machine/stopAtWeight`. `assets/api/rest_v1.yml` and
  `doc/Api.md` both carry it. One of the tests reads the spec and fails if the
  route or its `grams` response is ever dropped, so the pair cannot drift.
- **Compatibility**: additive. Nothing existing changes.
- **Security**: none. It reads one firmware register behind the existing Bengle
  gate.

## Contributor Responsibility

### DIFF
--- a/assets/api/rest_v1.yml
+++ b/assets/api/rest_v1.yml
@@ -538,6 +538,39 @@ paths:
           description: Machine write queue is full
         "500":
           description: No machine was connected, the machine changed during the write, or the write failed
+  /api/v1/machine/stopAtWeight:
+    get:
+      summary: Get the firmware stop-at-weight target
+      description: |
+        Read the stop-at-weight target the machine currently holds, in grams.
+        The app writes this target from `WorkflowContext.targetYield`, which is
+        its single source of truth, so this endpoint reports what the firmware
+        actually has rather than what the app last sent. `0` means
+        stop-at-weight is off. Returns 404 when the connected machine is not a
+        Bengle. The `stopAtWeight` capability on
+        `GET /api/v1/machine/capabilities` reports whether this endpoint is
+        available.
+      tags: [Machine]
+      responses:
+        "200":
+          description: Current stop-at-weight target
+          content:
+            application/json:
+              schema:
+                type: object
+                required: [grams]
+                properties:
+                  grams:
+                    type: number
+                    format: double
+                    minimum: 0
+                    description: |
+                      Target beverage weight in grams. `0` means stop-at-weight
+                      is off.
+                    example: 36.0
+        "404":
+          description: Machine connected but does not support stop-at-weight
+
   /api/v1/machine/ledStrip:
     get:
       summary: Get LED strip configuration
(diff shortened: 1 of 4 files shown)


assets/api/rest_v1.yml
```diff
@@ -538,6 +538,39 @@ paths:
           description: Machine write queue is full
         "500":
           description: No machine was connected, the machine changed during the write, or the write failed
+  /api/v1/machine/stopAtWeight:
+    get:
+      summary: Get the firmware stop-at-weight target
+      description: |
+        Read the stop-at-weight target the machine currently holds, in grams.
+        The app writes this target from `WorkflowContext.targetYield`, which is
```
tadelv, line 546: Blocking design/scope question: this sentence says `WorkflowContext.targetYield` is the single source of truth, but the new public endpoint gives clients a second observable SAW state (`what the firmware actually has`). That may be useful for diagnostics, but #791 previously rejected the `getSaw` direction as overengineered/not the right fix, and this PR does not establish an independent consumer or debugging requirement that justifies committing a permanent API contract. Please document/obtain acceptance for the use-case and explain why clients should read firmware state instead of the workflow source of truth; otherwise I would not add this endpoint.

tadelv (changes requested):
Requesting changes. The handler/test implementation is small and consistent with the neighboring Bengle-only endpoints, and CI is green, but I don't think this PR currently clears the repository's design and verification gates.

1. The public API addition is not independently justified. This endpoint was split out of #791 after the `getSaw` direction there was called overengineered/not the right direction. The accepted SAW design makes `WorkflowContext.targetYield` the single source of truth specifically to avoid a separate machine SAW control surface. A read-only endpoint does not create a second writer, but it does create a second observable state and a permanent API contract. Please establish an accepted consumer/debugging use-case for exposing firmware state rather than workflow state (ideally in an issue/design discussion), or drop the endpoint.
2. `CONTRIBUTING.md` requires an end-to-end scenario for API surface changes. This PR adds only handler/spec unit coverage. Please add/extend the relevant `.agents/skills/decent-app/scenarios/` scenario to exercise the route through the running API, including the supported Bengle value and unsupported-machine behavior.
3. `AGENTS.md` explicitly requires API/spec changes to be smoke-tested through `scripts/sb-dev.sh` plus `curl`/`websocat`. The PR verification section has no such evidence. Please run and document the live smoke test.
4. The PR template's contributor-responsibility checkbox is still unchecked. `AGENTS.md` makes completing the PR template a pre-PR hard gate; please acknowledge it before merge.

I did not find a separate merge-blocking correctness bug in the eight-line handler itself. The blocker is whether this API should exist at all, plus the repository-required API verification that is currently missing.

PR #844, opened 2026-09-13
feat(plugins): expose connected scale metadata

MaxRink:
## Summary

Expose session-scoped opaque firmware and nullable battery metadata for connected plugin scales, with the existing narrow `/api/v1/scale/info` projection as the public boundary.

- Add the typed `PluginProtocolDevice.connectionId` identity used by later guarded actions.
- Fence metadata publication to the owning connected session and clear it when that session retires.
- Keep USB provenance, arbitrary extension metadata, and metadata out of device inventory.

Depends on the retained native scale-info boundary in #743.

## Linked Issue

Related #852, #713, and #714. Depends on the retained scale-info boundary in #743; consumed by #846.

The new discussion issue is awaiting maintainer triage. This PR remains a draft while maintainers discuss the proposal; opening the issue does not satisfy the accepted-issue gate or imply approval of the proposed contract.

## Verification

- Combined Linux Flutter gate for the current publication stack: **4,315 passed, 1 skipped, 0 failed**; analyzer clean (**48.0s**). Source was unchanged during the gate.

- Combined simulated runtime checks passed for the current stack; no production fixes were needed during runtime validation.

## Impact

Connected plugin scale clients can read firmware and battery values through the existing scale-info contract. Battery `0` and `100` remain valid, unavailable battery remains nullable, and stale sessions cannot update a replacement device. No new inventory field or settings store is introduced; the corresponding API and plugin documentation changes are part of this boundary.

## Contributor Responsibility

AI-assisted development is allowed. The submitter remains responsible for the submitted work.

- [ ] I have reviewed and understand all changes in this PR and take responsibility for their correctness, security, behavior, licensing, and provenance, including any AI-assisted or AI-generated work.

Issue #833, opened 2026-09-12
Reserve a second scale for weighing the dose


SongPaul:
## What

Baristas weigh the dose on one scale and the shot on another. Today the gateway holds exactly one scale, so the second one cannot be connected at all.

`ScaleController` has a single slot:

```dart
class ScaleController {
  Scale? _scale;
```

and everything that reads weight hangs off it — `connectionGeneration`, `currentWeightSnapshot`, `weightSnapshot`, and through those `ShotSequencer` (tare at shot start, stop-at-weight, the shot record). The REST and WebSocket surface is singular to match: `PUT /api/v1/scale/tare`, `ws/v1/scale/snapshot`, with no way to name a scale.

## Why open this separately

Concurrent multi-scale streams are already logged as a P2 gap from the Bengle work (`doc/plans/archive/bengle-integrated-scale/2026-05-05-bengle-integrated-scale.md`, D1), and I understand role-keyed multi-scale is not currently planned. This issue is narrower than that: one extra scale, reserved by id for dosing, with the brewing path left as it is.

## Proposed shape

- A `DosingScaleController` with its own slot, connection subscription and snapshot stream. `ScaleController` is not modified.
- A `dosingScaleId` setting beside the existing `preferredScaleId`. The brewing path's selection gains exactly one exclusion — "every scale except the one reserved for dosing".
- New endpoints beside the existing ones rather than a role parameter on them, so no request an existing skin already sends changes:
  - `ws/v1/scale/dosing/snapshot` (new), `ws/v1/scale/snapshot` unchanged
  - `PUT /api/v1/scale/dosing/tare` (new), `PUT /api/v1/scale/tare` unchanged
- Unavailable while a Bengle is connected, since that path deliberately skips external scale discovery.
- No part of a shot reads it. A dose is weighed before extraction, so nothing during extraction should be able to be broken by it.

Two scales connected at once is verified on hardware — BLE adapter contention is not the open question here.

I have this implemented and tested and would like to offer it as a PR. If role-keyed multi-scale lands instead, this should come out rather than sit beside it.

SongPaul:
Opened #835 in Ideas to discuss the design choice behind this — a second scale reserved by id with the brewing path untouched, versus a proper role-keyed multi-scale model. The implementation is up as #834.

tadelv:
I agree with the problem this issue identifies, but I would like to broaden the solution before we merge the current dosing-specific shape.

## Architectural decision

[...]

Issue #716, opened 2026-08-27
feat(scale): add a manual USB-powered override for Skale


MaxRink:
#### Problem

No reviewed Atomax SDK exposes a reliable USB/external-power or charging signal.
Decaid therefore cannot autodiscover that a Skale is powered by USB, and battery
polling can display misleading information in that configuration.

#### Proposed scope

- Add a Skale-specific **Powered by USB** setting, default off.
- When enabled, suppress battery reads, clear any previous percentage, and
  report `usb` as the connected device's power source.
- When disabled during a connection, resume battery reading and refresh policy.

#### Acceptance criteria

- USB power is never claimed automatically.
- The override persists and survives settings export/import.
- UI and REST/WebSocket metadata distinguish manual USB mode from a
  device-reported battery percentage.

#### Evidence

No reviewed Atomax SDK exposes external-power or charging state:

- [Complete public iOS `SKSkale` interface](https://github.com/atomaxinc/SkaleKit/blob/624c07436a577706dcc26cb2d62ca76ad18f7da9/Distribution/SkaleKit.xcframework/ios-arm64/SkaleKit.framework/Headers/SKSkale.h#L56-L86)
- [Complete Android documented operation surface](https://github.com/atomaxinc/SkaleKitAndroid/blob/faae8a96879f45c935712c11f25674e4d5dc9e82/README.md#L41-L115)
- [Flutter public weight/button/battery/display interface](https://github.com/atomaxinc/SkaleKit-flutter/blob/7df4a9d9bd975367c3ac27502323225ac61b116d/lib/src/skale_kit.dart#L128-L310)
- [macOS operation and delegate surface](https://github.com/atomaxinc/SkaleKit-macOS/blob/0f20dc545a6ca5cd2b4821ac849b0443a3cb9798/README.md#L114-L164)
- [de1app's historical USB assumption](https://github.com/decentespresso/de1app/blob/24851f5b1b0faf77b92fdd837b2932a8d8e26e15/de1plus/bluetooth.tcl#L2645-L2652)

tadelv:
Thanks for the suggestion @MaxRink — we'd gratefully accept contributions on this one. If you'd like to implement it, a PR would be very welcome. Adding `ready-for-human` so it lands in the right triage queue.

PR #846, opened 2026-09-13
feat(plugins): add single-device Skale JS driver

MaxRink:
## Summary

Add an opt-in JavaScript Skale driver for one connected physical scale, using the host's existing one-binding limit. JavaScript owns 4/5/9-byte weight parsing, timer/display/tare commands, nullable battery and firmware metadata, manual USB-power settings, and session-local timers and callbacks.

Physical button actions are staged separately with held #845/#853. This runtime-validation PR does not modify the machine-control API or issue machine-control requests, including when an older stored square-action setting is enabled.

### Delivery stages and dependencies

The driver consumes connected metadata from #844. It has no guarded-action API prerequisite. The host quota increase #843 and primary/auxiliary multi-instance validation remain in #859. The generic native settings entry #849 is separate; the plugin-owned settings endpoint works without it.

## Linked Issue

Related #713, #714, #716, #852 and runtime-v2 tracker #890. Physical button behavior from #715/#853 remains separate and held for human review. The concurrent consumer follows #833/#851 in #859. This PR remains a draft.

## Verification

- Focused Skale suite: 15 passed, including protocol, reconnect, metadata, USB settings and absence of machine actions from physical buttons.
- Full `flutter test --no-pub --concurrency=2`: 4,207 passed, 1 skipped, 0 failed.
- Full Dart analysis found one unused test import; removed it and the final targeted analyzer reports no issues. Formatting and `git diff --check` pass.
- Simulated runtime smoke was attempted with `sb-dev`; macOS build is blocked by unavailable `xcodebuild` (Xcode is not installed). No physical hardware validation.

## Impact

Single-device protocol behavior and per-device USB settings remain supported. Physical button actions are deliberately deferred; tare, timer and display commands remain available through the host Scale API. No native/plugin identity migration or new persistence authority is introduced.

## Contributor Responsibility


tadelv:
0.9.0 priority: **P1 runtime consumer/validation**, tracked in #890.

The single-device Skale JS driver is useful proof that real device protocol behavior can live on the generic runtime, but it should not pull safety-sensitive guarded machine-action work into the P0 release path. If #845/#853 is the only blocker, prefer staging/splitting the button behavior so the driver-runtime validation can remain independently reviewable.

PR #847, opened 2026-09-13
fix(steam): select the declared milk-temperature sensor

MaxRink:
## Summary

Fix SteamSequencer sensor selection so a declared temperature-capable source is selected by its channel contract and remains identity-fenced through lifecycle changes.

- Prefer the declared DecentEspresso/Bengle Milk Probe temperature contract over unrelated object sensors.
- Reselect during an open steam record when the selected source is removed or replaced.
- Preserve the selected probe across registration order, attach/reconnect, unrelated payloads, and E64-like object sensors.

## Linked Issue

Fixes #831

## Verification

- `test/controllers/steam_sequencer_test.dart`: 17 passed.
- `test/integration/steam_sequencer_integration_test.dart`: 1 passed.
- `flutter analyze`: no issues.
- Parent Linux Flutter suite: 4,158 successful, 1 skipped, 0 failed.
- No hardware claim is made.

## Impact

Steam milk-temperature snapshots no longer depend on registry order or an unsuitable sensor's payload shape. This is an isolated sensor-selection bug fix and is a prerequisite for coexistence with the read-only E64 sensor plugin.

## Contributor Responsibility

AI-assisted development is allowed. The submitter remains responsible for the submitted work.

- [x] I have reviewed and understand all changes in this PR and take responsibility for their correctness, security, behavior, licensing, and provenance, including any AI-assisted or AI-generated work.

Issue #747, opened 2026-08-31
feat(devices): add MOTTO80 BLE grinder support


Sofronio:
## Request

Add support for the **MOTTO80** (Bookoo) BLE grinder as a new device kind. The grinder advertises as "BOOKOO MT80" with a custom service (4d543830-0001-4b80-8f00-424f4f4b4f4f) and ~200ms JSON status broadcasts; a full implementation is ready on branch `feature/motto80-grinder` of a fork.

## Status

- New `DeviceType.grinder` with a typed `Grinder` abstraction parallel to `Scale` (GrinderSnapshot DTO, start/stop, presets, grind sections, geneSetting adjustments)
- `BookooGrinder` protocol driver: A5 01 framed JSON with seq, long-frame reassembly, handshake + section/preset queries, generic `grindSetting` (mapped to the MOTTO80 bladeGap wire key)
- Auto-connect via `preferredGrinderId`; devices/debug/settings UI; `MockGrinder` simulated device
- REST/WS API for skins: GET /api/v1/grinder, PUT /api/v1/grinder/settings, PUT /api/v1/grinder/<command>, WS /ws/v1/grinder/snapshot
- Hardware-verified: discovery, connect, feedingRpm/grindRpm changes reach the physical grinder and its broadcasts confirm the new values

tadelv:
0.9.0 direction: #889 now defines the connected Grinder domain/API on top of the generic driver runtime from #749.

Please treat this issue as the **MOTTO80/Bookoo reference-driver consumer**, not as the authority for the generic Grinder abstraction. The existing protocol/hardware work is valuable and should be preserved (framing/reassembly, handshake, matching, fixtures, hardware findings), but the runtime implementation should be reworked onto the transport-independent contract in #889 and preferably delivered as a JavaScript/plugin driver rather than adding a new native MOTTO80 protocol implementation to core.

Two boundaries in #889 are particularly important here:

- existing `/api/v1/grinders` and workflow `grinderId` remain persisted equipment/history records; connected grinder identity is the runtime `deviceId`;
- MOTTO80-specific controls such as feeder RPM, brightness, standby, cup detection, fast-clean, reboot, grind sections and protocol logs do not become mandatory generic Grinder API concepts.

I would sequence this after the runtime-v2 foundation and #889 contract checkpoint. The concrete driver can then prove the BLE side of the generic Grinder domain.

PR #848, opened 2026-09-13
feat(plugins): add read-only multi-instance E64 sensor

MaxRink:
## Summary

This draft is read-only runtime/transport reference work pending the connected Grinder contract in #889. Its Sensor registration is not the permanent grinder-facing API. After that domain lands, E64 should publish via `type: grinder` using the shared transport-independent controller/API; E64 configuration, machine information and logs remain plugin-specific. Read-only Grinder telemetry is sufficient; motor/write behavior is not required.

Add an opt-in E64 sensor plugin that reads grinder telemetry through the existing host-owned WebSocket transport, with independent same-model instances and no grinder writes.

- Keep one registration, socket, request counter, pending map, timer, and connection epoch per configured instance.
- Expose only `readState`, `readConfig`, `readMachineInfo`, and `readLogMessages`, with exact request/ref correlation and complete typed object publications.
- Validate configuration and secure tokens before opening sockets, isolate one instance's failure, and preserve existing primary/auxiliary scale connections.

## Linked Issue

Related #854. Depends on the encoded opaque-ID API fix #858. The merged #847 Bengle fix and separate #747/#748 MOTTO80 work are outside this plugin.

The new discussion issue is awaiting maintainer triage. This PR remains a draft while maintainers discuss the proposal; opening the issue does not satisfy the accepted-issue gate or imply approval of the proposed contract.

## Verification

- Simulated two-fixture E64 runtime passed REST info and all four reads for both instances, independent WebSocket snapshots, and left fixture B responsive after fixture A disconnected.

- Simulated two-fixture E64 runtime passed REST info and all four reads for both instances, independent WebSocket snapshots, and left fixture B responsive after fixture A disconnected.

## Impact

This adds an opt-in telemetry-only sensor path. It sends no motor, calibration, brew-event, or GBS writes, and it does not alter the existing scale API or inventory contract. The prerequisite sensor-selection fix is included in this branch and must land first through #858.

## Contributor Responsibility


tadelv:
0.9.0 direction: #889 is now the contract gate for connected grinder semantics.

This draft is useful evidence that the existing plugin network transport can support independent E64 instances, correlation, reconnect isolation and typed telemetry. I am happy for that work to remain the read-only proof, but I do not want the Sensor registration/payload here to become the permanent grinder-facing API.

After the generic Grinder domain lands, the intended follow-up is to adapt E64 onto `type: grinder` so it consumes the same transport-independent Grinder controller/API as a BLE-backed grinder. E64-only config, machine information and logs should remain plugin-specific. No need to add motor/write behavior merely to satisfy that migration; read-only Grinder telemetry is a valid first consumer.

Please treat the current PR as runtime/transport reference work pending #889 rather than as a competing Grinder API design.

PR #858, opened 2026-09-13
fix(api): share opaque path decoding for sensor IDs

### LINKED ISSUE #857: fix(api): use a shared decode-once boundary for opaque resource IDs
# Bug and shared convention: opaque resource IDs in API paths

## Reproduction

A plugin sensor returned by `GET /api/v1/sensors` has an ID such as `plugin:e64ws.reaplugin:e64ws:e64-a`. Sending `POST /api/v1/sensors/plugin%3Ae64ws.reaplugin%3Ae64ws%3Ae64-a/execute` with `{"commandId":"readState"}` returns 404 even though that sensor is registered. The same missing decode affects sensor info and snapshot lookup. This was reproduced with two E64 plugin instances in #848.

## Reworked design

Apply this convention only where the existing identity contract is opaque or externally supplied, beginning with sensor IDs. It does not make every resource ID arbitrary or externally assignable. Clients percent-encode each ID once as one URL path component. At the route boundary, decode the captured component exactly once, using a shared `decodeOpaquePathComponent` helper backed by `Uri.decodeComponent`. Pass the decoded ID to resource lookup and retain that same ID for WebSocket subscriptions and reconnect/rebind matching. Do not decode it again downstream.

Adopt the helper in sensor info, execution, and snapshot routes. The separate per-scale snapshot and tare implementation specified in #833 will reuse it for opaque device IDs. Existing singular `/scale/...` endpoints remain unchanged. Do not mechanically retrofit other `{id}` routes; any later adoption must first establish that the resource contract actually permits opaque identities.

This scope follows the [maintainer follow-up on #858](https://github.com/decentespresso/decaid/pull/858#issuecomment-5662191602).

### Compatibility

| Path component received | Application identity / behavior |
| --- | --- |
| Literal path-safe ID, including `plugin:...` | Existing supported literal spelling continues to work |
| `%3A`, `%2F`, encoded Unicode | Decode once to the corresponding character in the ID |
| `%252F` | Literal `%2F`, never `/` |
| `%25` | Literal `%` |
| Raw `%`, `%ZZ`, or trailing `%` | Dart normalizes these spellings before routing; one boundary decode yields the corresponding literal percent-bearing identity |
| `+` | Literal plus, not a space |
| Invalid UTF-8 such as `%FF` | HTTP 400 at the existing Shelf request boundary, before WebSocket upgrade |

The helper must not reinterpret the decoded identity or add another normalization layer. Preserve existing unknown-resource responses and subscription lifecycle behavior.

Baseline checks confirm supported colon-composed plugin IDs already work in literal form. Plugin manifest IDs, driver IDs, and public registration instance IDs are validated as safe components and cannot contain `%`. Directly injected synthetic IDs such as a literal `%2F` demonstrate that old raw-token lookup and decode-once lookup are different; the percent cases above define the opaque boundary semantics, rather than promising compatibility for unsupported injected identities. This change does not widen plugin registration validation or introduce a legacy raw-token fallback.

### Scope boundaries

- Query values from `Uri.queryParameters` are already decoded and must not use this path helper. No query-filter changes are included.
- Bean IDs are host-assigned UUIDs. Creating a percent-bearing bean ID through a crafted backup is outside the supported contract; #860 was closed as not planned. Do not add bean, bean-batch, or shot-filter changes or tests for arbitrary custom bean IDs.
- Existing profile, grinder, shot, steam, and presence-schedule routes are outside this implementation; the shared convention does not redefine their identity contracts.
- Command names, plugin endpoint dispatch tokens, filenames, filesystem paths, and account-proxy path normalization keep their existing contracts.
- Plugin IDs, skin IDs, and KV namespace/key routes with existing raw or persisted token semantics are excluded from this compatibility change. Any conversion needs its own explicit migration design.
- This introduces no grinder-specific API, device metadata payload, or scale-purpose model.

## Acceptance and validation

- Add regression tests before changing behavior, exercising real registered IDs through sensor info, execution, and snapshot routes.
- Cover canonical encoded IDs, supported literal forms, Unicode, slash, plus, literal percent sequences, and the raw-percent compatibility cases above. Use raw-wire requests where URI construction would otherwise hide normalization.
- Verify invalid UTF-8 receives 400 before HTTP handling or WebSocket upgrade proceeds.
- Verify two sensor instances remain independent, reconnecting the same ID rebinds its snapshot stream, and disposing/disconnecting one subscription does not affect its sibling.
- Verify adopted sensor routes decode once without changing query, UUID-resource, proxy, file-path, or stored-token behavior. The separate #833 per-scale tests must cover reserved-character device IDs using the same helper.
- Update REST/WS specifications and `doc/Api.md` in the implementation commit, and record the boundary convention in agent/API guidance.
- Run relevant tests, analysis, the full Flutter suite, and an HTTP/WebSocket runtime smoke using encoded E64 sensor IDs.

## Dependencies and review scope

Draft implementation: #858. Consumer and original reproduction: #848; related plugin discussion: #854.

The auxiliary-scale design accepted in #833 depends on this shared decode-once convention for `PUT /api/v1/scales/{id}/tare` and `ws/v1/scales/{id}/snapshot`. Keep that connection-management and per-scale API work in a separate PR. This issue defines identity handling only; #833 defines primary versus runtime-only auxiliary scale connections.

MaxRink:
## Summary

Fix sensor route lookup for plugin-owned IDs containing URI-reserved characters.
REST manifest, REST command execution, and WebSocket snapshot/rebind paths now
percent-decode the captured sensor ID exactly once before lookup. This restores
the encoded two-instance E64 runtime path that currently returns 404.

No new API route is added. The shared `decodeOpaquePathComponent` helper is the boundary for these sensor routes and is available for reuse by the separate #833 scale API.

## Linked Issue

Fixes #857, including its full compatibility matrix.

Consumer context: #848

Issue #857 now records the reworked decode-once design; this draft implements its sensor-only starting scope. Combined publication validation remains pending.

## Verification

- The encoded two-fixture E64 runtime passed REST info and all four reads for both instances, independent WebSocket snapshots, and left fixture B responsive after fixture A disconnected.

- The encoded two-fixture E64 runtime passed REST info and all four reads for both instances, independent WebSocket snapshots, and left fixture B responsive after fixture A disconnected.

## Impact

Clients encode each sensor ID once when placing it in a REST or WebSocket path component; the host decodes that component exactly once. Supported literal colon-composed plugin IDs continue working without client changes, and `%252F` identifies a literal `%2F`, never `/`.

Public plugin, driver, and instance-ID validation excludes percent-bearing components, so synthetic raw-percent identities do not justify a fallback, migration, or widened registration contract. UUID resources and query handlers remain outside this path convention; no #860 query-filter change is included. Unknown IDs retain existing errors, and invalid UTF-8 remains HTTP 400 at the HTTP boundary.

## Contributor Responsibility


tadelv:
@MaxRink can you bring this PR branch up to date with origin/main (preferably rebase if possible, no worry about --force)

doc/plans/archive/opaque-sensor-route-boundary/rationale.md
```diff
@@ -0,0 +1,24 @@
+# Opaque external sensor route boundary
+
+Issue #858 fixes external plugin sensor IDs containing reserved characters.
```
tadelv, line 3: This should reference issue **#857**, not #858. #858 is this PR; #857 is the accepted issue defining the decode-once boundary and compatibility contract. Please correct the archived rationale so the retained design history points to the canonical issue.

tadelv (dismissed):
The decode-once implementation itself looks appropriately scoped: it only changes the three sensor route boundaries, keeps the decoded identity through WS lookup/rebind, leaves UUID/query/proxy/file/token contracts alone, and the API/spec/agent guidance is updated consistently.

There is one merge-blocking repository-rule issue: the current PR body uses `Fixes [#857](...)`, but the contribution-policy gate only accepts forms such as `Fixes #857` / `Closes #857` / `Related #857`. As a result, the current head's **Contribution policy** check fails and the analyze/test/build jobs are skipped. This violates the required PR-template / accepted-issue gate in `AGENTS.md` and `CONTRIBUTING.md`, and it also means we do not have CI verification for this head yet.

Please change the linked-issue line to `Fixes #857` (the issue is open and currently carries `ready-for-human`) and trigger a fresh PR check run/event so the real CI gates execute. Once those are green, I don't see another blocking blast-radius concern in this patch.

tadelv (approved):

PR #867, opened 2026-09-14
fix(scale): separate original Decent Scale and HDS behavior

### LINKED ISSUE #839: [Bug Report] **Scale not connecting**
## Description
**Scale not connecting**

check with cottec

## System Info
**System Information:**
- App Version: 0.8.5 (build 2624)
- Commit: a08bc41e
- Branch: HEAD
- Platform: android
- OS Version: treble_arm64_bvS-userdebug 10 QQ3A.200805.001 eng.crossg.20210808.170014 test-keys

## Attachments
Logs and screenshots: https://gist.github.com/reaprime-bot/227aae4b07171c8d6f36481f2e3728ff

---
_Submitted via in-app feedback on 2026-09-13T18:42:59.988439_

tadelv:
## Summary

What changed, and why?

- Negotiate a conservative Decent Scale profile from protocol evidence so original scales do not receive HDS-only commands.
- Remove heartbeat and periodic maintenance writes; retain passive notification watchdog recovery.
- Gate HDS SoftSleep/power-off by capabilities and preserve original-scale firmware command reliability rules.
- Treat base display-off (`0A 00`), HDS SoftSleep (`0A 04`), power-off and BLE disconnect as independent. `ScalePowerMode.displayOff` no longer disconnects a healthy original/unknown/pre-modern-HDS scale; it sends the shared display-off command and keeps weighing. This corrects the policy exposed by field report #874.
- Add profile, lifecycle, disconnect, timeout, stale-evidence, command reliability and controller lifecycle coverage plus BLE/device-management documentation.

## Linked Issue

Fixes #839. Follows up field report #874.

## Verification

- `dart format --output=none --set-exit-if-changed lib test` — 0 changed
- Targeted Decent Scale suites + `de1_state_manager_wake_scan_test` — 66 tests passed
- `flutter test` — 4,233 passed, 1 skipped, 10 failed; all 10 failures are the localhost webserver/webui port-3000 tests and are caused by a running Decaid instance holding port 3000 (reproduced at HEAD with the change stashed)
- Mutation/behaviour checks confirm the `0x22` capability split, fw>=3 SoftSleep gate, failed-SoftSleep -> display-off fallback, failed display-off retaining the link, and no reconnect churn on sleep/wake
- HDS SoftSleep exit ordering: `0A 04 00` is sent before LED/status on both the same-connection and reconnect wake paths, and a failed or superseded exit stays logically asleep and retries on the next wake. Four ordering regressions cover failed entry, failed exit retry, obligation surviving a later sleep, and reconnect-before-confirmation.
- HDS SoftSleep exit resilience: a single `wakeDisplay()` retries the exit once (200ms apart) while the display generation is current; on exhaustion it sends no LED/status, does not confirm or disconnect, keeps the exit obligation, and fails the wake instead of silently completing. A later same-connection wake is not misread as a reconnect, and a write completing after a superseding sleep cannot re-enter confirmation. Six lifecycle regressions cover within-wake retry, exhaustion, supersession during retry and during a pending write, no false reconnect, and obligation survival.
- #874 field finding: a real original full-height scale (FW 1.1) was identified correctly and streamed ~30 minutes, then the old policy logged `Decent scale: disconnecting for sleep (SoftSleep unavailable)` and dropped the link. After this revision the same path sends `0A 00` and keeps the connection.
- Real HDS hardware pass (2026-09-15, Android tablet, BLE id `50:78:7D:1F:AE:E1`, model `hds`, firmware `3.1.14-custom`):
  - identified as `halfDecentScale` with `softSleep` from a valid `0x22` response and a decoded firmware 3.1.14;
  - Sleep used the HDS SoftSleep path and retained the BLE connection - log `Decent scale: entering HDS SoftSleep`, no `disconnecting for sleep`, no reconnect;
  - the connection served 4,802 notifications over 484s across the sleep/wake window;
  - Wake restored the display on the same physical connection (a status response followed the SoftSleep line, no new `on connect`);
  - explicit Disconnect powered the scale off correctly (`sending power off`, `powerOff=true`).
- Residual gap: the shared `0A 00` display-off path for unknown/original/pre-modern-HDS scales (the #874 correction) was not exercised on hardware because no such scale was present; it is covered by unit tests. Confirm on the #874 original-scale device before closing.

## Impact

- `ScalePowerMode.displayOff` keeps the BLE connection for unknown/original/pre-modern-HDS Decent Scales and sends the shared `0A 00` display-off command. There is no intentional disconnect merely because SoftSleep is unavailable.
- Proven HDS SoftSleep is unchanged; a failed SoftSleep write falls back to `0A 00` instead of disconnecting. A failed display-off write keeps the healthy link; only the transport watchdog tears down a dead one.
- Wake restores the same connection (`0A 01`, or `0A 04 00` then `0A 01` after SoftSleep); capabilities are not renegotiated merely because the display was toggled.
- `DecentScale` no longer implements `DisconnectToSleepScale`; `De1StateManager` no longer marks it sleeping in `displayOff` mode.
- No database migration or REST/WebSocket API/spec change.
- Original-scale firmware marker mapping is derived from public `pydecentscale` behavior and still needs hardware confirmation.
- Remaining Android GATT/connectivity work is tracked in #871, not here.

## Contributor Responsibility


lib/src/models/device/impl/decent_scale/profile.dart
```diff
@@ -0,0 +1,282 @@
...
+        : null,
+  );
+}
+
+DecentVoltageFrame? parseDecentVoltageFrame(List<int> data) {
+  if (data.length != 7 || !_hasHeader(data, 0x22)) return null;
+  var raw = (data[2] << 8) | data[3];
+  if ((raw & 0x8000) != 0) raw -= 0x10000;
+  return DecentVoltageFrame(voltage: raw / 10);
+}
```
ODevStudio, line 283: I’d validate the XOR before allowing these packets to count as capability evidence. Right now a malformed but correctly shaped `0x22` can promote the device to HDS and unlock SoftSleep/power-off. The same applies to timestamped weight evidence. This also makes `timestampMillis` match its name.

```suggestion
DecentWeightFrame? parseDecentWeightFrame(List<int> data) {
  if ((data.length != 7 && data.length != 10) ||
      !_hasHeader(data, 0xCE, alternateOpcode: 0xCA) ||
      !_hasValidXorChecksum(data)) {
    return null;
  }
  var raw = (data[2] << 8) | data[3];
  if ((raw & 0x8000) != 0) raw -= 0x10000;
  final timestamped = data.length == 10;
  return DecentWeightFrame(
    weight: raw / 10,
    timestamped: timestamped,
    timestampMillis: timestamped
        ? (data[4] * 600 + data[5] * 10 + data[6]) * 100
        : null,
  );
}

DecentVoltageFrame? parseDecentVoltageFrame(List<int> data) {
  if (data.length != 7 ||
      !_hasHeader(data, 0x22) ||
      !_hasValidXorChecksum(data)) {
    return null;
  }
  var raw = (data[2] << 8) | data[3];
  if ((raw & 0x8000) != 0) raw -= 0x10000;
  return DecentVoltageFrame(voltage: raw / 10);
}

bool _hasValidXorChecksum(List<int> data) {
  final checksum = data
      .take(data.length - 1)
      .fold<int>(0, (value, byte) => value ^ byte);
  return checksum == data.last;
}
```

lib/src/models/device/impl/decent_scale/scale.dart
```diff
@@ -340,50 +508,29 @@ class DecentScale implements Scale, TransportHandoffScale {
 
   @override
   Future<void> tare() async {
-    await _writeRequiredCommand([0x0F, 0x00, 0x00, 0x00, 0x01]);
+    await _writeRequiredCommand([0x0F, 0x00, 0x00, 0x00, 0x00]);
   }
```
ODevStudio, line 512: The canonical Decent implementations increment a tare command counter. Since `_writeRequiredCommand()` duplicates the same byte sequence for the unreliable profile, incrementing here once per logical tare preserves that behavior while still sending an identical retry.

```suggestion
  int _tareCounter = 0;

  @override
  Future<void> tare() async {
    _tareCounter = (_tareCounter + 1) & 0xFF;
    await _writeRequiredCommand([
      0x0F,
      _tareCounter,
      0x00,
      0x00,
      0x00,
    ]);
  }
```

lib/src/models/device/impl/decent_scale/scale.dart
```diff
@@ -340,50 +508,29 @@ class DecentScale implements Scale, TransportHandoffScale {
...
 
   Future<void> _sendOledOff() async {
-    await _writeCommand([0x0A, 0x04, 0x01, 0x00, 0x01]);
-    await Future.delayed(Duration(milliseconds: 100));
-    await _writeCommand([0x0A, 0x00, 0x01, 0x00, 0x01]);
+    await _writeNonEssentialCommand([0x0A, 0x04, 0x01, 0x00, 0x00]);
+    await Future.delayed(const Duration(milliseconds: 100));
+    await _writeNonEssentialCommand([0x0A, 0x00, 0x00, 0x00, 0x00]);
   }
```
ODevStudio: It would help if this reported whether the actual SoftSleep command was accepted. `_writeNonEssentialCommand()` intentionally swallows transport errors, so the caller otherwise cannot distinguish successful sleep from a timed-out write on a still-connected GATT link.

```suggestion
  Future<bool> _sendOledOff() async {
    final softSleepSent = await _writeNonEssentialCommand([
      0x0A,
      0x04,
      0x01,
      0x00,
      0x00,
    ]);
    if (!softSleepSent) return false;
    await Future.delayed(const Duration(milliseconds: 100));
    await _writeNonEssentialCommand([0x0A, 0x00, 0x00, 0x00, 0x00]);
    return true;
  }
```

lib/src/models/device/impl/decent_scale/scale.dart
```diff
@@ -392,21 +539,30 @@ class DecentScale implements Scale, TransportHandoffScale {
...
     _notificationWatchdog?.cancel();
-    _log.info('Putting Decent Scale display to sleep');
-    await _sendOledOff();
+    if (_profile.capabilities.supportsSoftSleep) {
+      _log.info('Putting Decent Scale display to sleep');
+      await _sendOledOff();
+      if (await _device.getConnectionState() != ConnectionState.connected) {
+        await _disconnect(powerOff: false);
+      }
+      return;
```
ODevStudio, line 565: Then the sleep path can explicitly fall back to disconnect when SoftSleep itself fails, even if the native BLE state still says `connected`. That avoids leaving `_isSleeping` true with the watchdog disabled while the scale never actually entered SoftSleep.

```suggestion
    if (_profile.capabilities.supportsSoftSleep) {
      _log.info('Putting Decent Scale display to sleep');
      final softSleepSent = await _sendOledOff();
      if (!softSleepSent ||
          await _device.getConnectionState() != ConnectionState.connected) {
        await _disconnect(powerOff: false);
      }
      return;
    }
```

ODevStudio (commented):
Thanks for the thorough work on this — the capability-based split between original Decent Scale and HDS looks like the right direction, and the lifecycle/reliability coverage is useful.

I went through the protocol and sleep/wake paths and found a couple of things I’d address before merging, plus two smaller protocol details. I’ve left concrete suggestions inline so the intended changes are easy to apply.

Main points:
- validate protocol integrity before using `0x22`/weight frames as capability evidence;
- fall back to a clean disconnect when HDS SoftSleep itself fails or times out;
- preserve the Decent tare command counter while still duplicating the exact same command for affected v1.0 firmware;
- make `timestampMillis` actually contain milliseconds.

Overall I think the architecture is solid; these look like fairly contained fixes around the new negotiation/reliability paths. I’d also still keep the planned original-scale/HDS hardware sleep-wake validation before considering #839 fully closed.

lib/src/models/device/impl/decent_scale/scale.dart
```diff
@@ -417,24 +604,70 @@ class DecentScale implements Scale, TransportHandoffScale {
   }
 
   Future<void> _runWakeDisplay() async {
+    final wasAsleep = _sleepMode != _DecentScaleSleepMode.awake;
```
ODevStudio: Blocking lifecycle race: the state captured here belongs to the first wake operation, but `wakeDisplay()` coalesces every later wake onto that same `_displayOperation` while still incrementing `_displayGeneration`. If another wake arrives before this operation unwinds — e.g. `wake -> sleep -> wake`, or a second wake while SoftSleep exit / channel confirmation is pending — this operation later sees a generation mismatch and returns. The newer caller is waiting on the same stale future, so no fresh wake operation runs and the final desired state can remain dark.

The previous regression test for **“latest wake runs after a superseded wake probe”** was removed in this revision, which leaves this case uncovered.

I’d preserve “latest intent wins”: when an in-flight generation is superseded and `_desiredDisplaySleeping` is now false, either loop again with freshly captured sleep/connection state, or chain a new wake operation after the stale one completes. Please restore a deterministic `wake -> sleep -> wake` regression around a blocked write/probe.

ODevStudio (changes requested):
Re-review on the latest head: this is substantially stronger. The `timestampMillis` issue is fixed, the sleep/display-off split now matches the #874 field evidence, the SoftSleep exit obligation/retry handling is much more explicit, and CI is green. I’ve resolved my two old SoftSleep/disconnect threads because that recommendation has been superseded by the new display-off policy.

I do see one lifecycle regression that I think should be fixed before merge: a newer wake request can be lost while an older `wakeDisplay()` operation is still in flight. I left the exact sequence inline. This is particularly worth fixing because the earlier regression test for “latest wake runs after a superseded wake probe” was removed in this revision.

Two earlier protocol comments are also still open: capability evidence is still accepted without validating the checksummed `0x22` / timestamped frame path, and tare still sends a constant counter byte instead of incrementing once per logical tare. I’d keep those threads open until they’re either addressed or explicitly justified with protocol/hardware evidence.

The hardware validation and the added lifecycle coverage are good progress; after the wake-coalescing race is covered again, I’d be comfortable doing another pass.

lib/src/models/device/impl/decent_scale/profile.dart
ODevStudio: Re-review note: the `timestampMillis` part is fixed now. The integrity/capability part is still open. I’d narrow the validation to the frames that actually widen capabilities: require a valid XOR for the `0x22` response and for the 10-byte timestamped-weight evidence, while keeping whatever legacy tolerance is needed for ordinary 7-byte/status traffic. That avoids changing normal weight compatibility but prevents malformed evidence from enabling HDS/power-off capabilities.

lib/src/models/device/impl/decent_scale/scale.dart
ODevStudio: Re-review note: this is still current on the new head — tare still sends a constant counter byte. Unless there is hardware/protocol evidence that the counter is deliberately unnecessary here, I’d keep the canonical behavior: increment once per logical tare, then let the v1.0 reliability workaround duplicate that exact same frame.

lib/src/models/device/impl/decent_scale/profile.dart
ODevStudio: Fixed on the branch. `0x22` responses now require a valid XOR before they can promote HDS capabilities, and 10-byte timestamped weight evidence is checksum-gated while ordinary 7-byte weight/status compatibility remains tolerant. `_recordWeightFrame()` now uses the validated parser result, and corrupt-evidence regressions plus corrected HDS fixtures cover the path. CI format/analyze/full Flutter tests are green.

lib/src/models/device/impl/decent_scale/scale.dart
ODevStudio: Fixed. Tare now advances a per-connection sequence byte once per logical tare. The v1.0 reliability workaround still duplicates the exact same generated frame 50 ms later, so the retry keeps the same sequence value. Regression coverage verifies the first logical tare uses 0, the next uses 1, and each duplicate is byte-identical. CI is green.

lib/src/models/device/impl/decent_scale/scale.dart
ODevStudio: Fixed. `_runWakeDisplay()` now captures sleep/reconnect state per loop generation and continues when an older generation is superseded but the latest desired state is awake. Added a deterministic blocked-SoftSleep-exit `wake -> sleep -> wake` regression that verifies the coalesced latest wake actually restores LED/status on the same connection. Full CI is green.

ODevStudio (approved):
Re-reviewed after the follow-up fixes. The three outstanding items are addressed: capability-widening evidence is checksum-gated without tightening legacy 7-byte traffic, tare sequencing now advances per logical tare while v1.0 retries remain byte-identical, and the wake lifecycle preserves the latest intent across an in-flight `wake -> sleep -> wake` supersession. Regression coverage was added for each path. PR checks are green (format, analyze, full Flutter tests, and Linux build smoke).

PR #882, opened 2026-09-15
fix(scale): harden HDS sleeping reconnect follow-up

ODevStudio:
## Summary

Follow-up to merged PR #867, based on review against the current `decentespresso/openscale` HDS firmware behavior.

This draft addresses two lifecycle gaps:

1. **Display-off reconnect state** — OpenScale intentionally restores the display after a BLE disconnect unless it is in true SoftSleep. A Decaid reconnect while `displayOff` was therefore able to leave the physical display on while Decaid still considered the scale asleep. Sleeping reconnects now re-assert the shared `0A 00` display-off command.
2. **SoftSleep wake confirmation** — OpenScale queues the `0A 04 00` wake from the BLE callback and applies it later on the main loop. A successful GATT write therefore proves that wake was queued, not that it was applied. Decaid now keeps the SoftSleep-exit obligation until a post-wake status frame is observed; the existing bounded exit retry is reused when confirmation does not arrive.

The sleeping lifecycle also retains the last positively negotiated power-off/SoftSleep capability evidence across the reconnect boundary. The ordinary per-connection profile still resets to conservative and is re-negotiated on wake, so this does not relax fresh-connection negotiation for normal awake connections.

## OpenScale reference

Current OpenScale behavior used for this follow-up:

- BLE disconnect calls `restoreDisplayAfterBleDisconnect()`, which restores the display unless `b_softSleep` is true.
- BLE SoftSleep enter/exit commands only queue main-loop work.
- `displayOn()` queues a BLE status response, and main-loop pending commands are applied before the queued BLE status response is emitted.

## Tests

Added focused regressions for:

- display-off reconnect re-asserting darkness;
- retained HDS power-off evidence surviving a sleeping reconnect;
- `wakeDisplay()` remaining pending until a post-wake HDS status frame arrives.

## Verification

- Reproduced all three CI failures in the shared display-off reconnect test helper, then asserted exactly one display-off packet rather than no writes.
- Focused disconnect, OpenScale follow-up, and lifecycle suites: 41 tests passed.
- `flutter analyze --no-pub`: no issues found.
- `flutter test --no-pub --concurrency=1`: 4255 passed, 1 skipped on Windows. Used the pinned plugin assets, QuickJS DLL and OpenSSL on PATH; serial execution avoids local port contention between web UI suites.
- Changed test formatting and `git diff --check`: clean.
- No additional hardware pass has been performed for this follow-up. GitHub CI will rerun after commit `539caafd`.

Refs #867.
## Linked Issue

N/A - maintainer follow-up to merged PR #867.

## Impact

Sleeping HDS reconnects reassert display-off, and SoftSleep wake waits for status confirmation. No API/spec, database migration, skin, or plugin contract changes. The CI follow-up only updates the existing reconnect test helper to assert the exact display-off packet instead of expecting no writes.

## Contributor Responsibility


tadelv (approved):
Reviewed adversarially against AGENTS.md, the surrounding Decent Scale lifecycle code/tests, and OpenScale's disconnect behavior. The underlying issue is real: OpenScale can restore the display after BLE disconnect unless soft sleep is active, so retaining the sleep-time capability/profile state and reasserting Display Off on reconnect is appropriate. SoftSleep is correctly not reissued, normal data-channel/profile negotiation resumes on wake, and the added regression coverage exercises the relevant reconnect/wake paths. I did not find a merge-blocking correctness or architecture issue in this PR.

PR #887, opened 2026-09-16
fix(import): correct DYE grinder specs and enjoyment scale

allofmeng:
## Summary

Three defects in the de1app importer, all found while planning DYE2's import story. None are reachable from Decaid's own UI, which is why they went unnoticed: `annotations.enjoyment` has no Flutter renderer, and the grinder spec fields only surface in the DYE2 plugin.

- **`GrinderTdbParser` read key names DYE never writes.** It looked for `setting_type` and `burrs`. DYE writes `is_numeric 1 min … max … n_dec … small_step … big_step … default …` for numeric grinders, or `is_numeric 0 default … big_step 5 values {…}` for preset ones (`DYE.tcl:2538-2541`). `specs['setting_type'] == 'numeric'` was therefore always false, so **every** imported grinder became `preset` — and `EntityExtractor.mergeGrinderSpecs` then overwrote the shot-derived grinder with that wrong type, costing numeric grinders their stepper.
- **Preset grinders lost their settings.** The parser never read `values`, and `mergeGrinderSpecs` did not carry `settingValues` through the merge, so `Grinder.settingValues` was always null.
- **`espresso_enjoyment` was copied across scales.** de1app declares it `Enjoyment (0-100)` (`app_metadata.tcl:284`) and writes the same value to `meta.shot.enjoyment` in `history_v2` (`shot.tcl:230`). Both parsers copied it verbatim into `annotations.enjoyment`, whose only consumer is DYE2's 5-star rater, so every imported shot rated above 5 pinned to full marks and the first click silently rewrote e.g. 87 to 4.

The `grinders.tdb` test fixture used the invented key names too, so the existing tests passed against a format de1app does not produce. Replaced with DYE's real output, including a preset grinder and an empty spec (`DYE.tcl:2547`).

`TclParser.splitList` is new: `_parseBracedValue` returns a preset `values` list as a `List` only when every element is numeric, a `String` for two or three word elements, and a `Map` for four or more (the pairs heuristic). Rather than change that shared heuristic and risk other callers, `GrinderTdbParser` normalises all three shapes, using the tokeniser for correct handling of braced elements containing spaces.

## Linked Issue

N/A — maintainer fix, no prior planning context needed.

## Verification

`flutter analyze` — no issues. `dart format lib test` — clean.

`flutter test test/import/` — **238 tests, all passing**. Tests were written first and confirmed failing against the old implementation:

```
Failing tests:
  test/import/entity_extractor_test.dart: EntityExtractor.mergeGrinderSpecs carries preset setting values onto the shot-derived grinder
  test/import/grinder_tdb_parser_test.dart: GrinderTdbParser maps is_numeric 0 to preset setting type with its values
  test/import/grinder_tdb_parser_test.dart: GrinderTdbParser maps is_numeric 1 to numeric setting type
  test/import/shot_v2_json_parser_test.dart: ShotV2JsonParser shot annotations rescales enjoyment from the de1app 0-100 range to 0-5
  ... and 2 more
```

Full `flutter test` run: 4251 tests, 10 failures, all pre-existing and environmental — `SocketException: Failed to create server socket (OS Error: Address already in use, errno = 48)` on ports 3000/4001 in `api_docs_server_port_test.dart` and `webui_token_injection_test.dart`, caused by a Decaid instance running on the same machine. Confirmed identical on a stashed clean tree; this change touches no networking.

Not verified against a real tablet's `de1plus` folder — fixture-based only.

## Impact

- **User-visible:** imported grinders keep their numeric/preset type and preset values. Imported enjoyment ratings land on the 0-5 scale DYE2 renders.
- **Re-import:** users who already imported carry the wrong values. `De1appImporter` dedupes shots on `de1app-$clock` and grinders on lowercased model, so re-running import skips them rather than correcting them. Repairing existing data would need a migration, which is not in this PR.
- **Pre-existing, not addressed:** de1app defaults `espresso_enjoyment` to 50, and `shot.tcl` writes it whether or not the user rated the shot, so unrated shots import as 2.5 rather than null. Distinguishing "unrated" from "deliberately rated 50" is not possible from the shot file alone.
- **API/spec:** `assets/api/rest_v1.yml` now documents `ShotAnnotations.enjoyment` as 0-5 with `minimum`/`maximum`. No endpoint changed, and `doc/Api.md` does not enumerate annotation fields, so it needed no update.
- **Migration/security:** none.

## Contributor Responsibility


lib/src/import/parsers/enjoyment_scale.dart
```diff
@@ -0,0 +1,8 @@
+const double _de1appEnjoymentMax = 100.0;
+const double _enjoymentMax = 5.0;
+
+double? rescaleDe1appEnjoyment(double? value) {
+  if (value == null) return null;
+  final scaled = value * _enjoymentMax / _de1appEnjoymentMax;
```
ODevStudio, line 6: Nice to make Decaid's canonical value 0–5 here. One integration boundary needs to move with it: `assets/plugins/visualizer.reaplugin/plugin.js` currently sends `annotations.enjoyment` directly as Visualizer `espresso_enjoyment`, and back-sync maps `espresso_enjoyment` directly back to `annotations.enjoyment`. Visualizer's field is 0–100, so e.g. de1app `80` becomes Decaid `4.0` here and would then be uploaded as Visualizer `4` rather than `80`.

Could we convert at the Visualizer boundary (`* 20` outbound, `/ 20` inbound) and cover both directions with tests? I think this one is worth fixing before merge because it affects the normal Visualizer sync path.

lib/src/import/parsers/grinder_tdb_parser.dart
```diff
@@ -28,4 +30,19 @@ class GrinderTdbParser {
...
+  static List<String>? _stringList(dynamic value) {
+    final values = switch (value) {
+      List() => value.map((v) => v.toString()).toList(),
+      Map() => [
+        for (final entry in value.entries) ...[
+          entry.key.toString(),
+          entry.value.toString(),
+        ],
+      ],
+      String() => TclParser.splitList(value),
```
ODevStudio, line 43: Small edge case, non-blocking from my side: for a single multi-word preset, e.g. Tcl `values {{Extra Fine}}`, `_parseBracedValue` collapses the one braced token to the plain string `Extra Fine`. `splitList` here then sees two ordinary words and returns `['Extra', 'Fine']` instead of `['Extra Fine']`.

Would be good to preserve the single-element grouping or add a regression test for this shape, but I wouldn't hold the PR on this alone.

assets/api/rest_v1.yml
```diff
@@ -6991,6 +6991,11 @@ components:
         enjoyment:
           type: number
           nullable: true
+          minimum: 0
+          maximum: 5
```
ODevStudio, line 6995: Non-blocking contract note: existing de1app imports can still contain the old 0–100 value, and re-import skips already-present shots rather than migrating them. Until those rows are migrated, a GET can therefore return an `enjoyment` above 5 even though the OpenAPI schema now declares `maximum: 5`.

Since migration is explicitly out of scope here, I'd either leave this as a documented follow-up or consider whether hard bounds should wait for that migration. Not a blocker from my side.

ODevStudio (changes requested):
Thanks for tracing this back to DYE's actual writer and fixing the fixture along with the importer. The grinder key/type/value changes look well targeted, and the import tests cover the main paths nicely.

I found one cross-component issue I'd like to see handled before merge: Decaid now makes `annotations.enjoyment` a 0–5 value, but the bundled Visualizer adapter still moves that field to/from Visualizer's 0–100 `espresso_enjoyment` without converting the scale. I left the details inline.

I also left two smaller notes around a single multi-word preset value and the OpenAPI bounds vs. already-imported data. I consider those non-blocking edge/contract follow-ups.

lib/src/import/parsers/enjoyment_scale.dart
allofmeng: Fixed in 63de508f — added a `VISUALIZER_ENJOYMENT_SCALE = 20` constant and applied it at all three sync points (`convertReaToVisualizerFormat` upload, `localShotToVisualizerUpdate` forward patch, `mapRemoteToLocal` back-sync). Covered with 3 new regression tests in `test/plugins/visualizer_plugin_test.dart` (upload, forward-sync, back-sync), plugin version bumped to 1.5.7.

lib/src/import/parsers/grinder_tdb_parser.dart
allofmeng: Agreed this is real, but leaving it as-is for now: fixing it properly means threading brace-awareness for single-element grouping through `TclParser._parseBracedValue`/`splitList`, which is the shared heuristic this PR deliberately avoided touching to limit blast radius on other callers. Filing it as a known limitation rather than a regression test asserting the current (wrong) two-word split, since that would just codify the bug. Happy to pick it up as a follow-up PR if you'd rather have it tracked as an issue.

assets/api/rest_v1.yml
allofmeng: Confirmed `rest_v1.yml` isn't enforced at runtime here — `webserver_service.dart` only serves it as the spec doc at a static path, no request/response validation against it. So the `minimum`/`maximum: 0-5` bounds are documentation-only and can't reject a GET returning an un-migrated legacy value. Leaving as a documented follow-up rather than loosening the schema, per your suggestion.

ODevStudio (changes requested):
One Finding
Existing ratings are multiplied twice at [upload](https://github.com/decentespresso/decaid/pull/887/files) and [forward sync](https://github.com/decentespresso/decaid/pull/887/files).
Older imports and Visualizer back-sync leave stored ratings on the 0-100 scale. The new conversion assumes every stored value is already 0-5. I reproduced an existing rating of 80 being sent as 1600 through both upload and full sync.
This makes the deferred migration an outgoing-data problem, not merely the previously discussed schema mismatch. Legacy ratings need normalization before conversion, with an upgrade regression test.

One acknowledged, non-blocking edge remains: a single multiword preset such as {Extra Fine} can become two presets instead of one

Otherwise, the linked update correctly covers all three conversion paths. Twelve in-memory conversion checks passed; GitHub’s analyze/test and Linux smoke checks are green. I did not run Flutter locally.

allofmeng:
Good catch — fixed in 583e99f4.

`toVisualizerEnjoyment()` now treats any stored `annotations.enjoyment` above the canonical 0-5 max as an un-migrated legacy value (still on Visualizer's native 0-100 scale) and sends it through unconverted instead of re-scaling it. Applied at both outbound sites (`convertReaToVisualizerFormat` upload and `localShotToVisualizerUpdate` forward sync) — back-sync was never affected since Visualizer's side is always canonically 0-100.

Added two regression tests reproducing your repro case directly (legacy `80` in `annotations.enjoyment` → `espresso_enjoyment: '80'`/`80`, not `1600`), one for upload and one for forward sync. Plugin version bumped to 1.5.8. `flutter analyze` clean, all 17 plugin tests pass.

ODevStudio:
(> 5) isn't sufficient to distinguish legacy data. de1app's source metadata defines the old field as 0–100 with smallincrement 1, so legacy values 1–5 are valid too. For example, an existing legacy 4 still goes through this branch as canonical and is uploaded as 80. The 80 regression tests cover the common case but not the full old domain. This needs provenance/migration rather than a value-range heuristic, or at minimum a reliable way to identify legacy shots independent of the numeric value.

allofmeng:
You're right, and the fix in 583e99f4 was the wrong shape. `> 5` is a value-range guess, and de1app's `espresso_enjoyment` is 0-100 with `smallincrement 1`, so a legacy 4 is indistinguishable from a canonical 4 at the point of upload. No threshold can separate them.

Reworked in cd7c1ce to repair the data at rest and close the write boundary, so the outbound path never has to guess.

**Schema 6 migration (`_upgradeToSchema6`)** selects rows by provenance, not by value. A row is rescaled when either holds:

- `enjoyment > 5` — no 0-5 writer can produce this, whatever wrote it.
- `id LIKE 'de1app-%' AND updated_at <= created_at` — the row still holds exactly what the importer wrote, which was the raw 0-100 value. The revision contract added in v5 makes this reliable: `PUT` only advances `updated_at` on a real content change, so an untouched import is identifiable independent of its rating.

Both `shot_records.enjoyment` and the `enjoyment` key inside `annotations_json` are rewritten. An annotations blob that fails to parse is left byte-identical rather than replaced.

**`PUT /api/v1/shots/<id>` now rejects an `annotations.enjoyment` outside 0-5.** This also answers your earlier contract note on `rest_v1.yml`: the `minimum`/`maximum` bounds are enforced now, not documentation only, so a 0-100 rating cannot re-enter the store after the migration.

**`toVisualizerEnjoyment` converts unconditionally** (`* 20` out, `/ 20` in) and clamps to 0-5 first. The clamp is not provenance detection, it just means no stored value can be multiplied past Visualizer's range. Your two regression tests now assert the clamp instead of the pass-through: a stored 80 uploads as 100, not 1600 and not 80, because after the migration a stored 80 does not exist.

**Residual, documented rather than guessed at.** An imported or back-synced shot that was edited after it was written is genuinely ambiguous at rest, since it may have been re-rated on the 0-5 scale since. Those rows are left alone in either direction rather than rewritten on a heuristic. Written up in `doc/AI_STORAGE_NOTES.md` under "Schema v6 (enjoyment scale repair)".

**Verification.** `flutter analyze` — clean, only the 4 pre-existing missing-asset-directory warnings. `dart format lib test` — clean.

`flutter test test/database/ test/webserver/shots_handler_test.dart test/plugins/visualizer_plugin_test.dart test/import/` — 381 tests, all passing. New coverage:

```
test/database/enjoyment_scale_migration_test.dart  (9 tests)
  rescales an untouched de1app import whose rating is inside 0-5   <- your repro, legacy 4 -> 0.2
  rescales an untouched de1app import whose rating exceeds 5
  rescales any shot holding a rating above the canonical maximum
  leaves a canonical rating on a native shot alone                 <- DYE2-rated 4 stays 4
  leaves an imported shot edited after import alone
  preserves the other annotation fields
  leaves rows without a rating untouched
  keeps an unparseable annotations blob rather than replacing it
  does not rescale a second time when reopened

test/webserver/shots_handler_test.dart  (5 tests)
  PUT with an enjoyment above the canonical max is rejected
  PUT with a negative enjoyment is rejected
  PUT with a non-numeric enjoyment is rejected
  PUT accepts an enjoyment inside the canonical range
  PUT accepts a null enjoyment to clear the rating
```

`test/database/schema_migration_parity_test.dart`, `shot_schema_migration_test.dart` and `database_startup_test.dart` updated for the version bump. v6 adds no columns, so schema parity between a migrated and a fresh database still holds.

Full `flutter test`: 4236 passing, 44 failures, all pre-existing and environmental in this checkout — missing bundled plugin asset directories (`dye2.reaplugin`, `dcamp.reaplugin`, `shot-upload.reaplugin`) and the `api_docs_server_port` address-in-use conflict. Confirmed identical on a stashed clean tree.

Not verified against a real upgraded database from a tablet — fixture-based only.

The single multi-word preset edge (`values {{Extra Fine}}`) is unchanged and still deferred, since fixing it means reshaping `TclParser._parseBracedValue`'s grouping heuristic for all its callers. Happy to take it as a follow-up issue.

assets/plugins/visualizer.reaplugin/plugin.js
```diff
@@ -1009,7 +1025,12 @@ function createPlugin(host) {
 
     set(annotations, "drinkTds", numberField("drink_tds"));
     set(annotations, "drinkEy", numberField("drink_ey"));
-    set(annotations, "enjoyment", numberField("espresso_enjoyment"));
+    const remoteEnjoyment = numberField("espresso_enjoyment");
+    set(
+      annotations,
+      "enjoyment",
+      typeof remoteEnjoyment === "number" ? remoteEnjoyment / VISUALIZER_ENJOYMENT_SCALE : remoteEnjoyment
```
allofmeng: Remote→local enjoyment conversion (`remoteEnjoyment / VISUALIZER_ENJOYMENT_SCALE`) isn't clamped to 0-5, unlike the outbound `toVisualizerEnjoyment` which clamps before scaling. If the visualizer returns `espresso_enjoyment` outside 0-100 (stale/corrupt remote data), the resulting value falls outside 0-5, the local PUT to `/api/v1/shots/<id>` gets rejected by the new `_enjoymentRangeError` check, `cursorBlocked` latches true, and that shot plus every later item in the run gets retried forever on subsequent back-sync ticks.

lib/src/import/parsers/grinder_tdb_parser.dart
```diff
@@ -11,13 +11,15 @@ class GrinderTdbParser {
       final specs = entry.value;
       if (specs is! Map<String, dynamic>) continue;
 
+      final isNumeric = specs['is_numeric']?.toString() != '0';
```
allofmeng: `isNumeric = specs['is_numeric']?.toString() != '0'` defaults to `true` (numeric) whenever the spec map is missing `is_numeric` entirely, not just when it's `'1'`. A DYE grinder entry with other fields set but no `is_numeric` key gets misclassified as numeric, which also forces `settingValues: null` below — dropping any preset values that should have been kept.

lib/src/import/parsers/grinder_tdb_parser.dart
```diff
@@ -28,4 +30,19 @@ class GrinderTdbParser {
 
     return grinders;
   }
+
+  static List<String>? _stringList(dynamic value) {
+    final values = switch (value) {
+      List() => value.map((v) => v.toString()).toList(),
+      Map() => [
```
allofmeng, line 37: `_stringList`'s `Map()` branch (fed by `TclParser._parseBracedValue`'s pairs heuristic) can silently drop preset setting values when the values list has an even count ≥4 with a duplicate label at an even token position. E.g. a preset grinder's `values {Fine Coarse Fine Medium}` tokenises to 4 tokens; `_buildMapFromTokens` in tcl_parser.dart pairs them as {Fine:Coarse, Fine:Medium}, and the duplicate key 'Fine' overwrites the first pair, so `_stringList` returns `['Fine','Medium']` instead of the true 4-value list `['Fine','Coarse','Fine','Medium']`.

lib/src/services/database/database.dart
```diff
@@ -143,6 +149,75 @@ class AppDatabase extends _$AppDatabase {
     );
   }
 
+  /// Repairs `annotations.enjoyment` rows still holding a 0-100 rating.
```
allofmeng: These rationale comments duplicate what's already documented in doc/AI_STORAGE_NOTES.md's Schema v6 section. Per AGENTS.md: "Do not add explanatory comments to new or substantially rewritten code" and "Put rationale, hardware constraints, and debugging history in the matching doc/AI_*_NOTES.md file." Worth trimming to avoid the two copies drifting out of sync when this migration is touched again.

lib/src/import/entity_extractor.dart
```diff
@@ -101,6 +101,7 @@ class EntityExtractor {
         byModel[key] = byModel[key]!.copyWith(
```
allofmeng, line 101: `Grinder.copyWith` uses `??` fallback semantics, so passing `dye.settingValues`/`dye.burrs` here can never clear a previously-set field with an explicit null — currently inert since `fromShots` grinders never set settingValues/burrs, but a latent footgun if that ever changes: merging in a numeric DYE grinder (settingValues=null) would fail to clear a stale preset list instead of overwriting it.

lib/src/import/parsers/tcl_parser.dart
```diff
@@ -4,6 +4,10 @@ class TclParser {
     final _Parser parser = _Parser(lines);
     return parser.parseBlock(endMarker: null);
   }
+
+  static List<String> splitList(String input) {
```
allofmeng, line 8: Minor: this instantiates a throwaway `_Parser(const [])` purely to call `_tokenise`, which doesn't use any instance state (`_lines`/`_pos`). Could be a static/top-level function to avoid the pointless allocation and make the stateless dependency explicit. Not blocking.

ODevStudio:
Post-commit audit: not ready for an unqualified merge approval.

Commits on this PR: 52b8cfdbe61c29fa8fe8d38a30b2c651224a160c implements the proposed 0-10 contract, Visualizer boundary conversion/clamping and native-rating preservation; 3f84eade6e37ae4db1ecc9dd7fd85c7a8a91a89d avoids treating pre-v5 backfilled timestamps as proof of untouched imports. Formatting follow-ups are 8cb0134efceee310fb99b9fad5a1a35e41c6fc87 and current head 288204c4e6fb5ec905eff47bd0886f54b0c1e272. The latter two change no production behavior.

Blocking integration mismatch: current decentespresso/dye2 still writes `stars * 20` and, outside its old 1-5 heuristic, reads `Math.round(enjoyment / 20)` in dye2-plugin/src/utils/shared-components.ts. Executing those source excerpts against this PR's constants reproduces both failures: a four-star edit produces 80, rejected by the 0-10 PUT check; Decaid 8 renders as zero stars. Source: https://github.com/decentespresso/dye2/blob/main/dye2-plugin/src/utils/shared-components.ts#L302-L321 . The embedded packages/dye2-plugin copy is not the current rating implementation.

Correction to my earlier rationale: archived design text/model tests show a 0-10 design, not proof that all deployed writers have used it. The live DYE2 writer uses 0-100. Also, annotations.extras.visualizer does not establish enjoyment's last writer: later local re-ratings can leave those extras intact. I did not add that unsafe classifier.

Historical overlap remains unresolved: old Visualizer 4 or an edited/pre-v5 de1app 4 without reliable provenance stays 4 and is exported as 40 under the new canonical assumption. Preserving a number does not repair its meaning. Databases already opened with the earlier experimental schema-6 code also need explicit recovery consideration; changing the same migration version does not rerun it.

Verification here: 15 isolated JavaScript boundary checks passed; 12 SQLite cases executed the exact selection SQL with the numeric/JSON update modeled in Python. A proposed DYE2 source-helper companion patch (/2 read, *2 write, no threshold heuristic) passed 21 isolated assertions and applies to the retrieved excerpt. It is not committed or deployed: this connection has read-only DYE2 access. It requires a coordinated 0-10 server release (or explicit scale negotiation), a rebuilt plugin, and API/UI testing; do not deploy it independently against old 0-100 servers.

CI update: run https://github.com/decentespresso/decaid/actions/runs/35850124592 passed the Dart formatting gate. At the last check Flutter analyze was in progress, Flutter tests were pending, and Linux build smoke was in progress. Flutter/Dart are not installed locally, so no local full-suite success is asserted. Earlier runs stopped at formatting before analyze/test.

I have not marked the review threads resolved or claimed a clean end-to-end audit. The acknowledged multiword preset edge is not fixed by these scale changes, and the PR description still describes the superseded 0-5 approach.

ODevStudio:
### DYE2 companion patch required for the 0-10 contract

Decaid PR #887 now keeps `annotations.enjoyment` on **0-10**. de1app and Visualizer remain **0-100** boundaries.

Current DYE2 main still assumes `annotations.enjoyment` is 0-100 in:

`dye2-plugin/src/utils/shared-components.ts`

and currently does:

```js
if (n >= 1 && n <= 5 && Number.isInteger(n)) return n;
return Math.round(n / 20);

function starsToEnjoyment(stars) { return stars * 20; }
```

That must change before DYE2 is released against this Decaid API.

#### Required source change

Use the 0-10 Decaid value as the only input/output contract for the helper:

```js
function enjoymentToStars(enjoyment) {
  if (enjoyment == null || enjoyment === '') return 0;

  const n = parseFloat(enjoyment);
  if (isNaN(n)) return 0;

  const canonical = Math.min(Math.max(n, 0), 10);
  return Math.round(canonical / 2);
}

function starsToEnjoyment(stars) {
  const n = Number(stars);
  if (!Number.isFinite(n)) return 0;

  const clampedStars = Math.min(Math.max(Math.round(n), 0), 5);
  return clampedStars * 2;
}
```

Expected mapping:

```text
Decaid enjoyment   DYE2 stars
0                  0
2                  1
4                  2
6                  3
8                  4
10                 5
```

A four-star click must therefore write:

```json
{"annotations":{"enjoyment":8}}
```

not `80`.

#### Remove the old 1-5 value heuristic

Do **not** keep:

```js
if (n >= 1 && n <= 5 && Number.isInteger(n)) return n;
```

Under the 0-10 contract, values 1-5 are valid canonical Decaid ratings, so that heuristic would misrender normal data (for example, canonical `4` should display as 2 stars, not 4).

Historical DYE2 rows written by the old raw-star bug are now intrinsically ambiguous with valid 0-10 data. If DYE2 wants to repair those rows, use a version/provenance-backed migration; do not infer scale from the value alone.

#### Consumers to verify

`edit-shot.ts` and `dashboard.ts` already use the shared `enjoymentToStars` / `starsToEnjoyment` helpers. After changing the helper, search the repo for any direct enjoyment/star conversions and remove stale `* 20`, `/ 20`, or direct star-index writes.

Useful search targets:

```text
enjoymentToStars
starsToEnjoyment
.enjoyment =
* 20
/ 20
```

#### Rebuild generated output

DYE2's `dye2.reaplugin/` directory is generated output; do not edit it manually.

Run:

```sh
cd dye2-plugin
npm run build
npm test
git status --short ../dye2.reaplugin/
```

Commit both the source change and the regenerated `dye2.reaplugin/` output.

#### Minimum regression coverage

Add tests/assertions for:

```text
enjoymentToStars(null) -> 0
enjoymentToStars(0)    -> 0
enjoymentToStars(2)    -> 1
enjoymentToStars(4)    -> 2
enjoymentToStars(8)    -> 4
enjoymentToStars(10)   -> 5

starsToEnjoyment(1) -> 2
starsToEnjoyment(2) -> 4
starsToEnjoyment(3) -> 6
starsToEnjoyment(4) -> 8
starsToEnjoyment(5) -> 10
```

Also cover clamping for values below 0 / above 10 if the helper retains defensive clamping.

#### End-to-end acceptance check

Against Decaid with PR #887:

1. Open a shot with `annotations.enjoyment: 8` -> DYE2 shows 4 stars.
2. Click 4 stars -> DYE2 PUT sends `enjoyment: 8` and receives HTTP 200.
3. GET the shot -> stored value remains `8`.
4. Reload DYE2 -> still 4 stars.
5. Visualizer sync of that shot -> `espresso_enjoyment: 80`.
6. Visualizer back-sync of `80` -> Decaid stores `8`.

This companion patch should be released **together with** the Decaid 0-10 contract. Do not ship only one side of the change.

Decaid-side CI is green on PR #887 run #1191 (format, analyze, tests, and Linux smoke all passed).

allofmeng:
Thanks for both of these, and for pushing the 0-10 contract and the pre-v5 provenance fix onto the branch directly. I checked your `created_at != timestamp` change against the write path and it is right: `DriftStorageService.storeShot` stamps `createdAt`/`updatedAt` with `now()` when the record carries none, and an imported shot's `timestamp` is the historical de1app clock, so a post-v5 import genuinely has `created_at != timestamp`. My original rule counted every v5-backfilled row as an untouched import. Good catch.

Taking your points in order.

**DYE2 companion patch — opened as decentespresso/dye2#8.** This was the blocking one, so I implemented it rather than leaving it as instructions. `enjoymentToStars` halves, `starsToEnjoyment` doubles, both clamped so nothing out of range reaches the 0-10 PUT check. The 1-5 integer pass-through from issue #7 is gone, for the reason you gave: under 0-10 those are ordinary canonical ratings, so it would misread a stored 4 as 4 stars instead of 2. Rows written by that old bug are now indistinguishable from valid data, so I did not add a value-based repair for them either.

`edit-shot.ts` and `dashboard.ts` already route through the helpers, and a search of `dye2-plugin/src` turns up no remaining direct `* 20`, `/ 20`, or star-index write. `npm test` passes, including a new `test/enjoyment-scale.test.mjs` with the mapping table, both clamps, the null/NaN cases, a 0..5 star round-trip, and the four-star regression. `dye2.reaplugin/` is regenerated and committed, version 0.1.12 in `manifest.src.json` and `package.json`. The regenerated diff is 8 lines, confined to the helper.

I have not run it end to end against a Decaid build of this PR, since there is no tablet or running Decaid instance here. Your six-step acceptance list is in that PR description as the check to run before either side ships. I have not merged it, and it says not to merge it alone.

**Your correction about 0-10 being design intent rather than deployed reality is taken.** The storage note said the field "remains on Decaid's pre-existing 0-10 scale", which overstated it given that the live DYE2 writer uses 0-100. Reworded in df08a3f to present 0-10 as the contract this change establishes. Same correction in the PR description.

**`annotations.extras.visualizer` as a last-writer signal — agreed, and neither of us used it.** A later local re-rating leaves those extras in place, so it proves the row was back-synced at some point and nothing about who wrote the rating last. Worth recording that it was considered and rejected, which the note now does.

**Historical overlap.** Agreed that this is not resolved and I do not think it can be from stored data alone. An old Visualizer 4, or an edited or pre-v5 de1app 4, stays 4 and exports as 40. Preserving the number does not repair its meaning, as you say. The alternative is a value-range guess that corrupts genuine ratings, so leaving them is the lesser harm rather than a fix. Documented as a known gap in both the storage note and the PR description.

**Pre-release schema 6 databases.** I worked through this and there is no sound automatic repair. The earlier revision of this branch stamped `user_version = 6` after dividing by 20 onto 0-5; the corrected step divides by 10. Both end states are `user_version = 6` holding in-range numbers, so a v7 step cannot tell them apart, and re-running the repair on an already-repaired row would halve it again. Since schema 6 has never appeared in a release, this reaches only developers and testers who ran the branch, and the recovery is to delete the development database or restore a backup from before that build. Written up under "Recovering a pre-release schema 6 database" in `doc/AI_STORAGE_NOTES.md` and in the PR description, with a note not to renumber a migration that has already shipped.

**PR description.** Updated. It had already been rewritten for 0-10 by the time I read your comment, so the remaining edits were the design-intent correction, the link to dye2#8 in place of "documented in the PR conversation", and the recovery section.

**Multiword preset edge.** Still not fixed and still acknowledged. Fixing it means reshaping `TclParser._parseBracedValue`'s grouping heuristic for every caller, which I would rather do as its own PR than fold into a scale change. Happy to open the issue if you want it tracked.

**Verification.** Latest head is df08a3f, a documentation-only commit. Run [#35866733096](https://github.com/decentespresso/decaid/actions/runs/35866733096) is green: contribution policy, format, analyze, tests, and Linux build smoke all pass. Locally on 288204c, `flutter test test/database/ test/webserver/shots_handler_test.dart test/plugins/visualizer_plugin_test.dart test/import/` is 385 passing.

Still unverified, and I would not claim otherwise: no run against a real tablet's `de1plus` folder, no run against a production upgraded database, and no end-to-end Decaid plus DYE2 check.

allofmeng:
Two documentation gaps found by auditing the repos rather than the code, both of which would have survived the change and re-argued it back out. Fixed in 5e524ad here and in the companion PR.

**`decentespresso/dye2` `docs/AI_DATA_NOTES.md` instructed the opposite of this change.** It stated `annotations.enjoyment` is 0-100, described the ×20 conversion as the contract, and said of the `[1,5]` pass-through: *"Keep that branch; removing it silently re-reads old shots as near-zero ratings."* Its troubleshooting table carried the same rule. dye2#8 removes that branch, so the note would have contradicted the shipped code and pointed the next maintainer straight back into issue #7.

That note now states the 0-10 contract, says where the 0-100 conversion actually happens, which is Decaid's own importer and Visualizer boundaries, and records both removed rules with the symptom each produces if it comes back: near-zero ratings or a 400 on a star click means the scaling returned, every rating reading two stars high means the pass-through did.

**`doc/Skins.md` describes the field to third-party skin authors as "Numeric rating", with no range.** Skins write to the same `PUT /api/v1/shots/<id>` that DYE2 does, so they are exposed to exactly the failure you reproduced: a five-star skin writing `stars * 20` now gets a 400, and one assuming 0-100 renders a stored 8 as nothing. The entry now gives the range, says out-of-range values are rejected, and notes that Decaid already converts at the de1app and Visualizer boundaries so a skin only converts its own display scale.

A sweep of the rest confirms nothing else asserts a scale. `doc/Api.md` does not mention `enjoyment` at all, `doc/Plugins.md` names it only as an example of an edited field, and the only remaining 0-100 references in either repo are the new text describing the external boundaries. Archived planning documents under `doc/plans/archive/` were left as historical record.

**Companion PR status.** decentespresso/dye2#8 is open, mergeable and rebased on current `main`, and now carries the doc fix alongside the helper change and its tests. `npm test` passes all four suites.

**Verification.** Head here is 5e524ad. Run [#35870563464](https://github.com/decentespresso/decaid/actions/runs/35870563464) is green: contribution policy, format, analyze, tests and Linux build smoke.

The six-step end-to-end check is still outstanding and still the right gate before either side ships, since it needs a Decaid build of this branch running against the rebuilt DYE2.

ODevStudio (approved):
Re-reviewed at 5e524adf. I found no remaining blocking correctness issue in the Decaid changes.

The new documentation correctly distinguishes the 0-10 design contract from previously deployed writers, documents the unrecoverable pre-release schema-v6 state and its developer/tester recovery path, and gives skin authors the enforced 0-10 API range. PR checks #1193 are fully green (format, analyze, Flutter tests, Linux/AppImage smoke).

I also checked the linked DYE2 companion PR #8. Its source helper now maps Decaid 0/2/4/6/8/10 to 0/1/2/3/4/5 stars, writes stars back as *2, removes the unsafe 1-5 pass-through, adds regression coverage, and regenerates dye2.reaplugin/plugin.js. That closes the active integration mismatch provided the two changes are released together. DYE2 #8 is still open and has no GitHub Actions run attached, so I am treating it as a release-coupling requirement rather than a Decaid code defect.

Non-blocking cleanup: the PR description's Verification section still names old head 288204c / run #1191; current head is 5e524adf and current green run is #1193. The previously acknowledged Tcl/grinder edge cases remain non-blocking.

PR #888, opened 2026-09-16
Add ATOM HEART Eclair battery reporting

### LINKED ISSUE #838: fix(scale): report ATOM HEART Eclair battery level
## Problem

The current `AtomheartScale` implementation always publishes `batteryLevel: 0` with weight snapshots. Eclair scales report their actual battery level through `B` notifications on the existing configuration characteristic, but Decaid currently subscribes only to the weight/data characteristic. As a result, users always see the Eclair battery at 0%.

The scale display name is also currently `Atomheart Eclair`; the product branding should be `ATOM HEART Eclair`.

## Proposed scope

- identify the scale in the app as `ATOM HEART Eclair`
- subscribe to the Eclair configuration characteristic after weight notifications are confirmed
- validate Eclair battery frames using the protocol header, XOR checksum, and 0-100 range
- cache the battery level and propagate it with subsequent scale snapshots
- clear the cached value for each new connection so stale battery data is not reused
- keep weight support available if configuration notifications cannot be subscribed
- retain compatibility with the current three-byte battery frame and the known legacy five-byte form

The change is isolated to the Eclair scale implementation and its tests. It does not change discovery, reconnect policy, BLE transport behavior, or any other scale implementation.

## Validation completed

- focused Eclair unit and reliability tests
- device matching regression tests
- full Flutter test suite
- Flutter static analysis
- hardware validation on a Samsung S23 with a physical Eclair scale
- battery notifications observed approximately every 10 seconds
- power-cycle test confirmed automatic reconnect, immediate weight recovery, and a fresh battery value: 25% before shutdown and 24% after reconnect

A small tested patch is ready in my fork. I will wait for maintainer acceptance before opening the pull request, in accordance with the contribution policy.

AtomHeart-Lang:
## Summary

- identify the existing native Eclair scale as `ATOM HEART Eclair`
- subscribe to its configuration characteristic and propagate validated battery notifications through normal scale snapshots
- clear cached battery state between connections and preserve weight support when optional configuration notifications are unavailable
- cover the current three-byte and known legacy five-byte battery frames

This PR is intentionally limited to correcting battery reporting in the existing native Eclair driver. We understand that generic plugin-backed device drivers are the preferred long-term direction, and AtomHeart is interested in maintaining an Eclair plugin separately once that workflow is ready.

## Linked Issue

Fixes #838

## Verification

- `flutter analyze`
- focused Eclair and device-matching suite: 58 tests passed
- full `flutter test`: 4265 tests passed, 1 skipped
- `./scripts/fetch_dye2_plugin.sh`
- Samsung S23 hardware testing with a physical Eclair scale:
  - battery notifications received approximately every 10 seconds
  - power-cycle auto-reconnect with immediate weight recovery
  - a fresh battery update changed from 25% before shutdown to 24% after reconnect, confirming that stale cached state was not reused

## Impact

- Eclair battery status changes from an incorrect constant `0%` to unknown until the first valid report, then `0–100%`.
- The user-visible product name becomes `ATOM HEART Eclair`.
- The change is isolated to the existing Eclair implementation. It does not change discovery, reconnect policy, transport APIs, or behavior for other scales.

## Contributor Responsibility

- [x] I have read and accept the [Contributor Responsibility](https://github.com/decentespresso/decaid/blob/main/CONTRIBUTING.md#contributor-responsibility).


tadelv (changes requested):
Two changes required before merge:

1. **Add regression coverage for stale battery state across reconnect.** This PR introduces persistent `_batteryLevel` state and explicitly promises to clear it between connections, but the new tests only exercise a single connection. Please add a test that uses the same `AtomheartScale` instance to: connect, receive a valid battery value, disconnect, reconnect, emit a valid weight frame *before* the next battery notification, and assert that `batteryLevel` is `null` rather than the value from the previous connection. Then confirm a fresh config notification is applied. This is one of the PR's stated behavioral fixes, and the repository requires regression coverage for bug fixes/new behavior.

2. **Move the legacy-frame rationale out of the implementation comment.** `AGENTS.md` explicitly says not to add explanatory comments to new/substantially rewritten code and says hardware/protocol rationale belongs in the matching `doc/AI_*_NOTES.md`. The new `// Legacy Eclair firmware may append two reserved payload bytes.` comment conflicts with that rule. Please remove it from the implementation and capture the 5-byte legacy compatibility rationale in `doc/AI_BLE_NOTES.md` (keeping the parser itself self-explanatory).

Otherwise, the implementation is well scoped: it stays inside the existing Eclair driver, does not change discovery/reconnect/transport APIs, and the battery protocol handling is consistent with the existing Eclair GATT contract. The latest Analyze / test job passes formatting, `flutter analyze`, and the full Flutter test suite; the remaining Linux smoke failure occurs while fetching external plugins before the build step, so I am not treating that as a defect in this PR.

AtomHeart-Lang:
Thanks for the detailed review. Both requested changes are now addressed in `28e44231`:

1. Added regression coverage using the same `AtomheartScale` instance across disconnect and reconnect. The test confirms that a previously reported 75% battery value is cleared, the first weight snapshot after reconnect reports `batteryLevel == null`, and a fresh 64% config notification is then applied to the next weight snapshot.
2. Removed the legacy-frame rationale from the implementation and documented the current 3-byte and legacy 5-byte Eclair battery frame formats in `doc/AI_BLE_NOTES.md`.

Verification completed with the repository's Flutter 3.44.2 toolchain:

- focused Eclair reliability test: 16 passed
- `flutter analyze`: no issues
- full `flutter test`: 4266 passed, 1 skipped

Could you please take another look when convenient?

tadelv (approved):

tadelv:
Thanks @AtomHeart-Lang !

PR #892, opened 2026-09-16
fix: repair AsyncAPI 3.1 spec and API docs viewer

tadelv:
## Summary

The API docs viewer at `assets/api/index.html` rendered nothing. Two independent defects:

1. `assets/api/websocket_v1.yml` failed AsyncAPI parsing. The `ScaleSnapshotById` channel parameter carried a `schema` key, which is not part of the AsyncAPI 3 Channel Parameter Object — channel parameters are implicitly strings and only accept `description`, `enum`, `default`, `examples`, `location`, and `x-*`. Fixes the spec to 3.1.0 and drops the invalid key.
2. The viewer loaded js-yaml from `https://app.unpkg.com/...`, which is the unpkg *browse UI* and returns HTML, not the library — so `jsyaml` was undefined. The path was also the v4 layout (`dist/js-yaml.min.js`); v5 uses `dist/browser/js-yaml.umd.min.js`. Points at the direct unpkg CDN URL and the v5 path, and aligns the AsyncAPI React Component CSS with 3.1.8 JS bundle (was 2.6.4).

## Linked Issue

N/A — repository maintenance (spec validity + docs viewer).

## Verification

- `@asyncapi/parser` on `assets/api/websocket_v1.yml`: before — `parsed ok: false`, diagnostics `Property "schema" is not expected to be here` at `channels/ScaleSnapshotById/parameters/id/schema`; after — `parsed ok: true`, 0 diagnostics.
- `flutter test test/connect_result_schema_test.dart test/unit/models/device/impl/de1/steam_substate_mapping_test.dart` — All tests passed (8).
- Loaded the page against a local `http-server` on port 8111 and confirmed the served HTML references the corrected js-yaml and CSS URLs, and that `websocket_v1.yml` is served intact.
- No Dart changed, so no formatting or analyze impact.

## Impact

- API/spec: `assets/api/websocket_v1.yml` no longer declares a `schema` on the `{id}` channel parameter. Parameters are strings in AsyncAPI 3, so the runtime contract is unchanged — this only makes the published document valid.
- Docs: the WebSocket/AsyncAPI tab of the bundled API docs viewer renders again.
- No user-visible app behavior, migration, or compatibility impact.

## Contributor Responsibility


assets/api/websocket_v1.yml
```diff
@@ -1,4 +1,4 @@
-asyncapi: 3.0.0
+asyncapi: 3.1.0
```
ODevStudio, line 1: Nice catch on the invalid parameter shape. Since this also bumps the authoritative spec from AsyncAPI 3.0.0 → 3.1.0, could we update the active repo guidance that still calls this spec “AsyncAPI 3.0”? In particular `.agents/skills/decent-app/websocket.md` and `doc/AI_API_NOTES.md` both identify `assets/api/websocket_v1.yml` that way. Keeping those in sync will avoid agents/contributors getting conflicting version information.

assets/api/index.html
```diff
@@ -61,10 +61,10 @@
   <script src="https://unpkg.com/react-dom@18/umd/react-dom.production.min.js"></script>
 
   <!-- AsyncAPI React Standalone UMD -->
-  <script src="https://unpkg.com/@asyncapi/react-component@2.6.4/browser/standalone/index.js"></script>
+  <script src="https://unpkg.com/@asyncapi/react-component@3.1.8/browser/standalone/index.js"></script>
```
ODevStudio, line 64: Small scope/docs note: this line is actually a 2.6.4 → 3.1.8 upgrade in the base diff (for both the JS bundle and CSS), while the PR description says the JS was already on 3.1.8 and only the CSS needed aligning. Could we update the PR description to match the actual base state and briefly note why the component upgrade is needed? That’ll make the verification story much easier to follow later.

ODevStudio (changes requested):
Nice focused fix overall — the invalid channel parameter cleanup makes sense, and it’s good to get the bundled API viewer healthy again. I found one consistency item I’d like to see addressed before merge, plus one small scope/description cleanup. Both are inline below. CI is green and I didn’t spot any other functional concerns in the changed code.

ODevStudio (approved):
Thanks for the follow-up. The version consistency point is addressed now: `.agents/skills/decent-app/websocket.md` and `doc/AI_API_NOTES.md` both match `assets/api/websocket_v1.yml` at AsyncAPI 3.1.0, and the latest CI run is green (analyze/test + Linux smoke).

I re-checked the full diff and don't see any remaining code-level blocker. The earlier `assets/api/index.html` note about the PR description is still applicable — the base file used `unpkg.com`, not `app.unpkg.com` — but that's documentation accuracy rather than a reason to hold up the fix.

Looks good from my side.

PR #893, opened 2026-09-16
feat(startup): restyle failure screens and add database recovery actions

tadelv:
## Summary

This PR now has two related startup-failure improvements: it brings the standalone early-boot failure screens into Decaid's `shadcn_ui` design language, and it adds recovery actions to the database-failure path so users can preserve their local data or reset the broken database state.

- Add a shared `StartupFailureShell` (`ShadApp` + Decaid light/dark `buildDecentTheme`, `ThemeMode.system`) used by both standalone startup-failure apps.
- Restyle `webserver_port_conflict_app.dart` with `ShadAlert`, `ShadButton`, and Lucide icons while preserving its existing port-probe, retry, result, and close behavior.
- Restyle `database_failure_view.dart` in the same design language and add two serialized actions:
  - **Save recovery package**: preserves raw SQLite plus `-wal`/`-shm`/`-journal` sidecars, Hive storage, and the two known log files.
  - **Reset local database**: confirmed destructive reset of the SQLite file/sidecars and Hive store, while leaving credentials, preferences, logs, plugins, and web-ui untouched.
- Extract shared archive-delivery primitives and promote `StreamingZipWriter` so Data Management and startup recovery use the same bounded-memory archive infrastructure without changing the HTTP export contract.
- Stage desktop archives in app-owned temp space before copying to the selected path, which keeps the signed macOS build inside its App Sandbox grant. Recovery-package destination checks prevent ordinary path/symlink aliases from overwriting the data being preserved.
- Add previews plus regression coverage for the startup themes, recovery UI state, archive delivery, support-package contents, reset behavior/retries, and the existing data-export path.

## Linked Issue

N/A - maintainer change; scope and plan agreed in review before implementation.

## Verification

- `dart format lib test` -> 873 files, 0 changed.
- `flutter analyze` -> No issues found.
- `flutter test` -> 4341 passed, 1 skipped (40 new tests).
- Focused coverage includes `test/unit/ui/webserver_port_conflict_app_test.dart`, `test/database_failure_view_test.dart`, `test/unit/services/export/*`, `test/unit/services/storage/database_recovery_test.dart`, and `test/data_export/`.
- Regression tests cover Shad/Decent theme wiring, occupied/free port probing, serialized recovery actions, compact-screen/large-text behavior, support-package source allowlisting, destination alias protection, app-temp archive staging, SQLite rollback-journal preservation, partial-reset reporting, and retry cleanup of prior quarantine.
- No hardware is required for the automated coverage. The macOS sandbox constraint is documented explicitly; CI itself runs the Linux smoke/build path.

## Impact

- **User-visible:** the port-conflict and database-failure startup screens now match the rest of Decaid. The database-failure screen additionally provides save and reset recovery actions.
- **Storage/recovery:** reset can remove the local SQLite database, its `-wal`/`-shm`/`-journal` sidecars, and the Hive store after explicit confirmation. Credentials, SharedPreferences, logs, plugins, and web-ui are preserved.
- **Security/safety:** recovery-package destinations are checked against source directories and symlinked paths; desktop staging occurs in app-owned temp space so the signed macOS build does not require sibling access beside the user-selected file.
- **Compatibility/API/spec:** no REST/WebSocket contract, schema version, migration, or dependency changes. Existing HTTP export contracts and limits remain intact.
- **Documentation:** storage/build notes and the archived recovery design document record the recovery semantics, sandbox constraint, review findings, and accepted residual risks.

## Recovery actions

The failure screen was previously read-only: it asked the user to share a data
package that nothing in the app produced, with no way off the screen. Two
serialized actions are now available.

- **Save recovery package** - raw SQLite plus its `-wal`/`-shm`/`-journal`
  sidecars, the Hive store, and `log.txt`/`webview_console.log`, written to a
  user-picked location. All three sidecars are collected because AppDatabase
  does not pin `journal_mode`, so an interrupted transaction may leave a rollback
  `-journal` rather than `-wal`/`-shm`. Log collection is an allowlist of two
  literal names: on mobile `AppDirectories.logs` is the whole support root, so a
  recursive walk would also capture the database, Hive store, plugins, and web-ui.
- **Reset local database** - rename-then-delete of the SQLite file, its
  `-wal`/`-shm`/`-journal` sidecars, and the Hive store. Credentials, preferences,
  logs, plugins, and web-ui are untouched. A partial reset is reported rather
  than shown as success. Reset is not gated on a successful save, but save/reset
  operations are mutually exclusive.

Archive build/save primitives were extracted from `DataManagementPage` into
`lib/src/services/export/archive_export.dart`, and `StreamingZipWriter` was
promoted out of the webserver package, so the failure screen and Data Management
share one implementation. The HTTP export path keeps its existing caps and
compression level; the writer enforces ZIP32 structural bounds independently of
those optional HTTP limits.

Large payloads are staged in the app's own temporary directory and then copied
onto the selected path. Staging must not sit beside the destination because the
macOS App Sandbox grants only the selected URL. Destination safety is enforced
by directory rather than per-file path comparison: drift/log directories by
equality, Hive by containment because it is enumerated recursively, all after
`resolveSymbolicLinks`.

Design rationale and review history: `doc/plans/archive/db-failure-recovery/`.

## Contributor Responsibility


ODevStudio (approved):

tadelv:
## Added after this PR was opened: database-recovery actions

This branch gained a second, independent change after the PR was opened. Commit `a387f27a` adds recovery actions to the database-failure screen that this PR restyles, and touches shared export code. Re-requesting review because the earlier approval predates it.

### What the failure screen does now

It was read-only: it told the user to share a "data package" that nothing in the app produced, and offered no way off the screen. There are now two actions, serialized against each other.

- **Save recovery package** - raw SQLite plus `-wal`/`-shm`, the Hive store, and `log.txt`/`webview_console.log`, written to a user-picked location.
- **Reset local database** - rename-then-delete of the SQLite trio and the Hive store. Credentials, preferences, logs, plugins, and web-ui are untouched. A partial reset is reported explicitly rather than shown as success. Reset is deliberately not gated on a successful save (a failed save would otherwise strand the user on the failure screen), but the two operations are mutually exclusive.

### Scope beyond the screen

- Extracted archive build/save primitives from `DataManagementPage` into `lib/src/services/export/archive_export.dart`; Data Management's three exports now use them.
- Promoted `StreamingZipWriter` from `lib/src/services/webserver/data_export/` to `lib/src/services/export/`. The HTTP export path keeps its own caps and compression level; the writer now enforces its ZIP32 structural bounds independently of the optional HTTP limits, so a permissive caller cannot truncate a 32-bit or 16-bit field into a corrupt archive.
- Large payloads stream into a uniquely created staging directory beside the destination instead of a temp build plus a full copy. No predictable temporary path is ever opened for writing.
- Android/iOS keep the deferred share-sheet cleanup.

### Two things worth a reviewer's attention

1. **Log collection is an allowlist of two literal names.** On mobile `AppDirectories.logs` is the whole support/Documents root, so a recursive walk would have embedded the database and the Hive store inside a `logs/` prefix. A regression test pins the exact entry set.
2. **Destination safety is enforced by directory, not by per-file path comparison.** The drift and log directories are rejected by equality (their sources are direct children); the Hive directory by containment, because it is enumerated recursively. Directories are compared after `resolveSymbolicLinks`, which resolves symlinked ancestors and returns on-disk case. A lexical per-file check cannot see a symlinked parent, and on case-insensitive APFS `settings.hive` and `Settings.hive` are the same file while comparing unequal.

### Verification

- `dart format lib test` -> 873 files, 0 changed
- `flutter analyze` -> No issues found
- `flutter test` -> 4339 passed, 1 skipped (36 new tests)

Five independent a-team review rounds ran against this. Every blocking finding was fixed, and each fix has a test that was proven to fail without it; the three sharpest were red-proved, one of which reproduces real truncation (`Expected: 'SOURCE'` / `Actual: 'archive'`).

### Deliberately accepted residual risk

- No test forces a real desktop rename failure or a real reset rename/delete failure; both need OS-level failure injection that was not reproducible in the test environment.
- The Windows replacement path is untested on Windows.
- `_freeSiblingPath` checks for a free name rather than reserving it (single `exists` check window).

Full design rationale and the review history: `doc/plans/archive/db-failure-recovery/`.

tadelv (commented):
Blocking findings on the newly added recovery/export scope:

1. **The desktop staging strategy is not safe for the sandboxed macOS release.** `writeArchiveToDestination` calls `Directory(p.dirname(destinationPath)).createTemp(...)` after `FilePicker.saveFile()` returns a user-selected file. Decaid's release entitlements enable App Sandbox + `com.apple.security.files.user-selected.read-write`; an `NSSavePanel` grants access to the selected file URL, not arbitrary sibling files/directories. Creating `.decent-export-*` beside the selected file can therefore fail in the signed macOS app even though Linux CI passes. Please make the macOS path operate only on the granted URL (or stage in an app-owned location and copy/coordinate the final write), and add whatever sandboxed macOS verification is practical.

2. **A retry can turn a partial database reset into a false success.** If a target rename succeeds but deletion of `<target>.reset-<timestamp>` fails, that path is returned in `leftovers`. The next `DatabaseReset.run()` only enumerates the original DB/WAL/SHM/Hive paths; because the original was already renamed, it is skipped, the prior `.reset-*` artifact is never revisited, and the retry can return `isClean == true`. The UI then says `Data cleared. Close and reopen Decaid.` while the previous leftover still exists. This contradicts the design's requirement that a partial result never be presented as success. Please make previous reset leftovers part of subsequent cleanup/cleanliness accounting (or otherwise persist the partial state), and add a deterministic regression test for the actual I/O failure/retry path rather than only constructing a synthetic `ResetReport`.

The rest of the new scope looks carefully designed, and the current CI run is green; these are the two issues I would fix before merging.

tadelv:
Both blockers are fixed in `9682b541`. Thanks - the sandbox one in particular would have broken the existing full-backup path on the signed macOS build, and I would not have caught it from CI.

## 1. macOS sandbox vs sibling staging - fixed

You were right, and I confirmed the mechanism: `macos/Runner/Release.entitlements` (and `DebugProfile.entitlements`) carry `com.apple.security.app-sandbox` plus `com.apple.security.files.user-selected.read-write`, which covers the selected URL rather than its directory.

Worse than "the new feature fails" - the previous commit had also moved `DataManagementPage._exportFullBackup` onto `deliverArchive`, so this would have regressed a shipping path. The pre-change code staged in `TempArchiveDir` and then did `zipFile.copy(outputFile)`; I had replaced that with a sibling staging directory to avoid a second write.

Reverted to staging in the app temp directory (`TempArchiveDir`) followed by a copy onto the granted path. That also removes the entire Windows rename-and-backup dance - the destination is only ever touched by a plain copy, which already overwrites on every platform. I dropped `ArchiveDeliveryException` and `_freeSiblingPath` with it. The delivery path now additionally deletes a pre-existing symlink at the destination before copying, so a link cannot be followed into a source file.

Cost, stated plainly: delivery is a copy, not an atomic rename, so an interrupted copy can leave a partial file at the destination. That is exactly the behaviour that shipped before this PR, so it is not a regression, but it is a real trade-off and it is now recorded in `doc/AI_BUILD_NOTES.md` Footgun #6 instead of the Windows rename note that no longer applies.

## 2. Retried partial reset reporting success - fixed

Confirmed by reading it: `run()` renamed `X` to `X.reset-<timestamp>`, reported a failed delete in `leftovers`, and a second press iterated only `[driftFile, driftFile-wal, driftFile-shm, hiveDir]`. With the originals already renamed away those were skipped, so the retry returned `isClean == true` and the UI said "Data cleared" while the quarantine was still on disk.

`run()` now reclaims every `<target>.reset-*` sibling before it touches the live targets, and reports any it cannot delete. A retry therefore either finishes the job or stays unclean - it can no longer claim success while quarantined data remains.

## Test coverage for both

- `archive_export_delivery_test.dart`: asserts the staging path is **not** inside the destination's directory, so returning to sibling staging fails the suite.
- `database_recovery_test.dart`: a POSIX-only `chmod`-based test that forces a real delete failure and asserts `isClean == false` with the quarantine path in `leftovers`; and a retry test that pre-seeds `<driftFile>.reset-1` and asserts the retry reclaims it.
- `a destination symlink is replaced rather than followed` covers the link case directly.

I red-proved both rather than assuming. Restoring sibling staging fails with `Expected: not a string starting with '<tempdir>/'`, and removing the reclaim makes the retry report `isClean == true` while `.reset-*` data remains - the exact bug you described.

## Verification

- `dart format lib test` -> 873 files, 0 changed
- `flutter analyze` -> No issues found
- `flutter test` -> 4341 passed, 1 skipped

Design doc and AI notes updated; the round-6 findings and the accepted trade-off are recorded in `doc/plans/archive/db-failure-recovery/2026-09-17-design.md`.

tadelv (commented):
One remaining change before I would call this merge-ready: the archived design doc still describes the superseded delivery implementation.

`doc/plans/archive/db-failure-recovery/2026-09-17-design.md` currently still says, in the main **Shared module** section, that desktop delivery creates a sibling partial in the destination directory and renames it into place; the **Slow path** section still says the double write was removed and desktop pays one write; and the test plan still talks about a failing rename. The actual code now correctly stages in `TempArchiveDir` and copies to the selected path specifically because sibling staging is invalid under the macOS sandbox.

The correction row appended near the bottom records why the design changed, but it leaves the retained design itself internally contradictory. Since `AGENTS.md` explicitly keeps archived design docs as the project's durable “why”, please update those earlier sections to describe the final temp-stage + copy design, including that the desktop double write is intentionally back as the sandbox-safe tradeoff, and remove the obsolete rename-test language.

I did not find another runtime blocker in the new fix: the macOS staging issue and reset-retry false-success issue are both addressed, with regression coverage, and CI is green.

ODevStudio (changes requested):
Two remaining issues I would fix before merging:

1. **Recovery misses SQLite's rollback journal.** `AppDatabase.defaults()` doesn't opt into WAL mode, so SQLite can use its default rollback-journal mode and leave `streamline_bridge.sqlite-journal` after an interrupted transaction. The new recovery package only saves `-wal`/`-shm`, and `DatabaseReset` only removes those sidecars as well. On a database-open/migration failure, the rollback journal can be exactly the state needed for recovery. Please include `-journal` in both the support-package and reset sidecar sets, with regression coverage.

2. **The archived design doc still describes the superseded delivery mechanism.** The retained design still says desktop export stages a sibling partial, renames it into place, removes the double write, and tests rename failure. The implementation now correctly stages in `TempArchiveDir` and copies to the selected path because sibling staging is incompatible with the macOS sandbox. Since archived design docs are kept as the durable rationale in this repo, please update the main design sections and test-plan text to describe the final temp-stage + copy tradeoff rather than leaving the correction only in the review-history appendix.

I intentionally did not include the hard-link aliasing case here; I consider that follow-up hardening rather than a merge blocker.

tadelv:
Both fixed in `86cfc10d`.

## 1. Rollback journal missing from recovery and reset - fixed

Correct, and I verified the mechanism rather than assuming it. `AppDatabase.defaults()` builds `driftDatabase(...)` with no `setup:` callback and sets no `journal_mode`; nothing in the repo pins it either, and the only mention of `journal_mode` in the pinned Drift version is a doc comment telling callers they *can* set WAL (`drift-2.34.0/lib/native.dart:143`). SQLite therefore runs its default rollback journal, and an interrupted transaction leaves `streamline_bridge.sqlite-journal` rather than `-wal`/`-shm`.

That made this the worse of the two failure modes: on an open/migration failure the journal can hold exactly the state needed to recover the file, so the package was omitting recovery data while the reset was leaving it behind.

- `support_package.dart`: sidecar allowlist is now `['-wal', '-shm', '-journal']`.
- `database_recovery.dart`: reset targets now include `$_driftFile-journal`.

Only one sidecar family exists at a time and each is existence-checked, so covering all three costs nothing. Red-proved: dropping `-journal` from either site fails its test (`Expected: Set:[...streamline_bridge.sqlite-journal...]` for the package, and the `moved` assertion for the reset).

## 2. Archived design doc describing the superseded delivery - fixed

Fair, and I agree with the reasoning: if the archived design is the durable "why", a doc that contradicts the shipped code is worse than none. I had appended the correction and updated the appendix, but left the main body describing the old mechanism.

Corrected in `2026-09-17-design.md`:

- **Shared module**: `deliverArchive` now documents staging in a `TempArchiveDir` and copying to the selected path, explains why `ArchiveTarget` carries both paths, and states explicitly that staging is not beside the destination because the sandbox grants only the selected URL.
- **Slow path**: the double write is now recorded as **deliberately abandoned**, with the reason and the fact that the desktop path pays the copy again - the same cost as before this change, not a regression. Items 2 and 3 (no per-entry buffering, packing off the frame loop) still stand.
- **Tests**: the "failing rename surfaces failure and cleans the partial" line is gone, replaced by the staging-location and destination-symlink tests that exist; the reset delete-failure retry coverage is described as it is now implemented.
- **Docs checklist**: the obsolete `File.rename`-on-Windows note is replaced by the macOS sandbox constraint.
- **Review rounds**: the table is now explicitly framed as per-round history, with a note that round 6 supersedes rows 1-5 for anything about staging and renaming.
- Package-contents table and reset target list now include `-journal`, with the journaling-mode rationale.

On the hard-link aliasing case you set aside: agreed it is not a merge blocker, and it is now noted in the design doc's residual risks rather than silently dropped.

## Verification

- `dart format lib test` -> 873 files, 0 changed
- `flutter analyze` -> No issues found
- `flutter test` -> 4341 passed, 1 skipped

tadelv (commented):
The previous design-doc blocker is fixed, and I found no new code/test blocker in the latest delta. The `-journal` addition is appropriate given that Decaid does not pin SQLite to WAL, and the package/reset tests cover it.

One merge-hygiene issue remains in the PR description itself: the original required `## Summary` / `## Impact` sections still describe the PR as presentation-only (`without touching startup/error behavior`, `Behavior ... unchanged`, `No ... database recovery logic touched`). That is now materially false because this PR adds recovery-package export and a destructive database reset. The later `## Recovery actions` section explains the new behavior correctly, but it leaves the required template sections internally contradictory.

Please fold the recovery functionality into the main Summary/Impact sections (or remove the stale presentation-only claims) before merge. `AGENTS.md` treats the PR template as a hard gate, and the template specifically asks Impact to record user-visible, security, and behavioral effects.

Otherwise the current implementation looks clean from this review. The head CI run is still in progress at the time of review, so this comment is not asserting that final CI has completed.

